PR #74, opened 2014-06-17
Multihead Wallpaper support

pakdel:
It is a simple implimentation and does not center or resize.

### DIFF
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -445,7 +445,7 @@ void Application::preferences(QString page) {
 }
 
 void Application::setWallpaper(QString path, QString modeString) {
-  static const char* valid_wallpaper_modes[] = {"color", "stretch", "fit", "center", "tile"};
+  static const char* valid_wallpaper_modes[] = {"color", "stretch", "fit", "center", "tile", "multihead"};
   DesktopWindow::WallpaperMode mode = settings_.wallpaperMode();
   bool changed = false;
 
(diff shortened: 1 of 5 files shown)


jleclanche:
What does it actually do? "Multihead" is not exactly descriptive.

pakdel:
I simply wanted to be able to set a large wallpaper for my multihead configuration, and have it properly split and span all my displays.
For example, if you have three 1920 × 1080 monitors and you use Xinerama with a horizontal layout, it results in a 1920 × 3240 desktop, so you should be able to set a big 1920 × 3240 panorama picture as your wallpaper and see it properly divided into three 1920 × 1080 segments and set from left to right on your X screen root windows, which results in something that looks like the original picture.

jleclanche:
@pakdel I'm sorry nobody got to this sooner. Could you make this option only show up in the UI when the user either already has it selected, or has more than one screen?

pakdel:
Sure. I do not have any multihead setup around until Monday, so I will start then.

jleclanche:
@pakdel I'm sorry about the short notice, but letting you know: We're starting a feature freeze in a couple of days, so this will have to land soon if you want it in 0.9. I'd personally like to see it there.

pakdel:
@jleclanche I have made the change you asked for. Let me know if there is anything else.

jleclanche:
@pakdel Im testing it and I don't understand why it's better than "Center on the screen". If you have a wallpaper of the appropriate resolution, it should work fine.

pakdel:
@jleclanche nice catch! Changes that are made (to the work area) since my initial pull request have made my code useless :D

pakdel:
I am closing this.

jleclanche:
Ah, cool. =)

PR #112, opened 2014-12-16
Fix #111 - Can't change bookmarks

### LINKED ISSUE #111: Can't change bookmarks
A few days ago I changed my username and with it moved my home folder. So far I managed to change the settings of all applications except the bookmarks of pcmanfm-qt. I tried:
- Changing the existing enties.
- Removing the existing entries and creating new ones.
- Adding new entries
But when I click on OK all my changes are reverted. As I added new entries in the past I guess the problem is that the old folder don't exist anymore.

ixjlyons:
Fixes #111. `~/.gtk-bookmarks` is considered the legacy/fallback location for bookmarks in `libfm`, so the new location should be checked first before saving. See [`fm_bookmarks_init()`](https://github.com/lxde/libfm/blob/06424ed5c0b1e83c603b1c94bf15cae748b61d78/src/base/fm-bookmarks.c#L184) in `libfm` for comparison.

### DIFF
--- a/libfm-qt/editbookmarksdialog.cpp
+++ b/libfm-qt/editbookmarksdialog.cpp
@@ -75,7 +75,11 @@ void EditBookmarksDialog::accept() {
   }
 
   // FIXME: should we support Qt or KDE specific bookmarks in the future?
-  char* outputFile = g_build_filename(g_get_home_dir(), ".gtk-bookmarks", NULL);
+  char* outputFile = g_build_filename(g_get_user_config_dir(), "gtk-3.0", "bookmarks", NULL);
+  if (!g_file_test(outputFile, G_FILE_TEST_IS_REGULAR)) {
+    g_free(outputFile);
+    outputFile = g_build_filename(g_get_home_dir(), ".gtk-bookmarks", NULL);
+  }
   // we use glib API here because the API is atomic.
   g_file_set_contents(outputFile, buf.constData(), buf.length(), NULL);
   g_free(outputFile);


jleclanche:
@PCMan @LStranger can either of you review this before 0.9 please?

jleclanche:
@PCMan @LStranger ping

PCMan:
This is not the optimal way to fix the issue.
Many people are using a gtk2/gtk3 mixed environment.
So the content of both files, if possible, should be kept in sync.
Only updating either one of them is not enough.
Please give me some time and I'll figure out how to do this.

PCMan:
Just confirmed with @LStranger. Since gtk+ 2.24.12, gtk+2 can read the bookmarks files of gtk 3 if it exists. In libfm he did the same. However, I still found the following problem.
1. start a gtk2 program while there is no gtk3 bookmarks, so it will load the gtk2 one.
2. start a gtk3 program and edit the bookmarks and now you have a gtk3 bookmarks.
3. edit the bookmarks with the gtk2 program and quit the program
4. restart the gtk2 program, it loads the new gtk3 bookmarks file instead, and your previous changes in the old file are gone.

PCMan:
After some discussions with @LStranger, we decide to leave it as it is. Gtk 2.24.12 already solves this issue and libfm currently has the same behavior as gtk2.24, let's not change it.

jleclanche:
Rejected then?

PCMan:
No, there is a bug in pcmanfm-qt that also cause the problem.
Part of the fix lands in b832a89c80313c253447c6d07ab2c2ee48d6c711.
But editing bookmarks from the side pane still does not work.

PR #119, opened 2015-01-15
Portuguese update

smarquespt:


luis-pereira:
@smarquespt Thanks. Landed on dd9806a22980fbf4b3ed08d232b6eb75

PR #124, opened 2015-01-28
Remember maximized window state

dshizzle:
At the moment only the window size is being saved but not the window's maximized state, if the "Remember the size of the last closed window" option is active. This patch adds the ability to open new windows not just with the last size but also the last maximized state.


PCMan:
merged. Thanks!

PR #139, opened 2015-02-08
- Bump cpack versioning

heliocastro:


jleclanche:
This should go away, we depend on liblxqt already

luis-pereira:
@jleclanche actually we don't.

jleclanche:
Hmm, this is hard to justify. I'll think about it.

luis-pereira:
@jleclanche Not sure what you meant. pcmanfm-qt doesn't depend on liblxqt in any way that I know of.

jleclanche:
You can merge this for now, it doesnt matter.

luis-pereira:
@jleclanche I won't. The version is hardcoded in two distintt places:
https://github.com/lxde/pcmanfm-qt/blob/master/CMakeLists.txt#L4
https://github.com/lxde/pcmanfm-qt/blob/master/CMakeLists.txt#L92

luis-pereira:
Superseded by 7aec9da0dfbdbc0d77f171a4e000de19b4fecdae

Issue #151, opened 2015-03-02
Don't confirm when moving files to the trash


goetzc:
As a file deleted to the trash can be undeleted, it is not so important to ask confirmation, and it can even be annoying to the user. But a confirmation before deleting a file (bypassing trash) is convenient to have. 

I propose to change in Preferences the behavior of `Confirm before deleting files` so that it only asks confirmation when deleting files bypassing the trash.

PCMan:
it's already implemented.

PR #152, opened 2015-03-02
Don't confirm when moving files to the trash

goetzc:
for lxde/pcmanfm-qt#151

Only asks confirmation when deleting files bypassing the trash.
This behavior is consistent with the Preference "Confirm before deleting files".

Settings changed to have two individual options, and "Deleting files or folders" selected by default:
```
Confirm before:
  [ ] Moving files or folders to the trash
  [x] Deleting files or folders
```

### DIFF
--- a/libfm-qt/filemenu.cpp
+++ b/libfm-qt/filemenu.cpp
@@ -58,6 +58,7 @@ FileMenu::~FileMenu() {
 
 void FileMenu::createMenu(FmFileInfoList* files, FmFileInfo* info, FmPath* cwd) {
   useTrash_ = true;
+  confirmTrash_ = true;
   confirmDelete_ = true;
   files_ = fm_file_info_list_ref(files);
   info_ = info ? fm_file_info_ref(info) : NULL;
@@ -311,7 +312,7 @@ void FileMenu::onCutTriggered() {
 void FileMenu::onDeleteTriggered() {
   FmPathList* paths = fm_path_list_new_from_file_info_list(files_);
   if(useTrash_)
-    FileOperation::trashFiles(paths, confirmDelete_);
+    FileOperation::trashFiles(paths, confirmTrash_);
   else
     FileOperation::deleteFiles(paths, confirmDelete_);
   fm_path_list_unref(paths);
--- a/libfm-qt/filemenu.h
+++ b/libfm-qt/filemenu.h
@@ -45,9 +45,17 @@ public:
   bool useTrash() {
     return useTrash_;
   }
-  
+
   void setUseTrash(bool trash);
-  
+
+  bool confirmTrash() {
+    return confirmTrash_;
+  }
+
+  void setConfirmTrash(bool confirm) {
+    confirmTrash_ = confirm;
+  }
+
   bool confirmDelete() {
     return confirmDelete_;
   }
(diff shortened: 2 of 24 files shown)


jleclanche:
UX, gtm, please land

paulolieuthier:
That's not the correct way to fix it. The "prompt" parameters is still there.

goetzc:
You are right, fixed.

paulolieuthier:
Good! But you can't simple remove it. This method is used in other places in pcmanfm-qt and this patch will break stuff.

```bash
➜  pcmanfm-qt git:(master) grep trashFiles . -Rn | grep -v build
./libfm-qt/filemenu.cpp:314:    FileOperation::trashFiles(paths, confirmDelete_);
./libfm-qt/fileoperation.h:104:  static FileOperation* trashFiles(FmPathList* srcFiles, bool promp = true, QWidget* parent = 0);
./libfm-qt/fileoperation.cpp:227:FileOperation* FileOperation::trashFiles(FmPathList* srcFiles, bool prompt, QWidget* parent) {
./pcmanfm/desktopwindow.cpp:637:      Fm::FileOperation::trashFiles(paths, settings.confirmDelete());
./pcmanfm/mainwindow.cpp:667:    FileOperation::trashFiles(paths, settings.confirmDelete(), this);
```

goetzc:
I didn't thought about that. Not sure what to do, should those calls be modified to not include the prompt parameter?

paulolieuthier:
Probably the prompt parameter should be kept there and every caller in my previous post (except the first one) call `settings.confirmDelete()` and pass it as the `prompt` parameter? If you want not to asked by default, you have to change that setting.

goetzc:
I have separated the delete confirmation settings, now it looks this way by default:

```
Confirm before:
  [ ] Moving files or folders to the trash
  [x] Deleting files or folders
```
And it seems to work without breaking other stuff :)

jleclanche:
Please don't, it's fine to remove the confirmation before moving to trash entirely. The trash is there to provide the user with an undo button, which confirmation messages don't do. That's the whole point :)

PCMan:
On Wed, Mar 4, 2015 at 12:40 PM, Jerome Leclanche <notifications@github.com>
wrote:

> Please don't, it's fine to remove the confirmation before moving to trash
> entirely. The trash is there to provide the user with an undo button, which
> confirmation messages don't do. That's the whole point :)
>
Please keep it.
There is nothing wrong to make it optional and turned off by default.
The confirmation message is for someone who might accidentally hit "delete"
key.
Then the files just disappear quietly.
BTW, moving to trash is not always a fast operation that can be finished
immediately.
For filesystems which do not support trash or there are permission
problems, this can be very slow.
Having an "option" to prevent accidental deletion is good.
Just turn it off by default.

jleclanche:
If the FS does not support trash, then the "move to trash" operation should be considered a delete operation and *would* pop up a confirmation dialog.

And if the operation takes time (it shouldn't), we should display a progress dialog which can be cancelled.

The confirmation message when *moving to trash* is redundant. Follow the basic UX principle of giving feedback to the user (eg by displaying "document.html has been moved to the trash") and providing a way to undo (an undo button and ctrl+z). Look at how gmail does it for example.

Forcibly deleting is an operation that power users should have access to, through an accelerator only, and which should always display a confirmation message. Really, this option should be gone altogether.

PCMan:
The current code in git already handles this and provided an option to turn off the confirmation.
So the PR is out of date and I close it.

PR #166, opened 2015-04-05
Sort indicators in detailed view, fixes #109.

### LINKED ISSUE #109: detailed list view improvements
PCManFM's detailed list view lacks some rather important features, imho:

Sorting should be configurable by clicking the columns' headers. This affects both choosing the column that's used to do the sorting and the order in terms of ascending vs. descending.

Some columns are just missing, in practice group names, permissions and targets of symbolic links.

The choice which columns are displayed should be configurable.

ufleisch:
It would be nice if the Qt-version had sort indicators on the headers of the table in detailed view, so that the columns could be sorted by clicking on the headers, not only using the menu. This commit implements this feature.

### DIFF
--- a/libfm-qt/folderview.cpp
+++ b/libfm-qt/folderview.cpp
@@ -195,6 +195,11 @@ FolderViewTreeView::~FolderViewTreeView() {
 void FolderViewTreeView::setModel(QAbstractItemModel* model) {
   QTreeView::setModel(model);
   layoutColumns();
+  if(ProxyFolderModel* proxyModel = qobject_cast<ProxyFolderModel*>(model)) {
+    connect(proxyModel, &ProxyFolderModel::sortFilterChanged, this, &FolderViewTreeView::onSortFilterChanged,
+            Qt::UniqueConnection);
+    onSortFilterChanged();
+  }
 }
 
 void FolderViewTreeView::mousePressEvent(QMouseEvent* event) {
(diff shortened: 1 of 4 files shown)


PCMan:
Thank you for the patch.
However, it might be better to manage this in one place by letting the model maintain the sorting information, and the view reflects the change of the underlying model automatically.
This is more natural for MVC-based designs.

PCMan:
I'll take some time to integrate your changes later. :)

ufleisch:
I agree that the information about the sort column and order should be taken from a model. The model information would be `Settings::sortColumn()` and `Settings::sortOrder()`. These properties are defined in `Settings` but not used anywhere in the code. I noticed that and thought that we could fix that in a later commit. The code in the current commit sets the sort column and order to fixed values (`Fm::FolderModel::ColumnFileName`, `Qt::AscendingOrder`) in order to work with the existing code. Whether we first set the sort indicators in the view as it is done by this commit and then fix the issue with the unused settings properties or do it the other way round should not matter for the final result.

ufleisch:
I have added another commit to store the sort column and order in the settings and use the stored values when sorting the model.

ufleisch:
I will soon refactor my commits accordingly.

PCMan:
Cool! Thanks!

ufleisch:
I have now added another commit and hope that the code is closer to what you expect. I had to use `FolderViewTreeView::setModel()` instead of the suggested `FolderView::setModel()` because the latter is not called when the view mode is switched. The new slot is therefore in `FolderViewTreeView` and not `FolderView`. By doing these changes, I was able to undo some of the changes I introduced in the last two commits, so you better look at the changes of the whole branch. You could also consider squashing the commits once you are satisfied.

PCMan:
Looks good. :) Thank you very much.
I'd like to merge your changes.
Would you please squash the commits into one commit?

ufleisch:
I've squashed the commits.

ufleisch:
I have amended the commit.

PCMan:
merged. Thanks!

PR #173, opened 2015-05-15
Add menu items and shortcuts for new folder and blank file, fixes #163.

### LINKED ISSUE #163: enhance creating new folders
New folders can be created by accessing the context menu available over blank space in right pane only. This menu can't be accessed if e. g. in Directory Tree view a folder is opened the content of which is filling all vertical space and the horizontal space available is limited.
Within the left pane folders can't get created at all.
So I'd suggest to add an item to create new folders to the context menu of folders both in right pane and in left pane's Directory Tree view.

The regular menu as accessed from the menu bar doesn't comprise an item to add new folders either.
I think this should be added, too.

ufleisch:
In detailed list view mode, it is only possible to create a new folder if there are less file entries than would fit into the window and the context menu can be activated in the empty space. This commit adds the "Create New" menu as we have it in the GTK version of pcmanfm. This adds also keyboard shortcuts as requested in issue #163.

### DIFF
--- a/libfm-qt/foldermenu.cpp
+++ b/libfm-qt/foldermenu.cpp
@@ -267,14 +267,14 @@ void FolderMenu::onCreateNewFile() {
   FmPath* dirPath = view_->path();
 
   if(dirPath)
-    createFile(CreateNewTextFile, dirPath);
+    createFileOrFolder(CreateNewTextFile, dirPath);
 }
 
 void FolderMenu::onCreateNewFolder() {
   FmPath* dirPath = view_->path();
 
   if(dirPath)
-    createFile(CreateNewFolder, dirPath);
+    createFileOrFolder(CreateNewFolder, dirPath);
 }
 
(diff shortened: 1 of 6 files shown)


excalibur1234:
this sounds really useful.
i support it.

palinek:
GTM.

Just a small note: as the `Fm::createFile()` is now able to create also a directory -> the name is a bit confusing.

paulolieuthier:
@ufleisch can you please update the name, squash and rebase?

ufleisch:
Is it OK if I rebase the branch to master and add a commit which renames `Fm::createFile()` and all its usages to `Fm::createFileOrFolder()`?

paulolieuthier:
Yes, but squash the two commits to make only one.

ufleisch:
OK, the commit is now rebased and fixed up.

paulolieuthier:
Thanks.

PR #178, opened 2015-05-25
Fix issues with resizing columns

ufleisch:


palinek:
both commits seems reasonable, GTM

jleclanche:
2nd commit gtm but please fix that conditional.

ufleisch:
OK, I have fixed it up and force-pushed.

PR #180, opened 2015-05-30
Display folders first when active and sort order descending, fixes #179.

### LINKED ISSUE #179: Sorting folder first does not work with descending sort order
When sort order is descending and folder first is active, folders are last instead of first. Besides this, the "Folder First" setting is not stored persistently.

ufleisch:


paulolieuthier:
Thanks!

PR #182, opened 2015-05-30
Hide tab bar when it contains less than 2 tabs.

ufleisch:


paulolieuthier:
There is an option to always show the tab bar.

ufleisch:
Oh, sorry, I did not notice that, thanks.

PR #187, opened 2015-06-21
Context menu in tree side pane, store side pane mode setting, #163.

ufleisch:
This should fix the remaining points mentioned in issue #163. The side pane mode is stored in the settings and there is a context menu in the directory tree side pane.


PCMan:
Merged. Thanks!

PR #191, opened 2015-06-27
Update column layout also if folder already loaded, fixes #190.

### LINKED ISSUE #190: Column layout is not always updated
###### Steps to reproduce

Create a test folder with some contents:
```
mkdir -p "layouttest/a bit longer filename 1"
touch "layouttest/a bit longer filename 1/short" "layouttest/a bit longer filename 2"
```
Now start pcmanfm-qt, set left side pane to "Places", right side pane to "Detailed List".

1. Change to directory "layouttest",
2. then to directory "a bit longer filename 1",
3. then back to parent (Alt-Up).

Everything is displayed correctly.

Now set left side pane to "Directory Tree", and repeat the steps 1-3. The folder name is not displayed correctly, only "a bit l..." instead of "a bit longer filename 1" although there would be enough horizontal space. Actually the "Name" column size has not changed and has still the size suitable for "short".

###### Analysis:

Looking at the debug output there is an additional "folder loaded" message after the "finish-loading" messages in steps 1 and 2 in the case of the "Directory Tree". The messages for step 3 are "start-loading, delete CachedFolderModel, delete FolderModel, finish-loading" in the "Places" case and "start-loading, finish-loading, delete CachedFolderModel, delete FolderModel" in the "Directory Tree" case. Looking at the code, this shows that in the "Directory Tree" case, the folders are also loaded by the `DirTreeModelItem` (hence the "folder loaded" message), this (`fm_folder_is_loaded()`) will cause `TabPage::chdir()` to directly call `onFolderFinishLoading() `(hence the different order of "finish-loading" message).

In the "Places" case, the model is loaded, which invokes `FolderViewTreeView::rowsInserted()`, which calls `queueLayoutColumns()`. In the "Directory Tree" case, the folder is already loaded and the layout is therefore not updated.

ufleisch:
An easier fix would be to connect `modelReset()` of the folder model to `FolderViewTreeView::queueLayoutColumns()`, however, this would bring some unneeded calls to `FolderViewTreeView::layoutColumns()`. The proposed solution will not cause such unneeded calls.

### DIFF
--- a/libfm-qt/folderview.cpp
+++ b/libfm-qt/folderview.cpp
@@ -535,6 +535,12 @@ void FolderView::setViewMode(ViewMode _mode) {
   }
 }
 
+void FolderView::updateAfterFolderLoaded() {
+  if(FolderViewTreeView* treeView = qobject_cast<FolderViewTreeView*>(view)) {
+    treeView->queueLayoutColumns();
+  }
+}
+
 // set proper grid size for the QListView based on current view mode, icon size, and font size.
 void FolderView::updateGridSize() {
   if(mode == DetailedListMode || !view)
--- a/libfm-qt/folderview.h
+++ b/libfm-qt/folderview.h
@@ -100,6 +100,8 @@ public:
 
   void invertSelection();
 
+  void updateAfterFolderLoaded();
+
   void setFileLauncher(FileLauncher* launcher) {
     fileLauncher_ = launcher;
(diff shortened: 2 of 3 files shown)


PCMan:
Thank you for the patch.
I already fixed this with a hopefully cleaner way so the PR no longer applies.
Cheers!

Issue #165, opened 2015-03-31
Build-in search/find modul have to add 


ayhanyalcinsoy:
We can not search file/directory in pcmanfm.

PCMan:
fixed.

PR #193, opened 2015-07-10
Implement a very basic file searching utility.

PCMan:
This is a frequently wanted feature and it solves #165.
Since this is quite a large chunk of changes, more testing and review might be needed before it's merged into master.
So, @luis-pereira, @pmattern, @jleclanche or anyone who is interested, please help.
Thanks! :)


luis-pereira:
@PCMan Testing it

pmattern:
Some thoughts after having a first look:

Would be good if the utility could be launched by a keyboard shortcut as well.

By submitting a search all settings made in the search window are lost. This can be annoying if e. g. several similar searches shall be performed one after the other or a search relying on sophisticated settings failed because just one of them was wrong.
Aside from this I wonder whether it wouldn't be good to have a means to store default search settings not only during consecutive searches but in the long run as well. E. g. either by always remembering the last used ones or adding some search preferences dialogue.

Maybe a history of strings that were filled in the form Name/Location - File Name Patterns would be helpful.
The meaning of "Documents" in tab 'File Type' isn't exactly obvious to me. Maybe some notes what exactly is covered by this topic would be helpful. Strictly speaking 'file type' is a bit misleading as the tab allows for selecting folders as well.

Placing the results in a basically regular PCManFM-Qt window doesn't seem intuitive to me. I think it would be much better if they'd be presented somehow in dialog window 'Search Files' itself.
An UI like this would also allow for keeping the search settings during some consecutive searches as stated above

paulolieuthier:
For a very basic search utility, it's very good. Besides @pmattern thoughts, I'd add a minor glitch:

screen

PCMan:
If there are no objections, I'd like to move this to master after fixing the UI glitches.
Other improvements can be done after getting more user feedback.
What do you guys think?

paulolieuthier:
I say add the keyboard shortcut and merge it.

jleclanche:
@PCMan go ahead and land it, ill do a ux pass on it once it's in.

PCMan:
Merged into master.

PR #196, opened 2015-07-13
CMake: Remove unused INSTALL variables

luis-pereira:
We are using GNUInstallDirs. They are just an artifact.


jleclanche:
gtm.

paulolieuthier:
Ship it.

PR #202, opened 2015-08-02
Update libfm-qt Russian translation and small fix of pcmanfm-qt trans…

Ilya87:
…lation

### DIFF
--- a/libfm-qt/translations/libfm-qt_ru.ts
+++ b/libfm-qt/translations/libfm-qt_ru.ts
@@ -310,12 +310,12 @@
     <message>
         <location filename="../createnewmenu.cpp" line="29"/>
         <source>Folder</source>
-        <translation type="unfinished">Папку</translation>
+        <translation>Папку</translation>
     </message>
     <message>
         <location filename="../createnewmenu.cpp" line="33"/>
         <source>Blank File</source>
-        <translation type="unfinished">Пустой файл</translation>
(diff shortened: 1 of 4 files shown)


pmattern:
Any particular reason why you left those strings in`libfm-qt/translations/libfm-qt_ru[_RU].ts`, context "RenameDialog" and `pcmanfm/translations/pcmanfm-qt_ru[_RU].ts`, context "AboutDialog" untranslated?

Ilya87:
1 - RenameDialog - They are never translated. In my opinion they are marked not right as translatable strings. dest, dest file info, src are always replaced by actual values of filename, path, etc.
screen

2 - Untranslated string in AboutDialog contains only License. And I don't think it is really needed to be translated.

pmattern:
→ libfm
You seem to be right, yes. This can definitely be left as is for now, you may even consider filing an issue to get the translation strings removed.

→ pcmanfm
This doesn't provide the license itself but some kind of short introduction into the legal situation of the software. Providing information like this in native languages does make sense in my opinion which at a glance seems to be shared by the majority of translators.
So no, this should be translated.

Ilya87:
I think now all is done.

pmattern:
So do I.
As you'll probably understand I had to adjust the commit message - stuff like "(part 2)" doesn't make sense in upstream master.
Thanks!

Issue #206, opened 2015-08-21
localisation: several strings missing in TS files


pmattern:
Several strings within the main window aren't subject to translation.

→ Dialog "Preferences" (Menu Edit - Preferences)
* Behaviour
  * Browsing: item "Bookmarks" and all values in corresponding drop-down
  * File Operations: "Erase files on removable ...", "Confirm before moving files ...", "Don't ask options on launch executable..."
* Display: header itself (left pane)
  * User interface: header itself; "Treat backup files..."
* Advanced
  * Templates: section as a whole

→ actual FHS view - context menus
* of FHS objects
  * left pane in tree view: "Open in {New Tab, New Window, Terminal} (correct in right pane); "Create New" plus "Folder" / "Blank File" [0]
  * right pane: "Create New" plus "Folder" / "Empty File" [0]
* over blank space in right pane: "Folder" / "Blank File" [0] in "Create New" (which itself does get translated here)

A probably related finding is that the source files corresponding to contexts AboutDialog, AutoRunDialog, DesktopPreferencesDialog, MainWindow and PreferencesDialog in pcmanfm-qt.ts don't exist according to Qt Linuguist.

[0] correct in menu File

gCurse:
See my pull request #209.

pmattern:
The commit proposed in #209 is addressing dialogue "Preferences" only. It worked as expected here so I merged both the commit and an adjusted version of the German translation.

Items summarized under "FHS view" still open, though. Seems like some tags in the sources are missing (which I can't check myself atm).

gCurse:
See my pull request #211.

pmattern:
Context menu items `Open in {New Tab, New Window, Terminal}` are inconsistent in left and right pane now (`Öffnen in...` vs. `In ... öffnen`). Yours are better.
So could you adjust the ones from right pane to those from the left one and squash it into your commit (`Öffnen in...` everywhere)? This would save a dedicated commit which actually isn't worthwhile.

pmattern:
Closed via #209 and #211.

PR #209, opened 2015-08-26
fixed paths in .ts files +  lupdate

gCurse:
This should fix issue #206. But have a close look at the results :)

In pcmanfm-qt/pcmanfm/translations/
```
for i in `ls *.ts`;
do
	sed -i 's/\.\.\/build\/pcmanfm\/ui_//g' ${i}
	sed -i 's/\.h/\.ui/g' ${i}
	lupdate ../ -ts ${i}
done
```


pmattern:
(see #206)

PR #211, opened 2015-08-28
libfm-qt: lupdated translation files

gCurse:
Here are the (l)updated ts files for libfm-qt.
The missing strings should be in here.

### DIFF
--- a/libfm-qt/translations/libfm-qt.ts
+++ b/libfm-qt/translations/libfm-qt.ts
@@ -300,6 +300,19 @@
     </message>
 </context>
 <context>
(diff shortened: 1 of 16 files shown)


jleclanche:
gtm

pmattern:
@gCurse 
Dropped a little note regarding this PR in #206 two days ago.
Other than that GTM from my side, too.

pmattern:
Made and quashed the changes proposed in https://github.com/lxde/pcmanfm-qt/issues/206#issuecomment-136383258 myself now so as to provide those in libfm-qt.ts for other translators, too.

Landed in e2f7c2f and 6abe362.

PR #212, opened 2015-08-28
xdg main category added

agaida:
prevent some warnings about missed main category in desktop file


agaida:
@jleclanche - What do you think? - Would prevent me to patch this downstream

paulolieuthier:
GTM.

jleclanche:
gtm

PR #215, opened 2015-08-31
Add few missing Polish translations and modify few existing.

dawikur:
Added some missing translations, there are still some blank spaces as I am not sure how to best interpret and translate them.

I changed also some already existing translations, like:
```
from: Wypełnienie 
to: Wypełnij 
```
Those are changes from noun to verb - in my fell this sound better.

*sek* -> *s* - I was taught that *s* is for seconds, not sek (as wiki says *sek* is historical, and *s* is current)
https://pl.wiktionary.org/wiki/sek#pl


paulolieuthier:
Thanks!

PR #216, opened 2015-09-01
Add Greek (el) translation

dglent:
Add the Greek (el) translation


pmattern:
Landed in d7765fc. Thanks again.

PR #218, opened 2015-09-03
Update czech translation

petrbal:


pmattern:
The translation proposed by this PR is incomplete, some strings were translated but are still tagged "unfinished".

6abe362 introduced a large number of necessary changes both in libfm-qt.ts and the preexisting Czech translation.
This is a pretty unfortunate coincidence. It could be attempted to merge the different versions of libfm-qt.ts manually. But imo it would be much quicker to start over.

Aside from this there's the recurring question of country-specific locales.
In some languages like Portuguese those are definitely needed, in some others like German or Italian they aren't. In languages where they aren't needed they should in fact be avoided as they may prevent translations from getting applied. E. g. with locale `de_AT` a translation `de_DE` doesn't get applied but `de` does.
I don't know Czech very well. (In fact I don't know it at all...) But I figure it could belong to those languages where country-specific translations aren't needed.
In that case it would make sense to simply rename the file to `libfm-qt_cs.ts`.

This is a bunch of problems, in particular the changes in libfm-qt.ts are somewhat nasty.
As stated above I fear the easiest thing is to start over. Did you consider using QtLinguist? As you probably still have the translations you made in mind the work could probably done rather quickly using that tool. Plus this would help avoiding flaws like those untranslated strings mentioned above.

PCMan:
I merged the strings manually with `lconvert` tool.
Thanks!

PR #221, opened 2015-09-04
Paste into folder from its context menu.

tsujan:
When a single folder was right clicked and the "Paste" item on its context menu was used, the copied/cut items weren't pasted into it but into the currently open folder (containing it). This commit makes them be pasted into the right clicked folder.

If MULTIPLE folders/files are right clicked, the copied/cut items will be pasted into the currently open folder, as before.


jleclanche:
gtm

PR #222, opened 2015-09-04
Select the folder from where we have gone up.

tsujan:
When the user goes up from a folder whose parent contains several other folders too, he/she may need to quickly see from where he has come there. This second commit makes that possible by selecting and scrolling to the child folder.

Nautilus (and Nemo, by inheritance) has had this useful feature for a long time. Dolphin/Konqueror has never had it. It's nice to have it with PCManFM-qt, IMO.

P.S. This comes after #221 (they have common files), so the diff of #221 is also included.

### DIFF
--- a/libfm-qt/folderview.cpp
+++ b/libfm-qt/folderview.cpp
@@ -712,6 +712,18 @@ FmPathList* FolderView::selectedFilePaths() const {
   return NULL;
 }
 
+QModelIndex FolderView::indexFromFolderPath(FmPath* folderPath) const {
+  QModelIndex index;
+  int count = model_->rowCount();
+  for(int row = 0; row < count; ++row) {
+    index = model_->index(row, 0);
+    FmFileInfo* info = model_->fileInfoFromIndex(index);
+    if(info && fm_file_info_is_dir(info) && fm_path_equal(folderPath,fm_file_info_get_path(info)))
+      return index;
+  }
+  return QModelIndex();
+}
+
 FmFileInfoList* FolderView::selectedFiles() const {
   if(model_) {
     QModelIndexList selIndexes = mode == DetailedListMode ? selectedRows() : selectedIndexes();
--- a/libfm-qt/folderview.cpp
+++ b/libfm-qt/folderview.cpp
@@ -713,6 +713,7 @@ FmPathList* FolderView::selectedFilePaths() const {
 }
 
 QModelIndex FolderView::indexFromFolderPath(FmPath* folderPath) const {
+  if(!model_ || !folderPath) return QModelIndex();
   QModelIndex index;
   int count = model_->rowCount();
   for(int row = 0; row < count; ++row) {
--- a/libfm-qt/folderview.h
+++ b/libfm-qt/folderview.h
@@ -95,6 +95,7 @@ public:
   QItemSelectionModel* selectionModel() const;
   FmFileInfoList* selectedFiles() const;
   FmPathList* selectedFilePaths() const;
+  QModelIndex indexFromFolderPath(FmPath* folderPath) const;
 
   void selectAll();
 
--- a/pcmanfm/tabpage.cpp
+++ b/pcmanfm/tabpage.cpp
@@ -40,7 +40,8 @@ TabPage::TabPage(FmPath* path, QWidget* parent):
   QWidget(parent),
   folder_(NULL),
   folderModel_(NULL),
-  overrideCursor_(false) {
+  overrideCursor_(false),
+  wentUp_(false) {
 
   Settings& settings = static_cast<Application*>(qApp)->settings();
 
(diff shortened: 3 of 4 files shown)


paulolieuthier:
Please, remove the global variable and add as a member variable of the class. Then, GTM.

tsujan:
You mean the static variable "wentUp"?

PCMan:
Thank you for the effort but this approach won't work reliably.
Instead of the item index, IMHO the file name should be remembered instead.
It's possible that the content of the parent folder is changed while you're browsing the child folder. So when you go up, the index you saved becomes invalid.
Care also must be taken to do this only after the parent folder is fully loaded. Otherwise the item you try to select might not be added to the model yet.

tsujan:
> Thank you for the effort...

Thank YOU for your excellent code! It's written so well that I can find its potentials easily.

> It's possible that the content of the parent folder is changed...

I'll think about that. Thanks for the hint!

tsujan:
@PCMan 
Now I remember: indexFromFolderPath() is called inside restoreScrollPos(), which, in turn, is called at the end of onFolderFinishLoading(). So, the index is found only when the parent folder is fully loaded.

tsujan:
Since I'm new here and for the whole team to have a general picture of my commits, I thought it would be helpful to add this:

I always use the commits I make. I'm very fallible but If there were such a big mistake, the commit wouldn't work for me at all ;)

jleclanche:
@tsujan Hey, sorry, this won't make it for 0.10 since there's a design issue with it. But this is something I'd like to have for the next version. Can you rebase and take care of what @PCMan mentioned please?

tsujan:
> ...there's a design issue with it

Which issue, specifically? I use it since I proposed it and have encountered no issue.

> Can you rebase....?

If  you mean making another PR based on the new version, yes, I can.

> ... and take care of what @PCMan mentioned?

I'm afraid what @PCMan mentioned was incorrect.

jleclanche:
@tsujan as he said, if the contents of the parent folder change, the index will change. I'm not  familiar with the pcmanfm-qt code, so I'll need @PCMan to confirm whether this is ok after all.

Regarding rebasing, you donot need to create a new PR. When you do `git push --force`, it updates this PR.

tsujan:
> @tsujan as he said, if the contents of the parent folder change, the index will change...

That's exactly what was incorrect, as I'd explained above.

PCMan:
Sorry for the late reply. Too busy this month.
I'm testing this PR now. There might still be some corner cases that can cause problems.
So it takes longer to review. Thank you for your patience.

PCMan:
OK, finished the review. Here are some comments.
The variable wentUP_ seems to be unnecessary. Also it does not work for all of the cases.
For example, if you go to the parent dir with "go back" rather than "go up", you did not cover this case.

QModelIndex index = folderView_->indexFromFolderPath(history_.at(prevIndex).path());
If the index is valid, select the file and scroll to it. I think this is enough and you do not need wentUp_.

tsujan:
> if you go to the parent dir with "go back" rather than "go up", you did not cover this case...

I used "wentUP_" to exclude "go back" intentionally because only "go up" is for showing the parent folder.

If I've understood your point correctly, this time you agree on the general structure but want "go back" to be covered too. OK, that's easy to implement. First I have to find a way to rebase this old PR if possible at all.

tsujan:
On further consideration, these are my reasons for keeping "wentUp_" and using this method only for "Go Up":

(1) "Go Back" *isn't supposed* to select a previously selected folder/file. It should only restore the scroll position and doesn't select anything with other scenarios either (otherwise, it would be confusing).
(2) Your point could be extended to "Go Forward" too (go up twice, go back twice and then, go forward). Point (1) holds for it too.
(3) We don't want redundant calculations, when we're sure that indexFromFolderPath() would return an invalid index.

PCMan:
I'm thinking about an alternative way to implement this.
Maybe we can store the old FmPath before chdir to a new dir everytime.
Then after a folder is loaded, we can check that if we are moving from a child dir to a parent dir.
All possible cases will work automatically and we do not need to add special handling for back, forward, or up. Even going to parent folder via bookmark will work.
What do you think? Will this be cleaner and easier to maintain?

tsujan:
First, good to know that this feature has got your attention! It's really needed in whatever way it's implemented.

> Maybe we can store the old FmPath before chdir to a new dir everytime.

Great idea!
I just think it would be worth remembering only for "Go UP" (-> points (1) and (2) above). If selecting the child folder is implemented in other cases, why not restoring selections in all cases (which seems useless or even confusing to me)? However, this won't be so important if your idea is implemented because point (3) won't hold then.

PCMan:
You can try other file managers, such as nautilus. They actually implement the the behavior I mentioned. For myself, I often use "back" instead of "go up" to switch to the parent folder. Well it's a matter of taste. However just storing the old path seems to be a good and simple way to implement this feature.

tsujan:
> You can try other file managers, such as nautilus...

Please read the description I wrote for this commit!

> For myself, I often use "back" instead of "go up" to switch to the parent folder...

That's just a workaround and doesn't work in most cases: the user should have come there from a parent folder and should remember that to use "back".

> However just storing the old path seems to be a good and simple way to implement this feature.

Yes! I vaguely remember that I'd thought about it before using the current method but couldn't implement it (easily?). It would be really great if could.

PCMan:
OK. It takes too long for this issue.
Let's merge it for now. Later if I have time, I'll try to improve it further.
Thanks!

PCMan:
Well, just did some quick hack and now it works in all of the cases.
FYI:  https://github.com/lxde/pcmanfm-qt/commit/513ed53c90e8cb4ef8099bbc304ae8bc60ff60ad

tsujan:
Great!

PR #223, opened 2015-09-04
Select new files when they are created.

tsujan:
When files/folders are created (pasted, extracted, etc.) in the currently open folder, this commit selects them and scrolls to the first one. If there's only one file, it'll have focus too.

Structure: A signal containing info on added files (filesAdded) is emitted by FolderModel and connected/disconnected to/from a selecting slot (onFilesAdded) by TabPage appropriately. That slot uses FolderView::selectFiles.

Dolphin does a similar thing. With folders that contain many files, this has two practical benefits, at least:
* The user sees where pasted/extracted/new files go and can find them easily for later work or copy/cut/delete all of them.
* When he creates a new file/folder, he can immediately open it by pressing Enter.

It isn't for Desktop (yet).

P.S. (1) Please ignore the commits of #221 and #222 included by GitHub. (2) Here I also fixed a related issue, namely dropping on file, in contrast to folder (see FolderModel::dropMimeData).

### DIFF
--- a/libfm-qt/foldermodel.cpp
+++ b/libfm-qt/foldermodel.cpp
@@ -103,8 +103,10 @@ void FolderModel::onFilesAdded(FmFolder* folder, GSList* files, gpointer user_da
   FolderModel* model = static_cast<FolderModel*>(user_data);
   int n_files = g_slist_length(files);
   model->beginInsertRows(QModelIndex(), model->items.count(), model->items.count() + n_files - 1);
+  FmFileInfoList* infoList = fm_file_info_list_new();
   for(GSList* l = files; l; l = l->next) {
     FmFileInfo* info = FM_FILE_INFO(l->data);
+    fm_file_info_list_push_tail(infoList, info);
     FolderModelItem item(info);
 /*
     if(fm_file_info_is_hidden(info)) {
@@ -115,6 +117,7 @@ void FolderModel::onFilesAdded(FmFolder* folder, GSList* files, gpointer user_da
     model->items.append(item);
   }
   model->endInsertRows();
+  Q_EMIT model->filesAdded(infoList);
 }
 
 //static
@@ -389,7 +392,10 @@ bool FolderModel::dropMimeData(const QMimeData* data, Qt::DropAction action, int
       info = fileInfoFromIndex(itemIndex);
     }
     if(info)
-      destPath = fm_file_info_get_path(info);
+      if (fm_file_info_is_dir(info))
+        destPath = fm_file_info_get_path(info);
+      else
+        destPath = path();
     else
       return false;
   }
--- a/libfm-qt/foldermodel.h
+++ b/libfm-qt/foldermodel.h
@@ -88,6 +88,7 @@ public:
 
 Q_SIGNALS:
   void thumbnailLoaded(const QModelIndex& index, int size);
+  void filesAdded(FmFileInfoList* infoList);
 
 public Q_SLOTS:
   void updateIcons();
--- a/libfm-qt/folderview.cpp
+++ b/libfm-qt/folderview.cpp
@@ -712,6 +712,18 @@ FmPathList* FolderView::selectedFilePaths() const {
   return NULL;
 }
 
+QModelIndex FolderView::indexFromFolderPath(FmPath* folderPath) const {
+  QModelIndex index;
+  int count = model_->rowCount();
+  for(int row = 0; row < count; ++row) {
+    index = model_->index(row, 0);
+    FmFileInfo* info = model_->fileInfoFromIndex(index);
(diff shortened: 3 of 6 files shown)


paulolieuthier:
@tsujan I would like to give you some tips to improve your skills on Git. The commits d794341007958204981d436dea029295803b4c43 and 08e15ede00c48433a3ec94e7f007bf899196eea0 were not simply included by Github. You pushed them to this branch.

Whenever you create a branch, git will base it on the current branch, unless you specify another branch. So everytime you want to create a branch for a new PR, first checkout master, and then create the new branch:

```shell
$ git branch
  master
* new-feature
$ git push origin new-feature # do some hacking and commit to your fork
$ git checkout master # checkout master
Switched to branch 'master'
$ git checkout -b another-feature
Switched to a new branch 'another-feature'
```

That way, every contribution will be based on master, and independent of the others. But now, how to fix your branches? Looks like `selected_created_files` was based on `select_on_go_up`, which was based on `context_menu_paste`. Well, the best way I can think of is editing the branch history. For this branch, it would be as follows:

```shell
$ git branch
  context_menu_paste
  master
* selected_created_files
  select_on_go_up
$ git rebase -i HEAD~4 # HEAD~4 here indicates we want to edit the four last commits
```

That will open your editor, with something like that:

```
pick e2f7c2f libfm-qt: updated german translation
pick 08e15ed Paste into folder from its context menu.
pick d794341 Select the folder from where we have gone up.
pick 6d95c0f Select new files when they are created.
```

Of course, the first commit listed shouldn't be edited, so simply ignore it. The two next commits should be discarded for this PR. So delete those lines. Save the file and close the editor. Git will now try to apply the changes, and it'll fail. There are conflicts in some files. But don't worry, they can be easily resolved. Open every file marked as in conflict and correct the conflicts marked by Git. Then add then (using `git add file`) and run `git rebase --continue`. Git will then confirm the commit message and your branch is now clean. Force-push it to Github using `git push origin select_created_files -f` and come back to this page to check if it worked.

Did it work? Or did it not? Feel free to ask questions. You can always find us on our IRC channel on Freenode: #lxde.

tsujan:
Thank you very much for taking time and writing down this thorough explanation! It'll be surely helpful and I really appreciate it. However it did't work in this case: no conflict was specified by git; just this general message:

    error: could not apply 6d95c0f... Select new files when they are created.
    When you have resolved this problem...
    ...

This is the dilemma:

My four commits provided what I really missed when I came to LXQT from KDE, three weeks ago. With them, LXQT feels cozier than KDE. But the time I've spent on dealing with github PR is comparable to the total time I spent on these changes :(

I used github just for receiving bug reports and feature requests. Now I see there's something cryptic about it. Documentation is scattered, inconsistent and unclear. It may take me some time to know the right commands and methods for PR.

The problem is that I made four changes one after the other. Only the first one (in libfm) is independent from the others. The other three have files and also methods/functions in common. I can't separate them from each other. One solution that came to me was PRing one by one and waiting to see if each of them would be accepted or not. However, I wanted to spend my free time on solving another problem, namely that of making a filterbar for pcmanfm-qt. A single PR seemed out of question too.

tsujan:
Oh, I forgot to say that for branching I'd already done as you described. All branches share the same source files, so if, for example, I wanted the first changes not to be reflected in the second branch, I should revert them before going to it. The result would be diff files that couldn't be applied consecutively because of common files/methods.

tsujan:
The filterbar is ready:
https://dl.dropboxusercontent.com/u/7793460/pcmanfm-qt_filtering.mkv
I'm waiting for the destiny of my remaining four PRs to give a clean PR for it.

palinek:
> The filterbar is ready:
https://dl.dropboxusercontent.com/u/7793460/pcmanfm-qt_filtering.mkv

..seems nice

PCMan:
The "selecting newly created files" part still requires more review.
From the initial code review, the approach used here might not be reliable.
However, the filter bar looks great!
Would you please make it another separate PR?
Thanks!

tsujan:
> Would you please make it another separate PR?

I will soon. Thanks for reviewing my PRs!

tsujan:
> From the initial code review, the approach used here might not be reliable.

Could you please be more specific (as you were at #222)?

PCMan:
This PR has some problems.
1. It's outdated (still mixed with the wentUp_ thing.)
2. It handles all newly added files regardless of whether they are created by the file manager or not. This is not the desired behavior.
Ideally, only files that are created inside the file manager should be selected after their creation.
So, I appreciate your effort, but I'm not able to merge the PR now.
FYI, I'm going to split this repo into two parts and libfm-qt will be moved to its own repo soon. So it's better to start working on new PRs after that.
Cheers!

PR #225, opened 2015-09-08
Fix (workaround) a crash in placesview.

tsujan:
If the second column (column 1, to the right) of an item in Places side-pane is middle clicked accidentally, the app will crash. The cause is that "path" obtained in PlacesView::activateRow on the basis of column 1 will be so "invalid" that no FmPath function could deal with it. Apparently there's no way to check an FmPath for its "validity" but a simple conditional can prevent the crash.

### DIFF
--- a/libfm-qt/placesview.cpp
+++ b/libfm-qt/placesview.cpp
@@ -105,6 +105,10 @@ void PlacesView::activateRow(int type, const QModelIndex& index) {
 
 // mouse button pressed
 void PlacesView::onPressed(const QModelIndex& index) {
+  /* FIXME: (Tsu Jan) If column 1 is middle clicked by chance, the app
+     will crash because of an "invalid" FmPath. I found no way to check
+     an FmPath for its "validity", hence the following conditional. */
+  if(index.column() > 0) return;
   // if middle button is pressed
   if(QGuiApplication::mouseButtons() & Qt::MiddleButton) {
     activateRow(1, index);


palinek:
...I can't to reproduce the crash based on the description here (I'm a bit confused where to click).

tsujan:
That's why the crash wasn't reported ;) The Places view of the sidepane has two columns. Middle click on the rightmost side of a bookmark item. You might need to pull the splitter handle to right.

I saw the crash by chance when I started to use pcmanfm-qt for the first time but didn't know how it happened. Yesterday I succeeded in reproducing it and used gdb to find its cause. The conditional could be seen as a safety measure even when the root of the problem is fixed (in libfm?).

palinek:
...got it... crashing :)

palinek:
Just tested... but with your patch, the new tab isn't spawned when middle-clicked on the "column 1" position. The following works IMO bettter:
```diff
diff --git a/libfm-qt/placesview.cpp b/libfm-qt/placesview.cpp
index 0d82888..fc2f99a 100644
--- a/libfm-qt/placesview.cpp
+++ b/libfm-qt/placesview.cpp
@@ -107,7 +107,8 @@ void PlacesView::activateRow(int type, const QModelIndex& index) {
 void PlacesView::onPressed(const QModelIndex& index) {
   // if middle button is pressed
   if(QGuiApplication::mouseButtons() & Qt::MiddleButton) {
-    activateRow(1, index);
+    // the real item is at column 0
+    activateRow(1, 0 == index.column() ? index : index.sibling(index.row(), 0));
   }
 }
```

tsujan:
Yes! Your solution is better. I just tested it.

palinek:
@tsujan will you change your commit based on the provided diff or should I commit it?

tsujan:
Please commit your diff and close this PR. I'll delete extra branches of my fork anyway.

PR #227, opened 2015-09-08
Created italian desktop files, updated translation

stefonarch:


pmattern:
Could you please complement the two desktop entry files with Italian translations of key `GenericName`?

Regarding `pcmanfm/translations/pcmanfm-qt_it.ts` QtLinguist is complaining about missing plural of `%n item(s)` in context `PCManFM::TabPage`. It's not intended that both singular and plural are stated under singular here.

Other than that GTM.

stefonarch:
I missed the singular/plural split.... thanks.

pmattern:
Landed in 0ff66bf. Squashed the commits into one.
Thanks.

PR #228, opened 2015-09-08
Fix (workaround) for right-click crash in placesview.

tsujan:
Like what palinek did for the middle-click crash but now for the right-click menu on the second column and then clicking on a menu-item.
The real solution should be somewhere else but these two fixes will suffice for now.


PCMan:
merged.

PR #229, opened 2015-09-12
Added a filterbar + Handle virtually hidden files

tsujan:
This commit consists of two parts that, contrary to appearances, are closely related to each other:

(1) The filterbar is tab-aware but is cleared on changing directory in the same tab. Although it's a fixed part of the GUI, this can be changed later if needed.

(2) Making files "virtually" hidden is a feature that Nemo/Nautilus have and is requested for Dolphin (but ignored by KDE devs) for a long time. The user may want to hide a file without changing or being able to change its name. "lost+found" is a good example.

The commit simply uses the ADT "ProxyFolderModelFilter", which PCMan has included in "proxyfoldermodel.h" for adding extra filters. For virtually hidden files, it reads the file ".hidden" inside the current folder once and hides the listed files from the tabpage but, intentionally, not from the sidepane tree. This exception aside, the virtually hidden files behave like the usual hidden files.

P.S. Because I had applied my commits consecutively for my personal use, I had to "extract" this one manually. Therefore, although I've used this feature for a week, I tested this particular commit only a few times.

### DIFF
--- a/libfm-qt/proxyfoldermodel.cpp
+++ b/libfm-qt/proxyfoldermodel.cpp
@@ -254,6 +254,10 @@ void ProxyFolderModel::removeFilter(ProxyFolderModelFilter* filter) {
   Q_EMIT sortFilterChanged();
 }
 
+void ProxyFolderModel::updateFilters() {
+  invalidate();
+  Q_EMIT sortFilterChanged();
+}
 
 #if 0
 void ProxyFolderModel::reloadAllThumbnails() {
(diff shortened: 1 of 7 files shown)


PCMan:
LGTM.
Would you please squash these two commits into one?
Thank you.

tsujan:
>Would you please squash these two commits into one?

If I can pick the correct git commands out of its jungle... Doesn't clicking on "Files changed 7" above give a unified diff?

BTW, now that you're here, please take a look into #222 and #223 again if you have time. Both are reliable.

paulolieuthier:
@tsujan you squashed both commits, which is good. But now the commit has a bad description. Please amend it with a descriptive title. I can do it, but you should learn it.

tsujan:
@paulolieuthier Ok, I'll search in Git's jungle again ;)  -- I can't call it "learning" though.

paulolieuthier:
Merged in 5fd3eb96aba40130bcad50f092fb6d6717402bc2.

PR #230, opened 2015-09-15
Improve the column width computation for the detailed view

npmiller:
- Filename column is now expanded to fill extra available width

It means that now the detail columns are aligned on the right, and that the `QHeaderView` actually fills whole of the available space instead of leaving an empty space on the right as it currently does.

- Take the filename's sizeHint into account for its minimum width

Without this pull request, when shrinking a window too much in a directory where all the file names are small, the filename column *grows* to `200` instead of staying at the *smaller* necessary size.

- Stretch the last column

With this options, we make sure that the header widget always fills whole the available space, even when the user is manually resizing the columns (in this case the custom layout rules are not applied).

### DIFF
--- a/libfm-qt/folderview.cpp
+++ b/libfm-qt/folderview.cpp
@@ -181,7 +181,7 @@ FolderViewTreeView::FolderViewTreeView(QWidget* parent):
   doingLayout_(false),
   activationAllowed_(true) {
 
-  header()->setStretchLastSection(false);
+  header()->setStretchLastSection(true);
   setIndentation(0);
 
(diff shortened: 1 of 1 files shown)


excalibur1234:
this pull request sounds really useful.
the exact same thing is bugging me for a while, too.

thanks a lot @npmiller for improving pcmanfm-qt for everyone!

jleclanche:
@PCMan Could you give this a review asap? eg should it land in 0.10 (due next week)?

PCMan:
Merged. Thanks!

PR #232, opened 2015-09-18
Mousenavigation

qsodev:
fixed url update on mouse back/forward buttons by forwarding signals to MainWindow

PR #233, opened 2015-09-21
mouse navigation url update fix

qsodev:
fixed url update during mouse navigation. 
replaces #232 with clean commit :)


palinek:
1. Why did you close PR #233 and opened a new one?
2. In this PR there isn't the `void FolderView::childMousePressEvent(QMouseEvent* event)` change (back/forward), is that intentional?

qsodev:
i closed it and made a new one because the history was messed up.
The changes are the same as in #232. I have never changed FolderView::childMousePressEvent(QMouseEvent* event). It is not needed.

PCMan:
Seems that it's a fix for https://github.com/lxde/pcmanfm-qt/commit/b47bd9f46ccb7281f11b9f20040c6741dd2421ac .

PCMan:
merged into master.

PR #234, opened 2015-09-23
Desktop folder

luis-pereira:
Should go with https://github.com/lxde/lxqt-session/pull/39
Discussion [here](https://github.com/lxde/lxqt/issues/780)

@PCMan This is your call.

### DIFF
--- a/pcmanfm/CMakeLists.txt
+++ b/pcmanfm/CMakeLists.txt
@@ -47,6 +47,7 @@ set(pcmanfm_UIS
     about.ui
     preferences.ui
     desktop-preferences.ui
+    desktop-folder.ui
     autorun.ui
 )
 
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -49,6 +49,9 @@
 
 #include <X11/Xlib.h>
 
+#include "xdgdir.h"
+#include <QFileSystemWatcher>
+
 using namespace PCManFM;
 static const char* serviceName = "org.pcmanfm.PCManFM";
 static const char* ifaceName = "org.pcmanfm.Application";
@@ -70,6 +73,8 @@ Application::Application(int& argc, char** argv):
   enableDesktopManager_(false),
   preferencesDialog_(),
   volumeMonitor_(NULL),
+  userDirsWatcher_(NULL),
+  lxqtRunning_(false),
   editBookmarksialog_() {
 
   argc_ = argc;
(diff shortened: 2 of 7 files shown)


PCMan:
Really thank you for the hard work. The change looks pretty good.
One little thing, though. Instead of turning this off at compile time, would it be better to enable the desktop folder editing feature at runtime? For example, it's easy to check the existence of lxqt-session dbus service, so we can know whether we're  running in LXQt session and need to hide the option. Otherwise, show the desktop folder editing option to the user. What do you think?

luis-pereira:
@PCMan I considered that aproach. I choosed this one for size optimization.

In an somehow related issue: What's your opinion about splitting pcmanfm-qt into two:
* pcmanfm-qt 
* desktop manager

luis-pereira:
@PCMan that would tie to LXQt, have you considered that ?

jleclanche:
@luis-pereira There's quite a bit of an advantage to have the desktop being managed by the file manager. Both windows and gnome do it this way.

jleclanche:
@luis-pereira so I'm not a fan of this approach at all. As I said before, it's OK to have two places where this stuff is configured - however of course they should use the *same* method for configuring it.

323baa58b4ee1307dc76cabc891cf94958106857 can be landed right away btw.

luis-pereira:
@jleclanche There's no point in landing https://github.com/lxde/pcmanfm-qt/commit/323baa58b4ee1307dc76cabc891cf94958106857 without the rest. What advantages do you refer to ?

By detecting LXQt session we tie pcmanfm-qt to LXQt. I will update the code.

luis-pereira:
@jleclanche @PCMan Updated

PCMan:
@luis-pereira  Sorry for the late reply. I don't have time for coding this month. :-(
About the idea of splitting the desktop manager part from the file manager, I'd say it might be better to leave it as it is. If you split them into two processes, the folder caches, icon caches, and other caches, which are heavily used, cannot be shared among them. This hurts both performance and memory usage. Besides, opening a folder is such a frequently done operation for a normal GUI user, so it's good to keep running the file manager in the background for fast startup. Since the desktop manager is running all the time, it does the job and we don't need a separate daemon. Moreover, the desktop manager is just a specialized folder view of pcmanfm-qt with its background painted with the wallpaper. So this achieve maximized reuse of existing code.
Give the merits mentioned above, I'd suggest leave it as it is and do not do the split.

luis-pereira:
@PCMan I was thinking in people that don't uses pcmanfm-qt file manager. For the ones who use it, I do agree with you.
This PR now detected LXQt and disables the Edit Desktop Folder and enables a watcher for changes.

PCMan:
@luis-pereira: Then it's possible to split it into two modules, which are loaded into the same process. For those who only want the desktop manager, they can install that module only. This does not save much, though. The desktop manager is just a specialized file manager window without window decoration and menu. For those who don't need desktop icons, it's safe to uninstall pcmanfm-qt and use whatever tool they want to set a nice wallpaper.
BTW, I'm currently handling this PR. Thank you for the effort.
I'd merge it soon.

PCMan:
Merged. Thanks!

PR #246, opened 2015-11-01
Upgrade of pcmanfm-qt to C++11

rbazaud:
Use ‘nullptr’ instead of ‘NULL’

        Safer and semantically more correct,
      	    in C: #define NULL ((void *)0)
            in C++ < C++11: #define NULL 0

    Use of C++11 features when it improved readability/legibility

    Minor optimizations

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,24 @@ include(LXQtTranslateDesktop)
 
 add_definitions(-DQT_NO_KEYWORDS)
 
+if (CMAKE_VERSION VERSION_LESS "3.1")
+    include(CheckCXXCompilerFlag)
+    CHECK_CXX_COMPILER_FLAG("-std=c++11" COMPILER_SUPPORTS_CXX11)
+    if(COMPILER_SUPPORTS_CXX11)
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+    else()
+        CHECK_CXX_COMPILER_FLAG("-std=c++0x" COMPILER_SUPPORTS_CXX0X)
+        # -std=c++0x is deprecated but some tools e.g. qmake or older gcc are still using it
+        if(COMPILER_SUPPORTS_CXX0X)
+            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x")
+        else()
+            message(FATAL_ERROR "Compiler ${CMAKE_CXX_COMPILER} does not support c++11/c++0x")
+        endif()
+    endif()
+else()
+  set(CMAKE_CXX_STANDARD 11)
+endif()
+
 if (CMAKE_COMPILER_IS_GNUCXX)
     # set visibility to hidden to hide symbols, unless they're exported manually in the code
     set(CMAKE_CXX_FLAGS "-fvisibility=hidden -fvisibility-inlines-hidden -fno-exceptions ${CMAKE_CXX_FLAGS}")
--- a/libfm-qt/appmenuview_p.h
+++ b/libfm-qt/appmenuview_p.h
@@ -35,7 +35,7 @@ public:
     if(menu_cache_item_get_icon(item))
       fmicon = fm_icon_from_name(menu_cache_item_get_icon(item));
     else
-      fmicon = NULL;
+      fmicon = nullptr;
     setText(QString::fromUtf8(menu_cache_item_get_name(item)));
     setEditable(false);
     setDragEnabled(false);
--- a/libfm-qt/dirtreemodelitem.cpp
+++ b/libfm-qt/dirtreemodelitem.cpp
@@ -26,24 +26,24 @@
 namespace Fm {
 
 DirTreeModelItem::DirTreeModelItem():
-  model_(NULL),
-  folder_(NULL),
+  model_(nullptr),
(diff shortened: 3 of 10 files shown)


PCMan:
Please avoid using the new range based for of C++11 with Qt containers.
This might cause implicit copy of the whole list sometimes because of the design of Qt.
So care must be taken when using it.
http://www.dvratil.cz/2015/06/qt-containers-and-c11-range-based-loops/
For me, this can be cumbersome so I'd rather avoid using it since Q_FOREACH does exactly the same thing.

jleclanche:
I think, with @PCMan's comments fixed, this can land after the release.

rbazaud:
I totally agree with the fact that care must be taken, _always_, however the article relies on a specific implementation of Qt containers however they are bound to change from  one release to the other, without any warning. The article is interesting , but to prove his point the author quotes a benchmark done by the Qt team, I’d put more faith in a benchmark done on this specific project, with its specific use cases rather than in a “textbook case”. AFAIK the Qt framework put the emphasis on memory consumption before algorithmic efficiency. I’m always eager to learn and I must admit that I’m much more experienced with the STL containers and algorithms than with their Qt framework counterparts.

New C++ constructs might be cumbersome in a few cases indeed, which is why I haven’t used them everywhere. As I’ve written, I used them when, and only when it was possible to write the very same algorithm (sometimes with a few less instructions) while being more readable, which makes for an easier maintainability. As for `Q_FOREACH` and “range-based for loops” they may be “functionally identical” but when one take a look at `Q_FOREACH` implementation, “efficient” isn’t what comes to mind, and the compilation time can hardly be equivalent. Moreover the use of an approved C++ standard is “future-proof”, I mean before GTK was used now, it’s Qt and in a few years maybe something else in which case if both are globally equivalent, `Q_FOREACH` is at a disadvantage. And I’m not even talking about Qt containers (I mean there is no real choice here since Qt methods expect Qt containers)…

Isn’t it one potential effect of const iterators, as stated in the article, to avoid a deep copy ? In other words, in the case you’re not interested in having both ways instrumented, should I revert every use of C++11 construct or as advocated in the article, whose you referred me to, only each construct who doesn’t iterate only for reading (by using a const type) ?

PCMan:
Actually we like to use C++ 11 and already used it in some parts of LXQt.
Your patch looks good. Thank you.
I personally do not dig into the implementation details of Qt containers too much.
So I'm not sure in what cases deep copy is performed.
I'm not aware of the issue until a KDE developer remind me of this.
This potential deep-copy problem is just a friendly reminder.
If you know how to avoid that, feel free to use range based for and we'll happily accept the patch.
While C++ 11 is standard, Qt needs to be cross-platform and backward compatible, so they use their own containers. I really prefer STL ones more, but I can understand their concern.
At the moment, we need to live with Qt containers since all Qt classes use them.

rbazaud:
@PCMan Yes, it’s actually what gave me the “idea” :smile: I’m quite new to the project, which is why I was looking at the code of several LXQt modules. Thank you for your understanding, as for the information. I’ve reverted (and smashed the commit) for the cases where I couldn’t honestly say that the change would be beneficial without instrumentation and/or an experiment.

@jleclanche I wasn’t aware of the release to come (soon ? is it the 0x0.A or the 1.0 ?)

jleclanche:
@rbazaud Tomorrow we're releasing 0.10.

rbazaud:
Ok, I see, is there something I could help with or is it “code freeze” ?

jleclanche:
We don't have enough contributions to do meaningful code freezes. The PR will just hang for a day more.

rbazaud:
Ok, that was mainly out of curiosity, as far as I’m concerned my pull request could hang for a week or a month, I mean it’s not my decision to make. And my proposition for help was not related to the PR…

jleclanche:
@rbazaud Ah, I'm off to bed right now but swing by tomorrow on `#lxde` on freenode if you're like to contribute, or you can email me directly (see my profile). I'll guide you about.

PCMan:
Sorry for the late reply. Too busy this month.
There is only one question remained. Is there any reason to replace
"namespace Fm { ... }" with "using namespace Fm;"?
Though it looks ugly, the first one is more frequently seen in large projects so I think this might be better practice as it pollute the namespace less. What do you think?

rbazaud:
No trouble, I figure you must be often solicited and «overbooked.» being the project lead. And no technical motivation to `using namespace xxx;` it's partly because it's a bit lighter and also by habit. Most of the projects I've worked on were always using `namespace xxx { … }` for the headers and reserved `using namespace xxx;` to the `*.cpp`. Since I don't have any motivated objection to oppose, I'll revert back on this.

PCMan:
Thanks for the patience.
I used "using" a lot in the past, until I read this.
https://google.github.io/styleguide/cppguide.html#Namespaces
Indeed large projects from Google such as Android and Chromium follow this suggestion.
Google did not provide good explanation about the reason behind this, but I guess it's good practice anyways. So recently I tried to gradually adopt similar coding style.

rbazaud:
Very interesting, sorry to have polluted the code :disappointed: I'm fixing it right now. I think I'll be using the «using» directive much less in the future… I agree, Google's guidelines or even Google's way of doing software cannot be bad.

PCMan:
merged. Thanks!

PR #248, opened 2015-11-03
Tab context menu for closing other tabs

tsujan:
This is a right-click menu, only for tabs and only when there is more than one tab, for closing the left, right or other tabs, depending on the the right clicked tab, which can be the current one or not. It comes in handy when there are several tabs.

### DIFF
--- a/pcmanfm/main-win.ui
+++ b/pcmanfm/main-win.ui
@@ -721,6 +721,27 @@
     <string>Filter</string>
    </property>
   </action>
+  <action name="actionCloseLeft">
+   <property name="icon">
+    <iconset theme="go-previous"/>
+   </property>
+   <property name="text">
+    <string>Close &amp;left tabs</string>
+   </property>
+  </action>
+  <action name="actionCloseRight">
+   <property name="icon">
+    <iconset theme="go-next"/>
+   </property>
+   <property name="text">
+    <string>Close &amp;right tabs</string>
+   </property>
+  </action>
+  <action name="actionCloseOther">
+   <property name="text">
+    <string>Close &amp;other tabs</string>
+   </property>
+  </action>
  </widget>
  <customwidgets>
   <customwidget>
--- a/pcmanfm/mainwindow.cpp
+++ b/pcmanfm/mainwindow.cpp
@@ -51,7 +51,8 @@ namespace PCManFM {
 
 MainWindow::MainWindow(FmPath* path):
   QMainWindow(),
-  fileLauncher_(this) {
+  fileLauncher_(this),
+  rightClickIndex(-1) {
 
   Settings& settings = static_cast<Application*>(qApp)->settings();
   setAttribute(Qt::WA_DeleteOnClose);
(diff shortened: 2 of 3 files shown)


jleclanche:
"Close all other tabs" is enough for a file manager imo, don't need the
other ones.

J. Leclanche

tsujan:
If it weren't for several occasions that I wanted to close left or right tabs, I wouldn't add them. The user may arrange tabs in a specific order.

pmattern:
Personally I didn't miss a feature like `Close {left,right} tabs` before but can very well figure it may be useful when tabs were kind of grouped. Given the way it's accessed I can't figure disadvantages like invoking by incident.
Technically it's working as expected.
Should better be `Close *all* other tabs`, though.

Except the minor naming change GTM as is for my part, considering it a possibly useful feature that doesn't have a downside.

tsujan:
@pmattern  Thanks a lot! Your testing is valuable to me.

> Should better be Close *all* other tabs, though.

Firefox and Chromium both use "Close other tabs" ;) Being the last item and coming below a separator, it can be picked quickly and easily.

pmattern:
> Firefox and Chromium both use "Close other tabs" ;)

Yeah. But with regards to closing both are covering tabs right of the actual only for whatever reason. Personally I still think "all" would be better. It's a matter of taste, of course.
> Being the last item and coming below a separator, it can be picked quickly and easily.

Sure. But that's independent from the wording and I didn't doubt this.

tsujan:
> Yeah. But with regards to closing both are covering tabs right of the actual only for whatever reason.

I didn't understand that sentence.
Perfectionism aside, let me logically explain the menu behavior to those who haven't tried it. Supposing that the user has right clicked on an active or an inactive tab:

(1) If there's another tab to its left,  "Close left tabs" will be displayed. It's always the first item if present and is *the only one* when there's no tab to the right.
(2) If there's another tab to its right,  "Close right tabs" will be displayed. It'll be the second item if present and if "Close left tabs" already exists. It's *the only item* when there's no tab to the left.
(3) If there are tabs both to the right and to the left, a separator will come after "Close right tabs" and after it, "Close other tabs" will be shown.

So, there's no room for any mistake by a drunken user either ;) For several years, this arrangement has worked for me with a personal text editor I use for coding even when I'm very tired -- but not drunken, of course!

pmattern:
>> Yeah. But with regards to closing both are covering tabs right of the actual only for whatever reason.

>I didn't understand that sentence.

While your proposal allows for closing tabs both left and right of the current one
pcmanfm-qt_close-distinct-tabs_pcman
Firefox and Chrome only allow for closing those right of the current one for whatever reason
pcmanfm-qt_close-distinct-tabs_firefox
And due to this difference I for one think adding "all" makes the action a bit more clear somehow.
So just a minor detail.

Aside from this I think I understood your concept very well and am in favour of if, as said above.

PCMan:
Merged into https://github.com/lxde/pcmanfm-qt/commit/8106aa787fede3df59bce43a01a56da7a33bc268 .
Thanks!

PR #249, opened 2015-11-03
Fix compiling with Qt < 5.4

philippwiesemann:
The global qUtf8Printable() is not available in versions before Qt 5.4.

Thanks!


luis-pereira:
Thanks @philippwiesemann

jleclanche:
erm, we depend on qt 5.4...

philippwiesemann:
The CMakeLists file currently requires Qt 5.2 only.

jleclanche:
That needs to be changed then, we're bumping up to 5.4 for this release.

PCMan:
There's no need to change that at the moment. We haven't use Qt 5.4 specific features.

Issue #251, opened 2015-11-04
Double selection of files


baghira:
Upon clicking on some files, pcmanfm-qt also selects either the file immediately above or below. I suspect the issue appears only with file with longish names (i.e. 50-60 characters).
Moreover the issue disappears once I resize the window (until, that is, the next time I open the folder).
![Sample screenshot](https://cloud.githubusercontent.com/assets/15658924/10950691/4cf99ef4-8339-11e5-93f6-5a6eddcf82e0.png)

tsujan:
I confirm that -- saw it only recently.

pmattern:
Probably related to lxde/lxqt#768?

tsujan:
Fixed by  #252.

Issue #97, opened 2014-10-10
Align icons to grid on desktop


paulolieuthier:
There should be an option for that.

PCMan:
This will change the UI. Leave this for 1.0.

pmattern:
Issue was fixed in the wake of lxde/pcmanfm-qt#252.

As discussed in this PR a remnant that may eventually still need to be addressed is a missing means to adjust the distance of objects on the desktop (that is to adjust the grid which is used right now).
But this is already addressed in lxde/lxqt#786.

pmattern:
Using a grid was introduced *as default behaviour* in lxde/pcmanfm-qt#252 solving several issues like multiple selection by clicking single objects.

So strictly speaking (and unlike I thought when closing) this issue is not fixed yet if aligning to a grid should be *optional*.

On a side note a related finding that eventually still needs to get addressed is a missing means to adjust the distance of objects on the desktop (that is to adjust the grid which is used right now).
This is already addressed in lxde/lxqt#786.

tsujan:
@pmattern 
I can't say that unaligned desktop icons are impossible to manage but, their ugliness aside, do they have any advantage?

pmattern:
@tsujan 
I'm not promoting anything here and personally I'd just be fine with a grid always being used.
Rather, I just wanted to briefly mention the impact of lxde/pcmanfm-qt#252 on this issue - and then to explain why closing was eventually premature.

tsujan:
@pmattern 
OK! I just thought there might be an advantage I didn't know of.

paulolieuthier:
I think it's safe to close. At least for me, unaligned icons are messy. Anyway, I think there is a issue with the grid on desktop: the grid does not stretch 100% vertically, and a gap is created in the bottom where you can't drag a icon to. This can be easy tested by using `Xephyr :1 -screen 800x600` to create a screen and then `DISPLAY=:1 pcmanfm-qt --desktop` to test pcmanfm-qt on it.

@tsujan can you reproduce it?

tsujan:
@paulolieuthier 
If the gap's height is less than the grid cell height, it's OK. The cell height is a function of font/icon size and the screen height may not be an exact multiple of it. If not, please show a screenshot of the problem!

paulolieuthier:
> If the gap's height is less than the grid cell height, it's OK.

[...]

PR #252, opened 2015-11-05
Fixed several problems with item selection and alignment

tsujan:
Fixes lxde/lxqt#768 (or lxde/pcmanfm-qt#251). In FolderItemDelegate::drawText(), the height of the returned rect shouldn't be that of the created layout because an extra line is added to the latter for calculating the height -- a line that won't be drawn. In fact, a maximum of 3 lines is planned in FolderView::updateGridSize(), while pcmanfm-qt showed 4 lines.

### DIFF
--- a/libfm-qt/folderitemdelegate.cpp
+++ b/libfm-qt/folderitemdelegate.cpp
@@ -54,12 +54,11 @@ QSize FolderItemDelegate::sizeHint(const QStyleOptionViewItem& option, const QMo
     opt.decorationAlignment = Qt::AlignHCenter|Qt::AlignTop;
     opt.displayAlignment = Qt::AlignTop|Qt::AlignHCenter;
 
-    // FIXME: there're some problems in this size hint calculation.
     Q_ASSERT(gridSize_ != QSize());
-    QRectF textRect(0, 0, gridSize_.width() - 4, gridSize_.height() - opt.decorationSize.height() - 4);
+    QRectF textRect(0, 0, gridSize_.width(), gridSize_.height() - opt.decorationSize.height());
     drawText(NULL, opt, textRect); // passing NULL for painter will calculate the bounding rect only.
-    int width = qMax((int)textRect.width(), opt.decorationSize.width()) + 4;
-    int height = opt.decorationSize.height() + textRect.height() + 4;
+    int width = qMax((int)textRect.width(), opt.decorationSize.width());
+    int height = opt.decorationSize.height() + textRect.height();
     return QSize(width, height);
   }
   return QStyledItemDelegate::sizeHint(option, index);
@@ -107,7 +106,11 @@ void FolderItemDelegate::paint(QPainter* painter, const QStyleOptionViewItem& op
       painter->drawPixmap(iconPos, symlinkIcon_.pixmap(opt.decorationSize / 2, iconMode));
 
     // draw the text
-    QRectF textRect(opt.rect.x(), opt.rect.y() + opt.decorationSize.height(), opt.rect.width(), opt.rect.height() - opt.decorationSize.height());
+    // The text rect dimensions should be exactly as they were in sizeHint()
+    QRectF textRect(opt.rect.x() - (gridSize_.width() - opt.rect.width()) / 2,
+                    opt.rect.y() + opt.decorationSize.height(),
+                    gridSize_.width(),
+                    gridSize_.height() - opt.decorationSize.height());
     drawText(painter, opt, textRect);
     painter->restore();
   }
@@ -141,6 +144,7 @@ void FolderItemDelegate::drawText(QPainter* painter, QStyleOptionViewItemV4& opt
   int visibleLines = 0;
   layout.beginLayout();
   QString elidedText;
+  textRect.adjust(2, 2, -2, -2); // a 2-px margin is considered at FolderView::updateGridSize()
   for(;;) {
     QTextLine line = layout.createLine();
     if(!line.isValid())
@@ -162,21 +166,25 @@ void FolderItemDelegate::drawText(QPainter* painter, QStyleOptionViewItemV4& opt
     ++ visibleLines;
   }
   layout.endLayout();
+  width = qMax(width, (qreal)opt.fontMetrics.width(elidedText));
 
   // draw background for selected item
   QRectF boundRect = layout.boundingRect();
   //qDebug() << "bound rect: " << boundRect << "width: " << width;
   boundRect.setWidth(width);
+  boundRect.setHeight(height);
   boundRect.moveTo(textRect.x() + (textRect.width() - width)/2, textRect.y());
 
+  QRectF selRect = boundRect.adjusted(-2, -2, 2, 2);
+
   if(!painter) { // no painter, calculate the bounding rect only
-    textRect = boundRect;
+    textRect = selRect;
     return;
   }
 
   QPalette::ColorGroup cg = opt.state & QStyle::State_Enabled ? QPalette::Normal : QPalette::Disabled;
   if(opt.state & QStyle::State_Selected) {
-    painter->fillRect(boundRect, opt.palette.highlight());
+    painter->fillRect(selRect, opt.palette.highlight());
     painter->setPen(opt.palette.color(cg, QPalette::HighlightedText));
   }
   else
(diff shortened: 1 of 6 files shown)


pmattern:
Just gave tsujan/pcmanfm-qt@26a9a754e676 a shot on Arch Linux x86_64 with all other from yesterday morning's VCS checkout: lxde/pcmanfm-qt#251 is fixed but lxde/lxqt#768 persists.

tsujan:
@pmattern 
Oh, I misunderstood lxde/lxqt#768! This is a fix just for lxde/pcmanfm-qt#251 (I thought 768 was identical to it).
I'll think about lxde/lxqt#768 tomorrow. It shouldn't be hard to fix.

tsujan:
The last commit should fix lxde/lxqt#768 (if I've understood it correctly this time -- the description wasn't clear). However, these are NOT for desktop. If you're satisfied with them, desktop could be included too (in this PR).

pmattern:
Hm, now I'm a bit confused.

Situation so far:
* lxde/lxqt#768: "Clicking items selects adjacent items in addition". Apparently written with desktop icons in mind. At least this is what I personally took for granted as I had faced the problem on the desktop only as well, and it doesn't seem to be contradicted by any of the other comments. Nevertheless you're absolutely right - the issue isn't clear at all.
* lxde/pcmanfm-qt#251: "Clicking files in the regular folder view (right pane) selects adjacent objects in addition". Solved by tsujan/pcmanfm-qt@26a9a75.

So far so good. But what I don't understand is
> However, these are NOT for desktop. 

So what else is the purpose of tsujan/pcmanfm-qt@9212f19? If I got it right the only problem that's left to get fixed is the one affecting the desktop described in lxde/lxqt#768. Also, I couldn't see any difference between tsujan/pcmanfm-qt@26a9a75 and tsujan/pcmanfm-qt@26a9a75 so far.

tsujan:
@pmattern 
> Hm, now I'm a bit confused...

Perhaps because I first focused on the same problems in icon views, not in desktop. Now the desktop is covered too.
The desktop isn't drawn by the same methods icon views are, although it could be (at least theoretically). So, the desktop part was a little tricky.

tsujan:
BTW, this is an example of text outside selection rects, that's fixed with the last commit:
example

tsujan:
@pmattern
After being confused about your confusion for a while, I'm sure this explanation will make everything clear. Let's forget about the issue reports for now.

There were three issues, each of which occurred both in icon view and desktop:

(1) Some items with long names overlapped, such that selecting one of them made others be selected too;
(2) The lack of space between items made it difficult to select one of them with rubber-band (left clicking on a empty space and dragging the cursor);
(3) Sometimes file names weren't completely inside selection rectangles.

All of them are fixed by these commits.

pmattern:
Regardless of the theory the problem of several desktop icons getting selected by clicking a single one persists in tsujan/pcmanfm-qt@cdfa309 if the icons are rather close to each other like in this screenshot.
pcmanfm-qt_single-click-selection
where clicking any of the three icons selects all at once.
Of course this is an insane placement that will hardly ever be chosen with intent. But it's something that may happen e. g. if a mouse key gets released too early by accident and hence it should be possible to deal with it reasonably. Btw. Otter Browser and Thunderbird still get both selected by clicking any of them after moving QupZilla out of the way so the problem is apparently not due to the text here.

In Xfce there seems to be a technique that's enforcing some minimum distance if two icons are placed to close. Maybe something similar could be implemented in PCamFM-Qt as well? Would possibly related placing along a grid which is discussed in lxde/pcmanfm-qt#97.

tsujan:
@pmattern
That's because **my fix is for aligned icons.** Nowhere in my comments, I told about *moved* icons.

You've manually moved those icons in the screenhot -- in other words, you have files with names desktop-items-X.conf in '~/.config/pcmanfm-qt/lxqt'. If lxde/lxqt#768 was about that (and according to what you explained, it was), the description was **really vague.** My desktop icons are always aligned to the grid because I don't move them. I saw the single selection issue with them and then fixed it.

Now, lxde/pcmanfm-qt#97 is yet another related issue to be solved (the 4th one in this series). In fact, I had it in my TODO list but apparently I should start to think about it now.

tsujan:
This time, I tried to address the problem(s) of desktop items with custom positions by *forcing* grid alignment on them. There were several unexpected issues, that I hope are fixed but, due to the complexity of the problem, there may be more issues, that could only be found by further tests.

@pmattern Your previous tests were very helpful. Would you please test it again whenever you have time?

pmattern:
First I've got another question.
Besides #97 is about aligning desktop objects to a grid I'm wondering whether some feature like this isn't already available since quite a while, at least partially.
I can remember sometimes objects didn't get placed where they were dropped but sort of jumped a bit away. Couldn't reproduce with 929f029 any more, though...

The issue addressed in lxde/lxqt#768 - as we understood it - seems to have gone with latest tsujan/pcmanfm-qt@e70f871.

This seems to be due to different behaviour when objects get moved.
Say one object gets moved towards another like Firefox towards QupZilla in this screenshot that's displaying the upper left corner of the desktop.
pcmanfm-qt_desktop_object-distance
As long as the distance between the two objects is beyond some threshold, the object that's moved gets placed right where it is dropped. If the distance is below the threshold the moved object gets thrown back to its old position. If the distance gets even smaller there's a dialog offering to copy or move and the likes.

There seem to two problems left:
The behaviour when one objects gets moved beyond another, e. g. Firefox left of QupZilla or QupZilla between Otter and the edge of the screen, seems inconsistent. Sometimes the moved object gets thrown back to its old position, sometimes the two objects change their position, both in a rather unreproducible manner.
The minimum distance at which two objects can get placed next to each other is rather high, at least if the actual icons are rather small. E. g. in the screenshot above it's not possible to get Thunderbird closer to Otter.

tsujan:
First thanks for testing :)

> I'm wondering whether some feature like this isn't already available since quite a while, at least partially.

If I've understood your question correctly, aligning to the grid after movement is provided by QListView::Snap. However, it wasn't enough alone. There were several complications: What about old config files, in which the coordinates aren't aligned? What if two moved items occupy the same location (that was possible before, although not easy to reproduce)? What if a new panel or a dock is introduced (moved items didn't change their locations accordingly)? etc.

> The behaviour when one objects gets moved beyond another seems inconsistent.

I agree. That's because we have two kinds of items: moved (stuck) and unmoved. That distinction was there before the commits. I think it should be possible to overcome it but this PR would be too inclusive with such an endeavor. Anyway, it isn't a usability issue, IMO: the user can exchange positions by moving some items to a distant location first.

> The minimum distance at which two objects can get placed next to each other is rather high

It depends on these things: max text width (~ 10 characters -- PCMan's choice); max text lines (3 lines -- PCMan's choice), a 16-px space for rubber-band selection (my choice), and, of course, the icon size (user's choice). IMO, it's possible to provide the user with settings for all of them but, again, that would be beyond the scope of this PR.

I also saw another problem, that existed before these commits: if you try to put an item beyond the desktop boundaries, it will return to its original location but the whole desktop might be a little scrollable with mouse wheel afterward. It's hard to reproduce though. I don't think this PR is a good place for fixing it.

I'll use the commits to see if there's something seriously wrong. Feedbacks and clear suggestions or criticisms are very welcome.

pmattern:
IMHO  the situation is now as follows:
Your `fix_single_selection` branch is fixing both lxde/lxqt#768 and lxde/pcmanfm-qt#251. Regarding the placement of icons on the desktop some problems still exist, some are related to the changes in `fix_single_selection`, some aren't. But these problems do *not* outweigh the benefit of fixing lxde/lxqt#768.
So I'd suggest to merge this PR and deal with those remaining problems in an issue on its own.

With this in mind, GTM.

tsujan:
Thanks! Please wait just a few days! There may still be minor issues from the past that can be fixed in this PR. For example, I remembered that deleting a stuck item didn't remove it from the list, and I'm trying to fix that now.

tsujan:
I've finished my work with this PR by squashing its commits into one. Perfectionism aside, it's good to me. Please consider it for merging!

pmattern:
The only thing that *could* be discussed is whether the distance between items imposed by the grid is a bit large. Everything else is fine, a problem I had seen in the meantime involving misplacement of icons moved a rather long distance on the screen is fixed, too. No problems regarding preexisting icons either.

As said above, should definitely be merged, IMO.

tsujan:
@pmattern  Thank you!
The grid distance of Desktop is exactly that of forder-view, which is even a bit smaller than before. Of course, now, with Desktop grid alignment, Desktop items cannot be closer to each other than that but we could implement a separate grid size setting for Desktop in the future.

PCMan:
Merged!
Thank you for the hard work!

pmattern:
> Of course, now, with Desktop grid alignment, Desktop items cannot be closer to each other than that but we could implement a separate grid size setting for Desktop in the future.

For the sake of completeness: just saw there's already an issue addressing this lxde/lxqt#786.

PR #254, opened 2015-11-07
Updated *_.hu.ts files

attuska:


pmattern:
The modifications themselves are alright but there's something pretty much wrong with that merged pull request that mustn't make it into master.
Didn't manage to squash it myself here and don't know what exactly is the matter. You could try to solve this an a way similar to the one described [earlier](https://github.com/lxde/lxqt-panel/pull/274#issuecomment-154716675).

PR #256, opened 2015-11-10
hu translations updated

attuska:
I calmed down. You all have been right.
I hope that I have finally made no mistakes.
I'm tough.


palinek:
GTM

pmattern:
Agree.

So welcome back on board then.

PR #257, opened 2015-11-11
Improve Italian translation

smart2128:


PCMan:
Thank you. Is it possible for you to squash these three commits into one? So the commit log can be cleaner?

smart2128:
I'm not an experienced git user.
I tried squashing those commits but it seems things went from bad to worse.
I created a new fork and open a new pull request for it.
Thank you for your support.

PR #258, opened 2015-11-11
Fix --version option not displaying version

philippwiesemann:
No application version string was set for QCommandLineParser to display. It therefore displayed "pcmanfm-qt " [sic] instead of "pcmanfm-qt 0.10.0".

Thanks!


palinek:
GTM

paulolieuthier:
GTM.

PR #259, opened 2015-11-12
Force the root config dir on the root instance

tsujan:
Otherwise, root might take the ownership of the user config file when gksu-properties is set to su (lxde/lxqt#874).


PCMan:
Good point! I'm not aware of the issue. Just think that if there is a better alternative to hard coding home+"/.config".

PR #266, opened 2015-11-17
Update French translation of GPL license

rbazaud:
* Fix translation, using common parts with GPL v3 from [GNU website](https://www.gnu.org/licenses/translations.html) and [here](http://org.rodage.com/gpl-3.0.fr.txt)
* The line breaks have been deleted while keeping each stanza as they were, in order to get the wrapping done automatically by the text edit widget instead of manually, see screenshot below

screenshot_pcmanfm-qt_about

### DIFF
--- a/pcmanfm/translations/pcmanfm-qt_fr.ts
+++ b/pcmanfm/translations/pcmanfm-qt_fr.ts
@@ -21,7 +21,7 @@
     <message>
         <location filename="../../build/pcmanfm/ui_about.h" line="138"/>
         <source>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;&lt;a href=&quot;http://lxqt.org/&quot;&gt;&lt;span style=&quot; text-decoration: underline; color:#0000ff;&quot;&gt;http://lxqt.org/&lt;/span&gt;&lt;/a&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</source>
-        <translation type="unfinished"></translation>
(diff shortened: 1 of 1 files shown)


rbazaud:
And here is the result after:
screenshot_pcmanfm-qt_about_fr

rbazaud:
@pmattern the missing words in translation are those I couldn't find a suitable description for.

pmattern:
In line 32 its "and/or" in English. You dropped "and" intentionally in French?

Translations tagged `type="vanished"` are no longer needed due to changes in the actual code and kept for purposes like helping translate successor strings and the likes. Translation isn't needed, though.
Here, it's rather complicating the review.

With regards to the main question, lacking translations, the file isn't in sync with the source code as Qt Linguist cannot find several source files and thus isn't able to display those previews.
Unfortunately I can't tell what exactly is happening here as I don't know the procedure of updating the TS files well. (One of the numerous topics on my ToDo list only so far...)
But I think it would be better to try and get that stuff synced as this enables the previews in Qt Linguist which save a lot of work.

palinek:
FYI
- [lupdate](http://doc.qt.io/qt-5/linguist-manager.html)
- [build time update](https://github.com/lxde/liblxqt/commit/c8a513bd)

rbazaud:
@pmattern the «and» hasn’t been «dropped intentionally», as indicated, I used the French translation available from the GNU website. It is not an official translation (none of them are indeed). In Boole algebra a ‘or’ may also mean ‘and’ unless it is an ‘exclusive or’ so that didn’t trouble me. If you want I can still add it. I'm now using Qt Linguist, I don't see «vanished» strings in the UI, should I get back to Vim ? I'll update the translations.

@palinek thanks for the information.

PR #267, opened 2015-11-21
Add bindings to Ctrl+PgUp/Ctrl+PgDn.

shlomif:
To switch to next/prev tab as is the general convention.

I hereby disclaim any implicit or explicit ownership of my changes, and
place them under CC0/Public Domain/X11L/Any other licence of your
choice.

I noticed there's quite a lot of duplicate code in creating and binding
shortcuts. Perhaps extracting a small inline method to do that will be a
good idea. What do you say?


palinek:
GTM

PCMan:
Merged. Thank you.
@shlomif any good idea to simplify the code futther?

shlomif:
Thanks for merging the pull-req!

@PCMan : well, I've though of something like:

```  
void MainWindow::add_shortcut(long key, void (*MainWindow::method_ptr)(void))
{
    QShortcut * shortcut = new QShortcut(key, this);
    connect(shortcut, &QShortcut::activated, this, method_ptr);
}
```

or something to this effect.

PR #269, opened 2015-11-22
Fix compiler warnings

rbazaud:
function g_io_scheduler_push_job() deprecated, replaced by its newer
    equivalent g_task_run_in_thread()

### DIFF
--- a/libfm-qt/filepropsdialog.cpp
+++ b/libfm-qt/filepropsdialog.cpp
@@ -295,7 +295,7 @@ void FilePropsDialog::initGeneralPage() {
   connect(fileSizeTimer, &QTimer::timeout, this, &FilePropsDialog::onFileSizeTimerTimeout);
   fileSizeTimer->start(600);
   g_signal_connect(deepCountJob, "finished", G_CALLBACK(onDeepCountJobFinished), this);
-  fm_job_run_async(FM_JOB(deepCountJob));
+  bool bret = fm_job_run_async(FM_JOB(deepCountJob));
 }
 
 /*static */ void FilePropsDialog::onDeepCountJobFinished(FmDeepCountJob* job, FilePropsDialog* pThis) {


libfm-qt/pathedit.cpp
```diff
@@ -111,9 +111,11 @@ void PathEdit::reloadCompleter(bool triggeredByFocusInEvent) {
   // qDebug("load: %s", g_file_get_uri(data->dirName));
   cancellable_ = g_cancellable_new();
   data->cancellable = (GCancellable*)g_object_ref(cancellable_);
-  g_io_scheduler_push_job((GIOSchedulerJobFunc)jobFunc,
-                          data, (GDestroyNotify)JobData::freeMe,
-                          G_PRIORITY_LOW, cancellable_);
+
+  GTask *task = g_task_new(NULL, NULL, NULL, NULL);
```
philippwiesemann: This breaks compatibility with older versions of glib while in the remaining source of libfm-qt the GLIB_CHECK_VERSION macro is used to still support them.

rbazaud: Alright then, what do you propose ? Keep the old code between `#if` (which version ?) `#endif`  to maintain backward compatibility or let the new code in standby as long as the deprecated function is present ?

PCMan:
Would you please add some #if GLIB_CHECK_VERSION(2, 36, 0) to keep the old code working for older glib versions?

PCMan:
After merging this PR, I'm going to try to split the repo and move libfm-qt to its own repo.
This is probably the last one. So, waiting for your updates. :-)

rbazaud:
I've  just seen your message.. It needs a bit more work to make it functional. So if it's going to be used as a «code base» I think it's best that I revert back right now to your code (which works)... Sorry if my mistake induced delays.

PCMan:
No problem. Thank you for the contribution. I'd like to merge this PR. Just want to keep backward compatibility at the same time. Alternatively, we can rewrite this part with QThread. Then we don't need to care about glib versions.
Thanks! I'll leave this PR open and start splitting the repos first.

rbazaud:
You're right it will be even «cleaner» with QThread. Thank you for your understanding.

PCMan:
I already replace this part with QThread. So I'm closing this PR.
Thanks!

PR #270, opened 2015-11-22
Update translations source files (*.ts)

rbazaud:
see PR #266

### DIFF
--- a/libfm-qt/translations/libfm-qt.ts
+++ b/libfm-qt/translations/libfm-qt.ts
@@ -811,12 +811,12 @@ Do you want to delete them instead?</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <location filename="../placesmodel.cpp" line="132"/>
+        <location filename="../placesmodel.cpp" line="143"/>
         <source>Bookmarks</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <location filename="../placesmodel.cpp" line="207"/>
+        <location filename="../placesmodel.cpp" line="222"/>
         <source>Trash</source>
         <translation type="unfinished"></translation>
     </message>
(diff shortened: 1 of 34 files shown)


jleclanche:
@rbazaud This is done once every release. Any reason why you want to do it now?

rbazaud:
Mmh, I didn't know that. I was working on French translation, but it appears many strings I've translated are in fact labeled as «vanished», which is why someone told me it would probably be best to «synchronize» it… Does it count as a valid or legitimate reason ?

pmattern:
There's a bunch of problems due to TS files being handled like they are right now.
Translations that have already been made may get lost or strings may get tagged `type="unfinished"` for apparently no reason, e. g. compare latest version of `pcmanfm-qt_de.ts` in lxde/pcmanfm-qt@8806913 with the one from lxde/pcmanfm-qt@d555e13 where everything had been complete.
The preview of UIs to get translated in QtLinguist sometimes stops working as QtLinguist cannot find the corresponding source files.
Also, diffing the untranslated template (?) TS files like `pcmanfm-qt.ts` with translated counterparts, which still makes sense at times besides tools like QtLinguist are normally used, is frequently pretty much impossible as the order of strings is different on a regular basis.

Of course there are very good reasons to wait for string freezes or releases to update the TS files and I for one don't know how to do better. But nevertheless the current proceeding represents a massive hindrance for people who want to contribute by translating right now (not talking about the lack of something like Pootle but about the problems stated above). And those people have time to do so right now. We cannot expect them to still have time once we're really ready to accept translations.
Also, when using VCS checkouts on production systems it's of course more convenient to have translations at hand.

On a side note people lots of PRs featuring updates of TS files have been accepted frequently already. Problem had been touched in lxde/lxqt#693, too.

PCMan:
Better to update these strings at once before release. But indeed, this create problems for the translators. Shouldn't we just use pootle or something else to manage the translations?

PCMan:
So, I'd say, let's don't touch the *.ts files right now. Let's update them all at once before the next release, which I believe is coming soon since several important bugs are fixed recently.

PR #272, opened 2015-11-26
Move libfm-qt to its own repository.

PCMan:
TODO: Use cmake imported target instead of pkgconfig for libfm-qt.


pmattern:
The new repo has its issue tracker enabled but on https://github.com/lxde/libfm-qt you state the one from PCMan-FM-Qt should be used.
Wouldn't it be good to have this consistent somehow?

Would it on the occasion of moving libfm-qt into a repo on its own eventually be possible to get rid of the GTK dependencies of PCManFM-Qt and libfm-qt?

luis-pereira:
@pmattern This is an WIP. https://github.com/lxde/libfm-qt is not yet ready to be used.
I'm working on it [here] (https://github.com/lxde/libfm-qt/pull/1). When done,  https://github.com/lxde/libfm-qt, will be added to the superproject and pcmamfm-qt updated accordingly.

PCMan:
@luis-pereira I handled the pcmanfm-qt part so it can compile with this new libfm-qt.
It's in "split" branch of pcmanfm-qt already.
To not waste your time, please base on that if possible. :)
Thank you for the help.

luis-pereira:
@PCMan Wasn't aware of it. But they are not overlapping.  I'm working on libfm-qt so that on pcmamfm-qt and lximage-qt we are able to simply right:
`find_package(fm-qt)`
and all needed stuff: dependencies, header, compile_flags, etc etc that are needed, just are added as _by magic_.

luis-pereira:
@PCMan Why do we `find_package(Qt5DBus)` [here](https://github.com/lxde/libfm-qt/blob/master/CMakeLists.txt#L23) and then we don't link to it ?

PCMan:
@luis-pereira That's a mistake. libfm-qt does not need dbus.

luis-pereira:
@PCMan Pls test with https://github.com/lxde/libfm-qt/pull/1.

PCMan:
@luis-pereira  Works like a charm! I'd say let's merge it.
Really appreciate your hard work on this complicated work.
However, there is only one mistake which was introduced by me.
It's pcmanfm-qt that requires xcb, not libfm-qt.
So it's better to move the xcb stuff from libfm-qt to pcmanfm-qt.
Would you please do it? And then let's merge your changes?
Cheers!

luis-pereira:
@PCMan It's done. Merging it

pmattern:
Now that the outsourcing of libfm-qt has taken place and everything seems to work smoothly I wonder whether it wouldn't make sense to revert to numeric tagging / a numeric release.

The text string that's set right now
```
$ git describe --always 
split-libfm-qt-4-g7cff9ef
```
is always considered as an older version than numeric values. So e. g. Arch Linux AUR package `pcmanfm-qt-git` which is - IMO rightly - still using a numeric release version is considered updatable all the time by AUR helpers like yaourt.

luis-pereira:
@pmattern I agree with you. To revert it, we only need to remove [this] (https://github.com/lxde/pcmanfm-qt/tree/split-libfm-qt) tag. @PCMan What's your opinion ?

pmattern:
Took the liberty to set another release / tag `0.10.1`.
Hope that's okay, but that textual `split-libfm-...` one was just too annoying...

PCMan:
Well, the reason I added that tag is just to mark the revision in which libfm-qt is removed.
If you have a better way to do the tagging, please go ahead. :-)

pmattern:
As said above I already too the liberty to set another tag `0.10.1` to get rid of those problems in AUR.
IMO there's no downside as tag `split-libfm-qt` still exists in the release history so everybody can see when the split took place at first glance.

PR #277, opened 2015-12-03
Fix referencing generated headers in translations

philippwiesemann:
The translation files referenced the header files generated for the *.ui files during build instead of directly referencing the *.ui files. This made translating with Linguist more difficult because no form previews were displayed. It was fixed by adapting the CMake file to update the translation files directly from *.ui files instead from generated files.

This maybe also fixes the problem with missing previews mentioned in #266 and #270.

Thanks!


palinek:
GTM

luis-pereira:
Thanks @philippwiesemann

Issue #273, opened 2015-11-28
inconsistencies in context menus


pmattern:
There are several inconsistencies / problems in the various context menus the solution of which would improve the UX.

If left pane is set to Directory Tree view context menus of directories comprise an item `Open with ...` that's missing in right pane.
Opening directories in a distinct application may be much less common than with files but use cases where it makes sense exist, see e. g. #238.
So directories' context menus in right pane should get complemented with the said item, too.

If `Open` gets clicked in a context menu of a directory in right pane all the content of the directory gets displayed in right pane. Plus, if left pane is set to Directory Tree view the corresponding subfolders get displayed there, too (tree view gets expanded automatically). Clicking `Open` in a directory's context menu in left pane in Directory Tree view doesn't seem to trigger any action, though.
So button `Open` in left pane in Directory Tree view should trigger the same actions as in right pane.

Files' context menus in right pane are comprising items `Create New {Folder,Blank File}` and `Paste` which cannot be applied to a file. In fact they get applied to the directory comprising the file which is non-intuitive and confusing. All these items are provided by the context menus of directories and blank space in right pane as well where they do make sense and work as expected.
So I suggest to remove `Create New` and `Paste` from files' context menus.

Directory context menus in left pane in Places view are comprising topics `Open in New {Tab,Window}` only.
I'm not sure about the motivation to do so, maybe it was done with intent. On the other hand other items in Places view are featuring additional topics, too, e. g. those summarized under "Devices" state actions `Unmount` and `Eject`.
So I wonder whether the it wouldn't improve the UX if the context menus of directories in left pane's Places view got complemented with additional actions like those discussed above.

hotvic:
> Clicking Open in a directory's context menu in left pane in Directory Tree view doesn't seem to trigger any action, though.

Works here, just like the right panel.

pmattern:
>> Clicking Open in a directory's context menu in left pane in Directory Tree view doesn't seem to trigger any action, though.

>Works here, just like the right panel.

Weird. Still seeing this here, latest VCS checkouts of all LXQt components on Arch Linux x86_64. Choosing `Open` from the context menu of a directory in Directory Tree view in left pane just doesn't seem to have any effect. Btw. double clicking the same directory behaves like `Open` in the context menus in right pane. The tree of subfolders get expanded which is what I'd consider expected behaviour.

hotvic:
[...]

PR #278, opened 2015-12-03
Fix some inconsistencies in Context Menus

hotvic:
These commits restore the 'Open With...' action for directories in right panel, and hides the 'Paste' action for files.

Both targeted in #273

### DIFF
--- a/pcmanfm/view.cpp
+++ b/pcmanfm/view.cpp
@@ -87,34 +87,42 @@ void View::prepareFileMenu(Fm::FileMenu* menu) {
 
   // add some more menu items for dirs
   bool all_native = true;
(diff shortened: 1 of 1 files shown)


luis-pereira:
GTM. Works as described.
Thanks @hotvic

pmattern:
Any particular reason why you preserved item "Create new" in files' context menu in right pane? As depicted in #273 it makes just as little sense as "Paste", IMO.

hotvic:
@pmattern no, I want to hide it too, but there's some mess in this part of code, I want to clean up and organize so remove/hide it too.

pmattern:
@hotvic 
Thanks for explaining.

PR #280, opened 2015-12-05
hide 'Create New...' menu for files

### LINKED ISSUE #273: inconsistencies in context menus

hotvic:
Hide the last (?) action remaining for files context menu.

closes #273 ?

PS: requires the pull request of libfm-qt to be accept too...

### DIFF
--- a/pcmanfm/view.cpp
+++ b/pcmanfm/view.cpp
@@ -119,6 +119,7 @@ void View::prepareFileMenu(Fm::FileMenu* menu) {
   }
   else {
     menu->pasteAction()->setVisible(false);
+    menu->createAction()->setVisible(false);
   }
 }
 


pmattern:
Works as expected in combination with commit proposed in lxde/libfm-qt#4.

Doesn't necessarily close #273 which is mentioning two more topics not solved yet, though.

PCMan:
Merged into a7533049f24bc02933ce00f61d123a069b51553a. Thanks!

PR #287, opened 2015-12-20
Fix --wallpaper-mode description in translations

philippwiesemann:
Four translations translated the names of wallpaper modes and therefore provided a wrong description for the --help option. The modes must not be translated because the application only accepts untranslated names.

Thanks!


f2404:
Could you maybe add a note that the mode names shouldn't be translated? for the future translators.

philippwiesemann:
There already is a comment. It should be visible in Linguist and the source context.

To prevent similar problems, the modes should not be in the translatable string. They could be added using a placeholder. But this is a feature unrelated to the fix here. :)

paulolieuthier:
It those words are not to be translated, they shouldn't be [inside a `tr()` call](https://github.com/lxde/pcmanfm-qt/blob/master/pcmanfm/application.cpp#L193).

PR #293, opened 2016-01-12
Let the style engine draw the text selection rect on desktop.

tsujan:
The same as https://github.com/lxde/libfm-qt/pull/11 but for desktop (as was promised).


tsujan:
Please consider this PR because I want to make a *clean* PR (to fix desktop bottom gap) that also changes `desktopitemdelegate.cpp`. Thanks in advance!

paulolieuthier:
GTM.

palinek:
As it is the same as in lxde/libfm-qt#11, @tsujan go ahead and merge it.

tsujan:
@palinek , @paulolieuthier  Thanks!

PR #294, opened 2016-01-14
Prevent large gaps at desktop bottom

tsujan:
Here the meaning of "large" depends on several parameters. This commit is based on https://github.com/lxde/libfm-qt/pull/13 and solves https://github.com/lxde/pcmanfm-qt/issues/97#issuecomment-170671223 by first trying to make room for an extra icon at the bottom and then, if not possible, spreading icons vertically. I encountered both possibilities by changing font/icon sizes and it worked in all cases.

### DIFF
--- a/pcmanfm/desktopitemdelegate.cpp
+++ b/pcmanfm/desktopitemdelegate.cpp
@@ -34,7 +34,8 @@ DesktopItemDelegate::DesktopItemDelegate(QListView* view, QObject* parent):
   QStyledItemDelegate(parent ? parent : view),
   view_(view),
   symlinkIcon_(QIcon::fromTheme("emblem-symbolic-link")),
-  shadowColor_(0, 0, 0) {
+  shadowColor_(0, 0, 0),
+  margins_(QSize(3, 3)) {
 }
 
 // FIXME: we need to figure out a way to derive from Fm::FolderItemDelegate to avoid code duplication.
@@ -74,7 +75,7 @@ void DesktopItemDelegate::paint(QPainter* painter, const QStyleOptionViewItem& o
   }
 
   // draw text
-  QSize gridSize = view_->gridSize() - QSize(6, 6);
+  QSize gridSize = view_->gridSize() - 2 * margins_;
   QRectF textRect(opt.rect.x() - (gridSize.width() - opt.rect.width()) / 2,
                   opt.rect.y() + opt.decorationSize.height(),
                   gridSize.width(),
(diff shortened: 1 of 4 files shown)


PCMan:
Looks good, but wouldn't it be better to put this in a private method which can also be invoked when the font or icon sizes are changed?

tsujan:
> ... wouldn't it be better to put this in a private method ...

The margins are "forced" on DesktopItemDelegate -- it can't know about them except from `delegate_ = new DesktopItemDelegate(listView_, cellMargins);` at startup. However, perhaps your idea can be implemented by adding a `setMargin()` method to it. I should test that carefully because the calculations in `desktopitemdelegate.cpp` are delicate. Thanks!

tsujan:
@PCMan 
In `desktopitemdelegate.cpp` you've rightly commented: "we need to figure out a way to derive from Fm::FolderItemDelegate to avoid code duplication."

FolderItemDelegate knows about changed cell margins because `setGridSize()` is used inside `FolderView::updateGridSize()`. But DesktopItemDelegate is a separate class and can't know about our new cell margins unless they're forced on it when it's created. In fact, in https://github.com/lxde/pcmanfm-qt/pull/252, I succeeded in fixing the desktop icons issue only by taking into account the 6-px margin that was introduced in `folderview.cpp`.

On the other hand and for the same reason, we have no way of accessing DesktopItemDelegate to inform it of the new cell margins when font/icon sizes are changed.

Therefore, putting the algorithm inside a private method would have no use to us.

paulolieuthier:
There is a bug. After I login, if I try to move an item, more than one item is moved (usually two items that were close to the one I actually intended to move). After that, I can't move an item to the gap in the bottom. If I restart pcmanfm-qt, I can.

I know my description is awkard. If I get more information of it, I'll certainly post it here.

tsujan:
@paulolieuthier 
Previously, when I set the big icon size to 48px, I got a bottom gap, as in the first screenshot you sent. Now, with this commit, it's filled, as in your second screenshot. I logged out and in but couldn't reproduce the first part of your problem: when I move an item, no other item is moved with it.

As for the second part, what do you mean by "the gap in the bottom"? Because there should be no gap with this commit.

paulolieuthier:
> As for the second part, what do you mean by "the gap in the bottom"? Because there should be no gap with this commit.

That's the problem. After the bug is triggered, I can't move an item to the bottom row.

tsujan:
@paulolieuthier  Just another question until you give me a way of reproducing the first part: do you have two installations pf pcmanfm-qt -- with and without the commit?

paulolieuthier:
Only one, with your commits.

But I think I figured out the reason for the problem. When the session starts, pcmanfm-qt creates the desktop window before the panel asks for its reserved space, so pcmanfm-qt has to resize and that's probably not taken in account for the margins. When I restart pcmanfm-qt, it recalculates the size and use it properly.

tsujan:
@paulolieuthier Nice catch! And the reason why it doesn't happen to me is that I use a left panel. I'll look into it. Thanks!

tsujan:
@paulolieuthier 
That was a "golden" bug; it improved several things. Would you please test the latest source at https://github.com/tsujan/pcmanfm-qt/tree/fix_desktop_bottom_gap before I squash the commits?

tsujan:
@PCMan You were right. Cell margins can be forced *after* DesktopItemDelegate is created. Now the gap is also covered on changing font/icon sizes and more. I was too cautious about "delicate calculations" ;)

paulolieuthier:
Working great now.

PCMan:
Merged. Thank you.

PR #295, opened 2016-01-15
Add some ``beautiful'' workarounds that bypass Qt bug #49947 and 47981.

PCMan:
This should fix the bug lxqt/#688 temporarily.
FYI https://github.com/lxde/lxqt/issues/688

With some dirty hacks digging into XCB and Qt5 internals, now Dnd works again.
I also made the workaround a reusable C++ class which is licensed under LGPL.
So if it works, we can send this gift to KDE developers as well.
Please help test this rigorously. I'm not sure if this will cause new problems.
Cheers!

### DIFF
--- a/pcmanfm/CMakeLists.txt
+++ b/pcmanfm/CMakeLists.txt
@@ -13,6 +13,7 @@ set(pcmanfm_SRCS
     desktopitemdelegate.cpp
     autorundialog.cpp
     settings.cpp
+    xdndworkaround.cpp
 )
 
 qt5_add_dbus_adaptor(pcmanfm_SRCS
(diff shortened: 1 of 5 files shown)


PCMan:
@Esclapion, @paulolieuthier @pmattern @q4a @tsujan I need your testing and code review.
Thank you very much.

PCMan:
@jleclanche if this works, I relly want to make  yet another new release after merging this PR.

jleclanche:
No problem. Email me once you are ready for a release okay?

tsujan:
You workaround works seamlessly here! I tested it with Firefox and Chromium, for both text and image files: all files were opened inside the browsers even when they had special characters in their names.

tsujan:
BTW, something like this might also be used in other Qt5 apps for working around the "scrolling jump" problem, that I still encounter in spite of the claim about it being fixed. It's a source of inspiration to me :)

palinek:
It's working here (debian unstable).

PCMan:
Well, since there are no new test results and no bugs have been found so far, I'm going to merge this.

PCMan:
Merged.

PR #296, opened 2016-01-22
Fixed a crash on right clicking trashed files.

tsujan:
Probably after https://github.com/lxde/pcmanfm-qt/pull/278, PCManFM crashed when a file was right clicked inside Trash because trashed files don't have a paste action (-> libfm-qt -> FileMenu::createMenu()).


paulolieuthier:
GTM. Ship it.

tsujan:
Thanks!

PR #303, opened 2016-02-01
Added options for min. item margins.

tsujan:
Fixes https://github.com/lxde/lxqt/issues/786. Two options are added, one for folder-view (Display section) and the other for desktop (General section). Min. horizontal and vertical margins can be set independently. A max. value of 48px is chosen, which gives a min. spacing of 96px.

Since the default margin is just 3px, these options are mostly useful for those who want a greater distance between items, especially in folder-view.

### DIFF
--- a/pcmanfm/desktop-preferences.ui
+++ b/pcmanfm/desktop-preferences.ui
@@ -201,6 +201,82 @@
         </widget>
        </item>
        <item>
+        <widget class="QGroupBox" name="groupBox_3">
+         <property name="title">
+          <string>Spacing</string>
+         </property>
+         <layout class="QGridLayout" name="gridLayout_3">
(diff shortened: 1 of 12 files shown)


PCMan:
Looks ok. Would you please use new Qt 5 syntax for signal/slot instead?
Thanks!

tsujan:
Done!
I wanted to avoid the complexity of the new syntax for the overloaded signal `QSpinBox::valueChanged()`, but you're right.

PCMan:
The new syntax provides "compile-time" type safety check and it's done without any string comparison. So, please use it whenever possible. Thanks!

PCMan:
Merged into e5644d5200d5b45ce1edaa8a37668581655acc38. Thanks!

PR #307, opened 2016-02-03
Add quit shortcut

npmiller:


palinek:
GTM

paulolieuthier:
GTM.

PCMan:
@palinek @paulolieuthier Thank you for reviewing this.
If you see this kind of little patches which are apparently LGTM, it might be a good idea to just merge them directly? :-)

PCMan:
Merged.

pmattern:
Now that the new shortcut is available, wouldn't it make sense to mention it in menu "File"?
If I'm not mistaken best as `Ctrl+Q` next to "Close Window" much as `Ctrl+W` is already next to "Close Tab" right now   

pcmanfm-qt_missing-ctrl q

tsujan:
> wouldn't it make sense to mention it in menu "File"?

Yes, it does, of course! As a mouse user, I was encouraged by the recent PR's to use the keyboard more and I naturally pressed Ctrl+Q without knowing it wasn't there before.

tsujan:
But then, it should be put in `main-win.ui` instead of `mainwindow.cpp`.

npmiller:
Is that better: https://github.com/lxde/pcmanfm-qt/pull/315 ?

tsujan:
@npmiller  Yes!

PR #308, opened 2016-02-04
Make it possible to hide the menu bar

### LINKED ISSUE #188: visual improvement: hide menu bar
after a week of using the file manager, i usually have set up everything the way i like it. then, i have no need for the menu bar anymore. i would like to be able to hide it for a cleaner look.

thunar and other file managers can hide (and unhide) their menu bars with a keyboard shortcut (usually: CTRL + M). they usually have an option in the "view" menu to hide the menu bar, too.

npmiller:
Fixes #188.

### DIFF
--- a/pcmanfm/mainwindow.cpp
+++ b/pcmanfm/mainwindow.cpp
@@ -164,6 +164,9 @@ MainWindow::MainWindow(FmPath* path):
   group->addAction(ui.actionAscending);
   group->addAction(ui.actionDescending);
 
+  // Show or hide the menu bar
+  ui.menubar->setVisible(settings.showMenuBar());
+
   // create shortcuts
   QShortcut* shortcut;
   shortcut = new QShortcut(QKeySequence::Quit, this);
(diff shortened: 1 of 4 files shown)


pcmanfm/mainwindow.cpp
```diff
@@ -260,6 +267,28 @@ void MainWindow::addTab(FmPath* path) {
   }
 }
 
+void MainWindow::toggleMenuBar() {
+  Settings& settings = static_cast<Application*>(qApp)->settings();
+  bool showMenuBar = !settings.showMenuBar();
+
+  if (!showMenuBar) {
+    QScopedPointer<QMessageBox> warning(
```
palinek: Why are you creating the `QMessageBox` widget on the heap and not on the the stack?

pcmanfm/mainwindow.cpp
```diff
@@ -260,6 +267,28 @@ void MainWindow::addTab(FmPath* path) {
...
+
+    warning->show();
+    warning->exec();
+    if (QMessageBox::Cancel == warning->result()) {
+      return;
+    }
+  }
+
+  ui.menubar->setVisible(showMenuBar);
+  settings.setShowMenuBar(showMenuBar);
```
palinek: Isn't the `setShowMenuBar` superfluous?

npmiller: No, it's to update the settings so that the state of the menu bar will be remembered.

palinek: Sorry, my fault.. I misinterpreted the code...

PCMan:
Looks good. Wouldn't it be simpler if you use QMessageBox::warning() directly?
http://doc.qt.io/qt-5/qmessagebox.html#warning
Thanks!

npmiller:
Good point, updated to use `QMessageBox::warning` instead.

PCMan:
Merged. Thanks!

PR #312, opened 2016-02-14
Update the tab bar visibility when the settings change

npmiller:
Currently when changing the tab bar visibility in the settings, the ui is not updated to reflect the new setting, with this pull request, it is.


tsujan:
GTM.

PCMan:
Merged. Thanks!

PR #313, opened 2016-02-14
Return focus to the folder view when hitting escape

npmiller:
This pull request makes it possible to use escape to "reset" the focus, or more precisely to return it to the folderview of the current tab page. This is quite handy when using pcmanfm with the keyboard.

This pull request relies on: https://github.com/lxde/libfm-qt/pull/18

### DIFF
--- a/pcmanfm/mainwindow.cpp
+++ b/pcmanfm/mainwindow.cpp
@@ -166,6 +166,9 @@ MainWindow::MainWindow(FmPath* path):
 
   // create shortcuts
   QShortcut* shortcut;
+  shortcut = new QShortcut(QKeySequence(Qt::Key_Escape), this);
+  connect(shortcut, &QShortcut::activated, this, &MainWindow::onResetFocus);
+
   shortcut = new QShortcut(QKeySequence(Qt::CTRL + Qt::Key_L), this);
   connect(shortcut, &QShortcut::activated, this, &MainWindow::focusPathEntry);
 
@@ -497,6 +500,12 @@ void MainWindow::on_actionThumbnailView_triggered() {
 
 void MainWindow::onTabBarCloseRequested(int index) {
   closeTab(index);
+}
+
+void MainWindow::onResetFocus() {
+  if(TabPage* page = currentPage()) {
+    currentPage()->folderView()->childView()->setFocus();
+  }
 }
 
 void MainWindow::onTabBarTabMoved(int from, int to) {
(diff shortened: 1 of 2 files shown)


tsujan:
Please also check for the existence of `currentPage()`:
````
if(TabPage* page = currentPage())
...
```

tsujan:
And there's no need to `setFocusProxy(folderView_);`. To return the focus to the folder-view, use `tabPage->folderView()->childView()->setFocus();`.

npmiller:
Ok, comments addressed, so it doesn't depend on https://github.com/lxde/libfm-qt/pull/18 anymore, although I think setting a proper focus proxy on the folderview is probably a good idea anyway.

I have however removed it from the tabpage, because even if it makes sense now, we might end up with several widgets on the same tabpage at some point (split view?), so the focus proxy might not make sense.

tsujan:
GTM. Necessary for keyboard-only use, IMO.

paulolieuthier:
That's better. GTM.

PCMan:
Would it be better to only do this when the focus is in the address bar or filter bar?
With shortcut, the Esc key is intercepted globally for everything in the main window so no child widget can get this key event. For now, there is no problem. However, we'd never know if in the future some widgets in the main window will need the Esc key?

npmiller:
This is a good point, although in terms of user experience I think these shortcuts should have only one behaviour in the main windows anyway.

For example currently tab is hijacked by the address bar for completion, which is fine but it makes it a bit annoying when using tab to switch widgets because you get stuck in the address bar.

Another widget in the main window could use escape for something else, but I think it probably shouldn't, and we should try to avoid it.

PCMan:
@npmiller One case in which Esc might be needed by other widget is, if the user want to close the filter bar with Esc key. This is supported in both Dolphin and Firefox.
IMHO it might be better to only install the QShortcut on the path bar rather than on the whole main window. For example, only do this:
shortcut = new QShortcut(QKeySequence(Qt::Key_Escape), pathEdit);

tsujan:
@PCMan
Please consider this scenario: the user filters a folder by pressing `Ctrl+I` and typing a string in the filterbar. Then he wants to quickly open it *without using the mouse and without closing the filterbar*. That would be easy with `Esc` as it is here.

npmiller:
Yes, I was thinking about @tsujan use case, I also want to add an option so that the filter bar auto-hides when it's empty and loses focus, this way you could also hide the filter bar with escape.

Also even though other widgets can't catch the `Esc` event, the can catch the focus out event and act accordingly.

PCMan:
OK, let's do it this way. If it cause problems later, we can fix it anytime.
Merged. Thanks!

Issue #311, opened 2016-02-12
Add filter bar shortcut


samuelgarcia:
Thank you for that tools and lxQt.

This file browser could be wonderfull with a filter bar like in dolphin accessible with Ctr+I.

palinek:
What version of pcmanfm-qt are you talking about?

Filter bar was implemented in 5fd3eb96aba40130bcad50f092fb6d6717402bc2, see below
![screenshot](https://cloud.githubusercontent.com/assets/10990929/13009894/3262fca2-d19f-11e5-9441-aebf769b9033.png)

samuelgarcia:
oups.
Infinite sorry for disturbing.
So the query is just : could it be possible to add Ctr+I shortcut ?

Thank a lot for LxQt gain. It is wonderfull.

tsujan:
Ctrl+F seems better.

samuelgarcia:
You choose.
Note that on dolphin when Ctrl+I : this show the filter bar AND set the focus in it
very very convinient.
Thanks a lot.
BTW what is your icon theme ?

tsujan:
> ... show the filter bar AND set the focus in it

Yes, it should be focused too. "F" for "Filter", "I" for what?

> BTW what is your icon theme ?

@palinek could answer that ;)

samuelgarcia:
In dolphin, I guess,  F is for Find : this open the search bar wich is something more complicated.

palinek:
> BTW what is your icon theme ?
Numix Circle

npmiller:
I think it might make more sense to keep Ctrl + F for the full blown search file dialog currently accessible with F3.

In addition to a classic shortcut (such as Ctrl + I), it would be nice to also have "/" focus the filter bar.

tsujan:
Usually F11 is used for (un-)full-screening, so Ctrl + F would have no problem.
Ctrl + F should also focus the filterbar and return the focus to the folder-view when hiding it, IMO.

tsujan:
> it would be nice to also have "/" focus the filter bar

That's not possible because one may want to write "/" inside the location-bar. After "Ctrl + F" is implemented, you could use it twice to give the focus to the filterbar. Or tab can be used for that purpose.

npmiller:
I tried adding a "/" shortcut on the mainwindow that focuses the filter bar and it works fine, you can still type an actual "/" in the filter bar and in the url bar. You can try it here: 
[...]

PR #314, opened 2016-02-15
Added shortcut for filterbar.

tsujan:
Also added Ctrl+F, in addition to F3, for the search utility.
This is for https://github.com/lxde/pcmanfm-qt/issues/311.

### DIFF
--- a/pcmanfm/main-win.ui
+++ b/pcmanfm/main-win.ui
@@ -92,6 +92,9 @@
           <property name="clearButtonEnabled">
            <bool>true</bool>
           </property>
+          <property name="toolTip">
+           <string>Focus with Ctrl+I</string>
+          </property>
          </widget>
         </item>
        </layout>
@@ -698,6 +701,9 @@
    </property>
    <property name="text">
     <string>&amp;Filter</string>
+   </property>
+   <property name="shortcut">
+    <string>Ctrl+B</string>
    </property>
   </action>
   <action name="actionCloseLeft">
--- a/pcmanfm/mainwindow.cpp
+++ b/pcmanfm/mainwindow.cpp
@@ -205,6 +205,13 @@ MainWindow::MainWindow(FmPath* path):
   shortcut = new QShortcut(QKeySequence(Qt::SHIFT + Qt::Key_Delete), this);
   connect(shortcut, &QShortcut::activated, this, &MainWindow::on_actionDelete_triggered);
 
+  shortcut = new QShortcut(QKeySequence(Qt::CTRL + Qt::Key_I), this);
+  connect(shortcut, &QShortcut::activated, this, &MainWindow::focusFilterBar);
+
+  // in addition to F3, for convenience
+  shortcut = new QShortcut(QKeySequence(Qt::CTRL + Qt::Key_F), this);
+  connect(shortcut, &QShortcut::activated, ui.actionFindFiles, &QAction::trigger);
+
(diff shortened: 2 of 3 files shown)


pmattern:
tsujan/pcmanfm-qt@43fb282 is not working as expected here:

If the filter bar is already visible (opened) hitting `Ctrl+I` hides (closes) it no matter whether or not it has the focus.

Also, I'm not sure whether it makes sense to state `Ctrl+I` as shortcut at menu View - Filter.
As discussed in https://github.com/lxde/pcmanfm-qt/issues/311#issuecomment-184201291 there are basically two actions involved, making the filter bar (in-) visible (close / open it) and bring the focus into it. Shortcut `Ctrl+I` is primarily meant to set the focus. In addition it's opening the filter bar should the latter be closed when the shortcut gets hit which makes perfect sense. But the shortcut is not really meant to hide the filter bar, is it? Therefore it's wrong to state it next to topic "Filter" in menu "View".

(Shortcut `Ctrl+F` is opening the search toll without preventing `F3` from doing the same which is both expected behaviour.)

tsujan:
@pmattern Thanks for testing!

Oh, you want the exact behavior of Dolphin while I thought in this way: the user may want to hide/show the filterbar so that when it's shown, it gets focus and when it's hidden, the focus returns to the folderview.

As I said earlier, I see no reason to imitate Dolphin or any other file-manager to such an extent but if others want it too, why not? It's just a matter of a few additions and changes. Let's wait for another vote ;)

tsujan:
@pmattern
OK, I supposed that @palinek was on your side and made some changes. Is it now what you had in mind?

palinek:
To be correct, I wasn't aware that the filter bar is hidable at all (I use pcmanfm-qt very sporadically).

Just tested and it's OK for me. (just missing the Esc shortcut, but that's the different PR)

GTM

pmattern:
Imitating or not imitating some other file manger isn't a concern for me (in fact I even don't like Dolphin at all). I'm just trying to think about what makes sense and what doesn't.
Regarding `Ctrl+I` I was just thinking that it should *always* bring the focus to the filter bar, not just if the bar happens to be closed, which is implemented now.
You even added a shortcut `Ctrl+B` to open/close the filter bar which is something I'd consider optional as said before but which sure doesn't hurt.

All in all definitely GTM from my side, too.

tsujan:
@pmattern 
Thanks! Surely I got your point *(keyboard-only use)* when I added `Ctrl+I` and especially afterward, when I used it. Good things could come up when the two sides of a dialog have different or opposite opinions -- like what happened here.

tsujan:
When the filterbar is closed (made hidden), the filter strings should be cleared for ALL tabs. That's about an old bug, which is now fixed by this commit too.

PCMan:
LGTM, but there is one problem. Ctrl+I is reserved for invert selection in old gtk+ version of PCManFM and I forgot to add this shortcut to the Qt version.
If we're going to keep the old behavour of the gtk+ version, it's better to use something other than Ctrl+I for the filter bar. However, I'm OK to change it if more users want Ctrl+I for the filter bar instead.

pmattern:
Given that Dolphin is using `Ctrl+I` for the filter bar as well I'd personally tend to say let's use it, too, even if this is inconsistent with the GTK flavour's behaviour.

PCMan:
Merged. Thanks!

Issue #144, opened 2015-02-28
removable media: tree view not handy


pmattern:
As of now the only way to display removable media in tree view is to open their mount point within the FHS view in left pane. This isn't exactly convenient as most of the time mount points are somewhere in /run/user/\<UID\>/gvfs/\<medium\>.

I think it would be good if the root directory of removable media was represented by a temporary object in tree view's top level in left pane as shown with an ODD and an SFTP share in Thunar here

![thunar vs removable media](https://cloud.githubusercontent.com/assets/9860801/6428500/f6910a64-bfa4-11e4-984d-ac36701cabfb.png)
(Thunar by default adds some additional GVFS objects and that default desktop folder as well. At least the former doesn't hurt either, imo.)

tsujan:
The problem is that if you go to the tree view, the current folder won't be selected automatically. Otherwise, cluttering the GUi just because the user doesn't want to turn the mouse wheel on the upper combo box (which is missing in the ancient screenshot attached to this report) isn't a good idea.

I'm closing useless reports but hesitating to close this one because the above-mentioned problem isn't reported.

PR #316, opened 2016-02-27
Put tabbar right above the folder-view.

tsujan:
Just a simple layout change. The previous position of tabs, above the left pane, seemed a little counterintuitive.

### DIFF
--- a/pcmanfm/main-win.ui
+++ b/pcmanfm/main-win.ui
@@ -32,19 +32,6 @@
      <number>0</number>
     </property>
     <item>
-     <widget class="PCManFM::TabBar" name="tabBar" native="true">
-      <property name="sizePolicy">
-       <sizepolicy hsizetype="Preferred" vsizetype="Fixed">
-        <horstretch>0</horstretch>
-        <verstretch>0</verstretch>
-       </sizepolicy>
-      </property>
-      <property name="acceptDrops">
-       <bool>true</bool>
-      </property>
-     </widget>
-    </item>
-    <item>
      <widget class="QSplitter" name="splitter">
       <property name="orientation">
        <enum>Qt::Horizontal</enum>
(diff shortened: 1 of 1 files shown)


PCMan:
No, this is done by design.
1. This can contain more tabs in the same window width.
2. Firefox has similar layout.
3. We've been using this in old gtk+ version of pcmanfm and no users complain of that.

PCMan:
However, if this is optional and is turned off by default, I'm happy to include it.

pmattern:
Besides I didn't open an issue yet something I'd love to see is being able to make the choice between Directory Tree and Places view not per window as right now but per tab. This is something I figure would be very useful, in particular as long as #144 isn't addressed. But if selecting the view mode per tab should ever get implemented the current placement of tabs would still make more sense.
Regarding the current behaviour to have the same kind of view in all tabs of a window I agree that the layout proposed by this PR is definitely better.
So all in all I agree with @PCMan that this new feature should somehow be made optional.

tsujan:
@PCMan
This change of layout doesn't seem so important to me. I used Qt styles with centered tabs since I came to LXQT. Once I applied a style with left-aligned tabs, I found the current layout somehow odd and also aesthetically unappealing. However:

> This can contain more tabs in the same window width.

Most screens are wide enough nowadays. Moreover, tab scroll buttons and "tab tear" automatically appear when needed (I've never seen them with PCManFM although I usually open multiple tabs).

> Firefox has similar layout.

It could be also said that Dolphin has a layout similar to this PR.

> We've been using this in old gtk+ version of pcmanfm

That's just a matter of habit. Perhaps, my point is so too ;)

PCMan:
@tsujan Would you please make this feature optional?

tsujan:
@PCMan Sure!

tsujan:
@PCMan 
Done at https://github.com/lxde/pcmanfm-qt/pull/323 !
Please close this PR if you prefer https://github.com/lxde/pcmanfm-qt/pull/323 over it.

PCMan:
Yes, let's fix this there.

PR #318, opened 2016-02-27
Add menubar actions to the main window

### LINKED ISSUE #317: bug: F2 (rename) option not available when menu bar is hidden
i have just built the latest pcmanfm-qt-git and found a bug:
after i have hidden the menu bar (with CTRL+M), i cannot rename files anymore by pressing F2.  but i can rename it, when i right click --> rename.
CTRL+B does not work anymore, too.
F3 does not work, too.

it seems that all keyboard shortcuts stop working with a hidden menu bar.
this behavior is clearly a bug.

btw, thanks for this great feature! its just a little buggy at the moment...

npmiller:
Fixes #317

### DIFF
--- a/pcmanfm/mainwindow.cpp
+++ b/pcmanfm/mainwindow.cpp
@@ -164,6 +164,10 @@ MainWindow::MainWindow(FmPath* path):
   group->addAction(ui.actionAscending);
   group->addAction(ui.actionDescending);
 
+  // Add menubar actions to the main window this is necessary so that actions
+  // shortcuts are still working when the menubar is hidden.
+  addActions(ui.menubar->actions());
+
   // Show or hide the menu bar
   ui.menubar->setVisible(settings.showMenuBar());
 


jleclanche:
gtm

And check your email Nicolas :)

tsujan:
This works as intended and should be merged ASAP. However, please put yourself in the user's shoes: how could he/she know about `Ctrl+M`? I think it's better to put the shortcut in the View menu, as an ordinary menu-item, before or after "Filter". Here seems a good place to do that too. @npmiller What do you think?

npmiller:
@tsujan yes, this feature definitely needs a ui element, but I think it would also be nice to have a way to enable the menu bar again from the ui without having to remember the shortcut, so just a menu entry might not be enough. One way that some applications do that is by showing a button to open the settings window when the menu bar is hidden.

For now I think it's probably fine to just have the shortcut as a lot of other Qt applications use this shortcut to hide the menu bar, so people looking for this feature should be able to find it.

tsujan:
> so just a menu entry might not be enough...

I think it's enough, especially with the messagebox you've added. And even if not enough, it's better than nothing, especially when its implementation is so easy.

npmiller:
@tsujan sure thing, see #320 has to be merged after this one though because it makes hiding the menu bar an action, so without this one it's impossible to get the menu bar to show again.

tsujan:
@npmiller
What I proposed was just this: Adding an uncheckable menuitem (actionMenubar) to the View menu with `CTRL+M` as shortcut, removing the same shortcut from `mainwindow.cpp` and replacing its connection with a QAction connection in the following way *and in this same PR*:
````
connect(ui.actionMenubar, &QAction::triggered, this, &MainWindow::toggleMenuBar);
````
Now that all actions are added to the main window, it'll work when the menubar is hidden too. Isn't this simpler?

npmiller:
@tsujan that's exactly what #320 does, just in an another PR, and in a checkable action but that doesn't really matter (and is what's seems to be done in KDE applications).

tsujan:
> that's exactly what #320 does

Oh, my bad! I mistook it for something else because of the checkable action. GTM.

PCMan:
Merged into ad346931e8f2bf908f5728db7d91a8c5bd7040dd.
Thanks!

PR #322, opened 2016-03-06
A toolbar button appearing on hiding menubar

tsujan:
It has a dropdown menu containing all menubar items, is added to the end of toolbar, after a separator, when menubar is hidden and is removed (hidden) with its separator when menubar is shown.

### DIFF
--- a/pcmanfm/main-win.ui
+++ b/pcmanfm/main-win.ui
@@ -248,6 +248,8 @@
    <addaction name="actionGoUp"/>
    <addaction name="actionReload"/>
    <addaction name="actionGo"/>
+   <addaction name="separator"/>
+   <addaction name="actionMenu"/>
   </widget>
   <action name="actionGoUp">
    <property name="icon">
@@ -745,6 +747,17 @@
    </property>
    <property name="shortcut">
     <string>Ctrl+M</string>
+   </property>
+  </action>
+  <action name="actionMenu">
+   <property name="icon">
+    <iconset theme="application-menu"/>
+   </property>
+   <property name="text">
+    <string>Menu</string>
+   </property>
+   <property name="toolTip">
+    <string>Menu</string>
    </property>
   </action>
  </widget>
--- a/pcmanfm/mainwindow.cpp
+++ b/pcmanfm/mainwindow.cpp
@@ -169,6 +169,25 @@ MainWindow::MainWindow(FmPath* path):
   addActions(ui.menubar->actions());
 
   // Show or hide the menu bar
+  QMenu *menu = new QMenu();
+  menu->addMenu(ui.menu_File);
+  menu->addMenu(ui.menu_Editw);
+  menu->addMenu(ui.menu_View);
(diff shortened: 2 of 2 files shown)


pmattern:
The rectangle around the icon that comes up upon hovering doesn't disappear when the menu was opened and closed again. It does not matter whether the menu gets closed due to picking one of its actions or without any action by clicking elsewhere on the desktop.

The menu doesn't open reliably when the button introduced by this PR gets clicked.
The button definitely reacts. Upon hovering over the button a border marking the rectangle around the icon appears if it hasn't been there before (see above). Each click on the icon is confirmed by the background getting dark grey for a moment. The menu doesn't open upon the first click but only upon one of the following ones, though. Didn't manage to figure out when exactly it opens, neither some distinct position of the mouse pointer nor a particular number of clicks made the menu open reproducibly here.

Other than that the new button works as expected and is certainly a good idea.

e5dd3ce plus all other LXQt components from yesterday's VCS checkout on Arch Linux.

tsujan:
@pmattern 
The first issue is a Qt5 bug that's reported in various forms but not completely fixed. You could see it everywhere -- in Dolphin, for example.

The second "issue" is not an issue at all. It's a toolbutton with *delayed popup* (`QToolButton::DelayedPopup`): you should hold pressing until the menu shows up. Strange that you hadn't seen such buttons before!

pmattern:
> The first issue is a Qt5 bug that's reported in various forms 

I see. Any links so as to be able to keep an eye on this?

> The second "issue" is not an issue at all. It's a toolbutton with delayed popup 

Confirmed. Not sure whether this makes really sense, though.
I mean the button will appear a the outer right of the so-called "Main Toolbar" and none of the other buttons in that bar features a delay like that.

tsujan:
> Any links so as to be able to keep an eye on this

Please see my comment at https://github.com/lxde/lxqt/issues/971 . There are other reports too but that one was the nearest report I found. You could also tell them that the bug isn't fixed in the Qt5 version you have if this is about the same bug.

> none of the other buttons in that bar features a delay like that

Because none of them has a popup menu in the first place. IMO, the delayed popup shows the user that the menu button is exceptional -- which really is.

palinek:
> >none of the other buttons in that bar features a delay like that

> Because none of them has a popup menu in the first place. IMO, the delayed popup shows the user that the menu button is exceptional -- which really is.

But the button doesn't provide any action on the click... therefore might be confusing (for me it was). IMO the click should trigger menu show-up instantly here.

pmattern:
Meanwhile I saw that the counterparts of the button discussed here of e. g. Chromium and Firefox are opening the corresponding menus right upon single click, too.
So I still figure the delay will be rather confusing for users.

tsujan:
OK. Now that nobody likes delayed popups (a gtk+ habit?), I'll change it.

pmattern:
GTM.

palinek:
GTM

PCMan:
Merged. Thanks for the nice work!

PR #323, opened 2016-03-06
Optional tabbar stretching

tsujan:
By default, the tabbar is stretched over the left pane too. This commit makes that optional with a check button in the View menu. It's checked by default but if unchecked, the tabbar will be positioned right above the folder-view.

Theoretically, the tabbar could be positioned at the bottom, right or left too but, IMO, such positions aren't practical in the case of PCManFM.

### DIFF
--- a/pcmanfm/mainwindow.cpp
+++ b/pcmanfm/mainwindow.cpp
@@ -85,6 +85,10 @@ MainWindow::MainWindow(FmPath* path):
   ui.tabBar->setElideMode(Qt::ElideRight);
   ui.tabBar->setExpanding(false);
   ui.tabBar->setMovable(true); // reorder the tabs by dragging
+  if(!settings.fullWidthTabBar()) {
+    ui.verticalLayout->removeWidget(ui.tabBar);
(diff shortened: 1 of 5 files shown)


pmattern:
GTM.

PCMan:
Thanks! The code looks pretty good. However, the meaning of "Stretched tabbar" seem to be obscure to me. In addition, this does not seem to be an option that you'll need to change often. So I'd say it might be a better idea to put this option as a checkbox in the UI page of preference dialog rather than adding a menu item for it. In this way, we can have a longer label to fully describe what the option actually do and avoid cluttering of the View menu.
What do you think?

excalibur1234:
some option in preferences called "full-width tabbar" would be nice...

tsujan:
@PCMan 
Good idea! I'll move it to Preferences Dialog.

@excalibur1234 
Yes, "full-width" is much better than "stretched". Thanks!

tsujan:
Done! Moved the checkbox to the Preferences dialog, renamed it as "Fullwidth tab bar", and added an informative tooltip to it.

PCMan:
Merged. Thanks for the hard work.

PR #330, opened 2016-03-19
Switch automatically to newly opened tabs

npmiller:
This feels a bit more natural as it matches the usual behaviour of tabbed applications, although the difference is small since the new tabs are opened in the same directory.

This only applies to when the user requests a new "empty" tab, so for example opening something in a new tab with a middle click still opens the tab in the background.

### DIFF
--- a/pcmanfm/mainwindow.cpp
+++ b/pcmanfm/mainwindow.cpp
@@ -279,7 +279,7 @@ void MainWindow::chdir(FmPath* path) {
 }
 
 // add a new tab
-void MainWindow::addTab(FmPath* path) {
+int MainWindow::addTab(FmPath* path) {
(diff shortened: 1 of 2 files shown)


tsujan:
It's just a matter of habit. If the user wants to open multiple new tabs in order to open different folders in each of them later, then the current behavior will feel more natural.

jleclanche:
Approving this on the UX side. gtm

npmiller:
@tsujan sure, either way it won't fit perfectly every use case, but I think this way is a lot more common, that's what web browsers do for example.

tsujan:
No argument against it if users like it. The code is GTM.

PCMan:
Merged.
Thanks!

PR #332, opened 2016-03-20
Fixes libfm-qt dependency contradiction on README.md

### LINKED ISSUE #331: libfm-qt dependency contradiction
In CMakeLists.txt:
`find_package(fm-qt REQUIRED)`

seems contradictive to the statement in README.md:
`Includes libfm-qt, the Qt port of the libfm - a library providing components to build desktop file managers.`

Seems one of the two is not up-to date, because without satisfying the latter
dependency, build fails as expected...

What's going on here?

luis-pereira:
libfm-qt and pcmanfm-qt were split. This is just and artifact.
Closes #331.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,6 @@
 
 The Qt port of the LXDE file manager PCManFM.
 
-Includes libfm-qt, the Qt port of the libfm - a library providing components to
-build desktop file managers.
-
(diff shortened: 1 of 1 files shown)


jleclanche:
README changes you can just land without PRing

Issue #334, opened 2016-04-04
Allow to choose side icons size with context menu too


goetzc:
Right now it is possible to choose the side panel icon size from within a submenu in preferences, but it is not so intuitive as directly right-clicking an icon in side panel and choosing one of the available icon sizes.

An example:
![panel icon size in context menu](https://cloud.githubusercontent.com/assets/2220440/14236026/bc111748-f9d3-11e5-8900-7683fc3cfbfc.png)

tsujan:
IMO, it isn't so intuitive but only what Dolphin does and so, a matter of taste.

pmattern:
I agree with @tsujan that this is a matter of taste and that configuring the icon size from FHS objects' context menus doesn't make really sense.
IMO it's a feature that's not needed to frequently and as such rather bloating the context menus.

But nevertheless I think the handling of icons does need improvement.
The purpose of the groups defined in Preferences isn't remotely obvious. E. g. the "big icons" in terms of Preferences - Display are covering the desktop as well but this isn't stated anywhere in the UI. Maybe the UI components these groups of icons apply to could be stated by a tooltip?
Also, I think setting the size of desktop icons should be possible from within `pcmanfm --desktop-pref=foo` / dialogue "Desktop Preferences" which is accessible from the desktop's context menu.

tsujan:
> But nevertheless I think the handling of icons does need improvement.

I agree. Since desktop has a separate font size, it's logical to give it a separate icon size too. I'll see whether that can be done without bloating the code.

pmattern:
When the same shortcut gets assigned to several actions `lxqt-config-globalkeyshortcuts` prints one of the shortcuts in bold but doesn't really throw a warning.
This should eventually be changed, too.

tsujan:
Apparently, we agree that menus shouldn't be cluttered, on the one hand, and @pmattern's suggestion is implemented by https://github.com/lxde/pcmanfm-qt/pull/335, on the hand. So, this issue can be closed.

pmattern:
Now that the size of desktop icons can be set from `pcmanfm-qt --desktop-pref=general` I cannot see an effect of "Size of big icons" in Preferences - Display any longer.
If so, shouldn't this setting be dropped? Otherwise, can anyone give me a hint which icons this setting is still referring to?

Wouldn't it eventually be good if the desktop icons' size could be tweaked from a new topic in Preferences - Display in addition to `pcmanfm-qt --desktop-pref=general`?

[...]

PR #335, opened 2016-04-05
Add setting for Desktop con size

tsujan:
Since Desktop has its own font and spacing settings, it seems logical to give it a separate setting for icon size too (see https://github.com/lxde/pcmanfm-qt/issues/334#issuecomment-205264654).

### DIFF
--- a/pcmanfm/desktop-preferences.ui
+++ b/pcmanfm/desktop-preferences.ui
@@ -114,6 +114,25 @@
         </widget>
        </item>
        <item>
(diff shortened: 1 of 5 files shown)


PCMan:
Merged. Thank you.
BTW, IMHO It might be better to group icon size and margins/spacing into the same group box since they are all about icons.

tsujan:
> BTW, IMHO It might be better to group icon size and margins/spacing into the same group box since they are all about icons.

Yes, at first, I thought so too, but then I saw that margins are for icons + texts = items.

PR #341, opened 2016-04-17
Reading and save desktop item sorting

tsujan:
The desktop sorting wasn't read or saved before!

@pmattern
Please test it thoroughly so that we won't bother PCMan for merging it!

The default sorting should be as before (time+ascending+folders-first).

### DIFF
--- a/pcmanfm/desktopwindow.cpp
+++ b/pcmanfm/desktopwindow.cpp
@@ -96,12 +96,14 @@ DesktopWindow::DesktopWindow(int screenNum):
     proxyModel_ = new Fm::ProxyFolderModel();
     proxyModel_->setSourceModel(model_);
     proxyModel_->setShowThumbnails(settings.showThumbnails());
-    proxyModel_->sort(Fm::FolderModel::ColumnFileMTime);
+    proxyModel_->sort(settings.desktopSortColumn(), settings.desktopSortOrder());
+    proxyModel_->setFolderFirst(settings.desktopSortFolderFirst());
     setModel(proxyModel_);
 
     connect(proxyModel_, &Fm::ProxyFolderModel::rowsInserted, this, &DesktopWindow::onRowsInserted);
     connect(proxyModel_, &Fm::ProxyFolderModel::rowsAboutToBeRemoved, this, &DesktopWindow::onRowsAboutToBeRemoved);
     connect(proxyModel_, &Fm::ProxyFolderModel::layoutChanged, this, &DesktopWindow::onLayoutChanged);
+    connect(proxyModel_, &Fm::ProxyFolderModel::sortFilterChanged, this, &DesktopWindow::onModelSortFilterChanged);
     connect(listView_, &QListView::indexesMoved, this, &DesktopWindow::onIndexesMoved);
   }
 
@@ -424,6 +426,13 @@ void DesktopWindow::onRowsAboutToBeRemoved(const QModelIndex& parent, int start,
 
 void DesktopWindow::onLayoutChanged() {
   queueRelayout();
+}
+
+void DesktopWindow::onModelSortFilterChanged() {
+  Settings& settings = static_cast<Application*>(qApp)->settings();
+  settings.setDesktopSortColumn(static_cast<Fm::FolderModel::ColumnId>(proxyModel_->sortColumn()));
+  settings.setDesktopSortOrder(proxyModel_->sortOrder());
+  settings.setSesktopSortFolderFirst(proxyModel_->folderFirst());
 }
 
 void DesktopWindow::onIndexesMoved(const QModelIndexList& indexes) {
--- a/pcmanfm/desktopwindow.h
+++ b/pcmanfm/desktopwindow.h
@@ -98,6 +98,7 @@ protected Q_SLOTS:
   void onRowsAboutToBeRemoved(const QModelIndex& parent, int start, int end);
   void onRowsInserted(const QModelIndex& parent, int start, int end);
   void onLayoutChanged();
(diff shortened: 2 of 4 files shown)


paulolieuthier:
I can't test it right now, but I guess sorting wasn't enabled before because many people use manual sorting in their desktops. I don't want my desktop icons automatically sorted. It should be optional.

tsujan:
> I don't want my desktop icons automatically sorted. It should be optional.

This PR is exactly for that.

PCMan:
@tsujan FYI: https://github.com/lxde/libfm/blob/master/src/base/fm-folder-config.h
Actually, we have a module in libfm dealing with per-folder settings. I haven't used it in pcmanfm-qt.
With this, view settings of the specified folder can be customized. It might be a good idea to reuse this rather than adding a special case for ~/Desktop. What do you think?
Well, it's also the same time to add this feature in pcmanfm-qt. Some UI modifications, such as a check item for "Preserve custom settings for this folder" or something alike, is needed. We can follow the UI design for this in the gtk+ version of pcmanfm.

PCMan:
@tsujan:
More info:
http://git.lxde.org/gitweb/?p=lxde/pcmanfm.git;a=blob;f=src/main-win-ui.c;h=35f0bf9ff578413852394f39eaad68c405dfa877;hb=HEAD#l278

tsujan:
@PCMan
Per-folder settings could be a nice new feature. Thanks for the links! I'll work on it.

However, this is simply about reading and saving the sorting order of the desktop *as desktop* and **not** as a folder shown in PCManFM's window. The user may want the mod. time sorting on the desktop but the name sorting in the desktop folder shown inside the main window. Desktop (as desktop) has its own settings but this special setting wasn't read or saved; hence the commit.

PCMan:
@tsujan : Thank you for the effort. I got the idea and agreed with you.

pmattern:
GTM yet some things remain to be discussed as two different topics are mixed by now:

**considering and modifying settings.conf of PCManFM-Qt**
Both didn't take place so far.
On a new user profile directives `SortColumn=name` and `SortOrder=ascending` in section `[Desktop]` were adopted from template `settings.conf` in `$XDG_CONFIG_DIRS` but not considered as the sorting of the desktop entry files on the desktop was always based on the mtime. Similarly changes made in submenu `Sorting` of the desktop context menu were valid for the current session only. They did not get written to `session.conf` in `$XDG_CONFIG_HOME` and hence not considered the next session.
Both is fixed in lxde/pcmanfm-qt@3b051ad of branch `desktop_sort_settings`.
@tsujan
> The default sorting should be as before (time+ascending+folders-first).

No, it isn't as the said keys of the template `settings.conf` have not be considered before but are in branch `desktop_sort_settings`. But this is absolutely the right thing to do.

**automatic sorting in general**
To some degree desktop entry files on the desktop are *always* subject to automatic sorting, both in current `master` and `desktop_sort_settings`.
E. g. set the desktop sorting to ascending by name and start a session with an empty desktop. When "QupZilla" gets dragged and dropped from the main menu to some arbitrary location on the desktop it will end in the upper left corner of the desktop (most of the time, when exactly?!). But when Firefox gets dragged & dropped the same way it will end at the upper left corner as well replacing QupZilla due to the sorting by ascending name.
This behaviour is probably the underlying reason of lxde/lxqt#1030.
I agree with @paulolieuthier that his is questionable behaviour and that it should be possible to have no automatic sorting at all.
This topic is completely independent of the one discussed in this PR and should hence be discussed in an issue on its own, maybe in lxde/lxqt#1030.

tsujan:
@pmattern 
When I said, "the default sorting should be..." I meant it technically, but ignore it because it wasn’t important.

So, if I've correctly understood, the bug you implicitly mentioned at https://github.com/lxde/lxqt/issues/874#issuecomment-211074539 is fixed now? I mean saving and reading of the sorting correctly, because that was the only purpose of this PR.

As I said elsewhere, please open a separate issue/feature request for the rest. Thanks in advance!

tsujan:
@pmattern 
BTW, I'd be grateful if you also test https://github.com/lxde/pcmanfm-qt/pull/343 whenever you have time. I'm using it but the more tests, the better.

pmattern:
@tsujan
Quote
> The sort order gets saved in section `[FolderView]` of PCManFM-Qt' `settings.ini` only. Would it eventually make sense to sync that setting with the one in section `[Desktop]`?

in https://github.com/lxde/lxqt/issues/874#issuecomment-211074539 was due to a misconception and can simply be forgotten. I hadn't realized that submenu "Sorting" on the desktop is meant to correlate with section `[Desktop]` in `settings.conf` of PCManFM-Qt.
The starting point regarding the problem addressed in this PR was lxde/lxqt-common#48 were it turned out as secondary finding that settings regarding sorting in the said section `[Desktop]` get neither read nor modified upon tweaking the settings in submenu "Sorting".

Pointing out the difference between this issue and the topic general automatic sorting was basically the only purpose of my lengthy comment above.
The first topic, sorting settings in section `[Desktop]` not getting read and modified is solved by this PR which should get merged, IMO. Regarding the second one, should automatic sorting on the desktop *always* take place y/n, a remarkable discussion has involved in lxde/lxqt#1030 by now anyway.

#342 / #343 are on my ToDo list but I've got the feeling it makes sense to finish this one before looking into those two.

tsujan:
@pmattern Thank you for this and for putting https://github.com/lxde/pcmanfm-qt/pull/343 in your ToDo list!
This is simple enough to be merged. @PCMan agreed too (https://github.com/lxde/pcmanfm-qt/pull/341#issuecomment-211379167)

PR #343, opened 2016-04-19
Per-folder saving of sorting and view mode

tsujan:
Separate per-folder savings of sorting and view mode is added to the View sub-menus as check boxes. All settings are saved to and retrieved from `~/.config/pcmanfm-qt/PROFILE/perfolder-settings.conf`. If a box is checked, the current folder will have its own sorting or view mode, which could be changed in the corresponding sub-menu later. If unchecked, the default will be used again.

With this commit, it's possible, for example, to set the wallpaper folder in the thumbnail mode while using the default sorting for it, *and* sort the download folder according to mod. time while it's shown in the default view mode.

### DIFF
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -347,6 +347,24 @@ void Application::onUserDirsChanged()
 
 void Application::onAboutToQuit() {
   qDebug("aboutToQuit");
+  // remove non-existent folders from the list of custom folders
+  QString configFile = QString("%1/perfolder-settings.conf").arg(settings_.profileDir(settings_.profileName()));
+  QSettings folderSettings(configFile, QSettings::IniFormat);
+  QHash<QString, QVariant> customFolders = folderSettings.value("customFolders").toHash();
+  if(!customFolders.isEmpty()) {
+    //qDebug() << "Custom folders: " << customFolders.keys();
+    QHash<QString, QVariant>::iterator i = customFolders.begin();
+    while (i != customFolders.end()) {
+      QDir dir(i.key());
+      if(!dir.exists()) {
+        i = customFolders.erase(i);
+      }
+      else ++i;
+    }
+    folderSettings.setValue("customFolders", customFolders);
+  }
+  if(customFolders.isEmpty())
+    folderSettings.clear();
   settings_.save();
 }
 
--- a/pcmanfm/main-win.ui
+++ b/pcmanfm/main-win.ui
@@ -145,6 +145,9 @@
      <string>&amp;View</string>
     </property>
     <widget class="QMenu" name="menuSorting">
+     <property name="tearOffEnabled">
+      <bool>true</bool>
+     </property>
      <property name="title">
       <string>&amp;Sorting</string>
      </property>
@@ -159,16 +162,29 @@
      <addaction name="separator"/>
      <addaction name="actionFolderFirst"/>
      <addaction name="actionCaseSensitive"/>
+     <addaction name="separator"/>
+     <addaction name="actionPreserveSorting"/>
+    </widget>
+    <widget class="QMenu" name="menu_View_2">
+     <property name="tearOffEnabled">
+      <bool>true</bool>
+     </property>
+     <property name="title">
+      <string>&amp;View</string>
+     </property>
+     <addaction name="actionIconView"/>
+     <addaction name="actionThumbnailView"/>
+     <addaction name="actionCompactView"/>
+     <addaction name="actionDetailedList"/>
+     <addaction name="separator"/>
+     <addaction name="actionPreserveView"/>
     </widget>
     <addaction name="actionReload"/>
     <addaction name="actionShowHidden"/>
     <addaction name="separator"/>
-    <addaction name="actionIconView"/>
-    <addaction name="actionThumbnailView"/>
-    <addaction name="actionCompactView"/>
-    <addaction name="actionDetailedList"/>
-    <addaction name="separator"/>
+    <addaction name="menu_View_2"/>
     <addaction name="menuSorting"/>
+    <addaction name="separator"/>
     <addaction name="actionFilter"/>
     <addaction name="actionMenu_bar"/>
    </widget>
@@ -531,6 +547,14 @@
     <string>&amp;Folder First</string>
    </property>
   </action>
+  <action name="actionPreserveSorting">
+   <property name="checkable">
+    <bool>true</bool>
+   </property>
+   <property name="text">
+    <string>&amp;Preserve sorting for this folder</string>
+   </property>
+  </action>
   <action name="actionNewTab">
    <property name="icon">
     <iconset theme="window-new">
@@ -758,6 +782,14 @@
    </property>
    <property name="toolTip">
     <string>Menu</string>
+   </property>
+  </action>
+  <action name="actionPreserveView">
+   <property name="checkable">
+    <bool>true</bool>
+   </property>
+   <property name="text">
+    <string>&amp;Preserve view for this folder</string>
    </property>
   </action>
  </widget>
--- a/pcmanfm/mainwindow.cpp
+++ b/pcmanfm/mainwindow.cpp
@@ -30,6 +30,7 @@
 #include <QShortcut>
 #include <QKeySequence>
 #include <QDir>
+#include <QSettings>
 #include <QDebug>
 
 #include "tabpage.h"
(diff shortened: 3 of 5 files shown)


PCMan:
@tsujan Looks pretty good at first glance. Don't have the time for a more detailed review, though.
However, is there any reason not to use the existing libfm APIs?
IMHO, it might still be better to just reuse existing mechanism to reduce code bloat, improve ease of maintenance, and encapsulation. What do you think?
Thanks!

tsujan:
@PCManFM
As soon as I have free time, I'll try your previous suggestion to see whether the same goal can be reached by using `fm-folder-config.h` functions but with less lines of code.

My "ideal" is independence from GLib. It may not be attainable (I used `fm_path_hash()` in this commit, for example) but, at least, less dependency is. That being said, I always try to put your preferences above mine in LXQT.

tsujan:
@PCMan
I may be totally lost here because I have no expertise in libfm but, as far as I understood, `FmFolderConfig` is about saving and retrieving of directory settings under `~/.config/libfm/dir-settings.conf`. That's exactly what is done here by `QSettings` and `~/.config/pcmanfm-qt/PROFILE/perfolder-settings.conf`. Both `FmFolderConfig` and `QSettings` support cache. It may even be said that `FmFolderConfig` is a GLib counterpart of `QSettings` but just for folders (am I right?).

Considering this, I wasn't able to see how the code could be simpler or easier to maintain with `FmFolderConfig` instead of `QSettings`. As I see it, it's frugal enough in its current form: there was a conf file for stuck desktop items and now, there's another one for customized folders beside it -- as there would be with `FmFolderConfig` too but in a different place(?)

Please tell me if I'm wrong! No haste, of course.

PCMan:
I'm ok with this change given it works. If other developers found no problems, feel free to merge it.
About FmPath, you might want to try Fm::Path C++ class in libfm-qt which wraps FmPath* raw pointer. It provides auto reference counting and conversion to QString.

PCMan:
BTW, it might not needed to use qHash here.
I'd say it's better to use the path name as a group name, and save all settings for this folder as sub keys of this group.
For example.
[/home/pcman/test-folder]
ViewMode=Icon
Sort=ascending;

[/home/pcman/Desktop]
Sort=mtime;ascending;
This is better for debugging and future extension.
I'm not sure, however, if QSettings allow using / in group name.
GKeyFile on the other hand support this well and have proper support for xdg desktop entry format.

tsujan:
@PCMan 
With QSettings, `/` cannot be used in group name because it's used to reach any sub-key without `beginGroup()`. I'm not a "purist" at all and, as I'd said elsewhere, I used to code in C and GTK+ but, IMHO, using GKeyFile in a Qt code wouldn't be good for readability. Hence using of qHash. There are alternatives but, AFAIK, all of them require converting strings to hex.

tsujan:
Self-criticism: a mechanism for removing non-existent custom folders is needed too.

tsujan:
Completely changed the saving method to a nested QHash:
````
QHash<QString(folder), QHash<QString(property), QVariant(value)>> customFolders
````
It isn't human-readable but should be faster and the list of custom folders is accessible for debugging (`customFolders.keys()`). The deleted folders are also removed from the list on quitting.

**Comment by paulolieuthier** [comment]
I suggest using `QMap` instead of `QHash`, for [it's significantly faster for fewer items](https://woboq.com/blog/qmap_qhash_benchmark.html).

**Comment by stefonarch** [comment]
Testing, that was one feature I was missing.
Just wondering if it would be possible to have a global setting in preferences: preserve folder view and/or order?  Example: picture folders with many subfolders, you have to set it for each one now. Or include subfolders?

**Comment by tsujan** [comment]
@paulolieuthier 
Thanks! The Qt doc says, "QHash provides faster lookups than QMap. (See Algorithmic Complexity for details.)" The article you mentioned says, "Use QMap only... *if you know that you always have a very small amount of items* in your map."

In fact, I first used QMap. With small number of customized folders, the lookup is always fast. My concern was about many of them (which is possible after a while). So I replaced it with QHash. On my older desktop computer, I customized many folders and saw no difference in speed at all (no exact measument, of course).

**Comment by tsujan** [comment]
@stefonarch 
It would be nice to have an option for recursive folder customizability. But before I think about the required code, I have to know how the GUI setting could be. Do you have an idea?

**Comment by paulolieuthier** [comment]
> My concern was about many of them (which is possible after a while). So I replaced it with QHash.

According to the benchmarks in the article, `QHash` is better than `QMap` for more then ~20 items. I went in the opposite direction when developing `PluginSettings` in lxqt-panel, moving from `QHash` to `QMap`. A lookup in a small and balanced tree will be much faster than a hash function, I believe. Anyway, if you tested and saw no difference, than I guess it doesn't matter.

Keep up the nice work!

**Comment by tsujan** [comment]
@paulolieuthier Thanks again for the info!

> QHash is better than QMap for more than ~20 items

I don't think we have a difference of opinions here because the number of custom folders has no upper limit. I already customized ~30 of them just for testing. More tests are welcome.

**Comment by paulolieuthier** [comment]
> I don't think we have a difference of opinions here because the number of custom folders has no upper limit.

Yes, you are right, although the great majority of users will probably use less then 20 IMO. There is no limit for the number of settings for the plugins either, but I optimized for the dominant use case.

**Comment by tsujan** [comment]
As for recursive customization, there are two ways: a global option or a per-folder setting. The former is too inflexible IMO; and I don't know where to add a checkbox for the latter without cluttering menus.

On the coding level, it can be implemented by using `QString::startsWith()` but I have no idea how fast it is for long paths or a great number of custom folders.

IMO, it could be better discussed in another PR.

**Comment by PCMan** [comment]
Some quick comments after code review.

1. I'd suggest removing all of the hash stuff and the complexity unless this is proved to be performance bottle neck. Internally QSettings already uses QMap for quick string lookup. So there is no need to do this again. Calculating the hash value for strings requiring iterating over all characters of all strings, which is not as cheap as it looks like. The internal QMap of QSettings already does a good job for fast lookup, so doing it manually again only brings complexity and break readability of the ini config file.

2. Please store the "columns" and "view mode" as names rather than their numeric IDs. Otherwise, if we change the IDs in the future, the settings will be broken. Please refer to the gtk+ version of pcmanfm, which does use the names of the columns and modes. You can probably copy some code from it.

3. It might be a good attempt to make per-folder setting a separate C++ class and encapsulate implementation details in it. Then the code in MainWindow will look much cleaner.

**Comment by PCMan** [comment]
4. Using QDir here does not always work. It only works for local folders while pcmanfm/libfm supports remote filesystems and things like mtp:/// via gio/gvfs. Using QDir here won't always work.

5. Checking for non-existant dirs is a good idea, but I/O operations are blocking and costly. It's a good idea not to do it too often. Doing in on application exit won't affect startup speed, but in the worst case if the exit is caused by shutdown of the X server, the program is actually receiving SIGKILL so you have no chance to check it.

**Comment by PCMan** [comment]
Wether entries for non-existant dirs should be removed is another issue.
Here are some cases for which the entries should be preserved.
1. remote filesystems, which might be unmounted
2. removable devices that are not always present.
3. volumes that are not mounted at boot time but can be manually mounted later.

Also, when moving existing dirs to another path, we might need to update the setting for it as well. So actually making everything right is not trivial.

**Comment by stefonarch** [comment]
> It would be nice to have an option for recursive folder customizability. But before I think about the required code, I have to know how the GUI setting could be. Do you have an idea?

Could be:   actual checkbox -preserve sorting|view
second checkbox (only active when first is checked): -include subfolders.

Preferences:General: below "Default view mode" a checkbox "Preserve view mode for each folder" o similiar.

Dolphin saves the stuff directly in the folder in .directory, don't know if this is a better approach for speed and folder which changes names or get deleted ecc. for the system directories has to be a list anyway somewhere. Thanks for your work!

**Comment by tsujan** [comment]
@PCMan I'm afraid I should disagree on several points:

1. QHash is used because (a) folder paths cannot be group names and (b) if they're converted to int or hex, there will be no inverse function and so, no way of tracking custom folders or removing deleted folders.

2. Yes, QSettings uses QMap but here, we have only one key, so practically, this is QHash.

3. Human-readability of the conf file is not important in this case. In Qt5, they even change the key order, that is given to the conf file by the user manually, on saving. As far as I remember, their argument was that those files are for saving configuration in the first place.

4. In my first approach, I used strings for columns and view modes. In this second approach, I intentionally used numbers for *more speed*. If IDs are changed in the future, they could be changed here too. After all, we're speaking about libfm-qt, not a third-party library out of our control.

5. Practically, when users want a customized sorting or view, they usually don't think about `mtp:///`, remote filesystems, removable devices or volumes without a fixed mount point.

6. Removal of non-existent folders isn't vital to this. It's just a way of setting an upper limit to the number of custom folders. Therefore, in those "worst" cases, when deleted folders aren't removed from the list, nothing bad will happen.

7. There are two basic approaches for per-folder settings: (a) saving the settings in a file inside the folder; or (b) saving all settings in a conf file. (a) has three drawbacks: root folders couldn't be customized, there would be no way of removing all customizations, and folders would be cluttered with hidden files. (b) has just this drawback that moved or renamed folders lose their custom looks (with the method used here, they'll regain it if put again in their original places with their original names *before quitting*). That doesn't seem a big deal to me.

8. I admit that my code seems unorthodox this time but I defend it by saying that it's compact and efficient enough. Using of a nested QHash doesn't mean complexity. QHash is a 2d table; I needed a 3d table here.

All that being said, whatever you do with this PR will be OK to me. IMO, you have complete right to dislike "weird" coding styles.

**Comment by tsujan** [comment]
@stefonarch
Your first suggestion is acceptable but adding another checkbox worries me a little.

Your second suggestion is good but it isn't about recursiveness.

Perhaps the best way is mixing both suggestions, i.e, adding a check box to Preferences dialog as you told and replacing the currenct menu checkboxes with "Include subfolders" ones.

As for Dolphin, please see point 7 in my reply to @PCMan!

**Comment by tsujan** [comment]
Addendum to point 4 at https://github.com/lxde/pcmanfm-qt/pull/343#issuecomment-218450732:

 Changing IDs will have no effect on this: columnId is read by `TabPage::chdir()` exactly as it's inserted by `MainWindow::on_actionPreserveSorting_triggered()`. The same is true for IconMode, which is written by `MainWindow::on_actionPreserveView_triggered()` and read by `MainWindow::updateUIForCurrentPage()`.

**Comment by PCMan** [comment]
@tsujan Thanks for the detailed explaination. I'll try to merge this PR as long as I have time.
However, I might rewrite the QSettings part to use FmFolderConfig because.
1. Internally,GKeyFile uses a hash table for fast lookup.
2. It supports arbitrary key names including slashes.
3. I already finished a C++ wrapper for it. (automatically generated)
FYI: https://github.com/lxde/libfm-qt/blob/gen-wrapper/libfm-wrapper-gen.py
With this tiny python script, now I can generate C++ wrappers automatically for most parts of libfm core.
It's still a work in progress but it already looks good. :-)
So, no more ugly manual reference counting with C code.

**Comment by PCMan** [comment]
@tsujan this is how the generate C++ code looks like.
http://pastebin.com/R7pM55j2

**Comment by palinek** [comment]
> this is how the generate C++ code looks like.
http://pastebin.com/R7pM55j2

This probably isn't what you've intened it to be :)
```c++
        if(dataPtr_ != nullptr) {
            (dataPtr_);
        }
```

Also the `dataPtr_` member isn't initialized (to `nullptr`) in the c-tors.

PCMan:
@tsujan  Sorry for delaying this for such a long period of time.
Busy with job interviews last month.
Just got a new job at a startup. So hope that I can continue handling this PR this weekend.
Cheers!

tsujan:
@PCMan  No problem at all! I knew you were busy and hope everything is fine with you.

PCMan:
@tsujan I'm going to merge this PR, but if possible, I'd like to use Fm::FolderConfig class to store the settings instead of the current QHash stuff.
Please see the PR: https://github.com/lxde/libfm-qt/pull/37
The new C++ APIs are available.
The idea is, if someday libfm/glib dependency really becomes a problem, we can change the underlying implementation of these C++ class while retain mostly the same APIs.

tsujan:
@PCMan Great! You've lready done a lot of work!

What I can do is just testing the final result and compare it with the previous approach. I have lots of custom folders and, perhaps, could preserve their customizations with some temporary code for testing.

PCMan:
Superseded by #391 .

PR #350, opened 2016-05-06
Fix memory leak in main window tool bar

philippwiesemann:
The replacement main menu had no parent and was therefore not deleted.

Found by Valgrind.

Thanks!


tsujan:
Apparently QAction::setMenu() doesn't give the ownership of a QMenu to a QAction. Hence GTM.

PR #352, opened 2016-05-12
ts-files removal

agaida:
remove only, leave history untouched


agaida:
@paulolieuthier @palinek @luis-pereira - ok with the removal?

PR #359, opened 2016-06-04
Use a wide spread icon theme as fallback for the time beeing

agaida:
The elementary icon theme is only available for ubuntu, Arch, Mandriva
and Gentoo


npmiller:
Wouldn't `hicolor` be a more sensible fallback? It's nearly guaranteed to be installed.

PCMan:
GTM. I'm ok with it.
BTW, hicolor by default contains no icons so it cannot be used.

PR #360, opened 2016-06-04
Auto-hide the filter bar and rework visibility settings

npmiller:
This pull request is a proposed rework of how the filter bar visibility is handled, this was already partially discussed in https://github.com/lxde/pcmanfm-qt/issues/311, but I still think the current way of handling it is not quite right.

So right now we basically have two actions for the filter bar, one that toggles visibility (and changes the settings), and one that focuses the filter bar. Obviously the filter bar can't be focused if it's hidden so the focus shortcut actually also triggers the visibility shortcut (and thus changes the settings), which means that to hide the filter bar again you _have_ to use the visibility shortcut.

I think this is bad because it ties the preference of having the filter bar visible or not by default to the actual use of the filter bar, for example if I close a pcmanfm-qt with an active filter an open it again the filter bar will be visible, this is wrong.

So this pull request:
- Adds a preference whether to filter bar should always be visible or not
- Automatically hides the filter bar if no filter is active (unless the preference is set to always visible)
- Replaces the filter bar visibility shortcut with a clear filter shortcut

Which means that we have one shortcut to start filtering, one shortcut to stop filtering, and a preference for the filter bar visibility, instead of one shortcut to toggle the visibility and one shortcut to start filtering _and_ activate the visibility.

### DIFF
--- a/pcmanfm/main-win.ui
+++ b/pcmanfm/main-win.ui
@@ -177,7 +177,7 @@
     <addaction name="menu_View_2"/>
     <addaction name="actionPreserveView"/>
     <addaction name="separator"/>
-    <addaction name="actionFilter"/>
+    <addaction name="actionClearFilter"/>
     <addaction name="actionMenu_bar"/>
    </widget>
    <widget class="QMenu" name="menu_Editw">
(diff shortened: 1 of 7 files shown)


agaida:
@tsujan - i guess this PR is obsoleted with the now merged changes - right?

tsujan:
@agaida Yes. the transient filter-bar auto-hides and more.

tsujan:
As for network, we can hide items in the side-pane for a long time.

Issue #305, opened 2016-02-02
Navigation icons don't get mirrored in RTL layout


SafaAlfulaij:
In RTL layouts (such as Arabic), navigation arrows (Go Forward/Back and Go) doesn't get mirrored. 
![pcmanfmqticonmirroring](https://cloud.githubusercontent.com/assets/3509617/12745613/45fae0e6-c9ab-11e5-8110-ce1f4dc73911.png)

SafaAlfulaij:
Ping?

tsujan:
Sorry! I'll look into it.

tsujan:
Are all other icons OK for RTL (tab righ-click menu with multiple tabs, for example)?

tsujan:
@SafaAlfulaij 
Please test https://github.com/lxde/pcmanfm-qt/pull/366 if you compile from the source! I could just use the `-reverse` flag to test it.

SafaAlfulaij:
I tested it and didn't work, flipped the two icon variables in the source code and didn't work either. Something is wrong with my configuration...
So, In case that they are mirrored with "-reverse", you can close the bug.
Thanks!

tsujan:
> I tested it and didn't work...

Did you stop the desktop process before testing?

![rtl](https://cloud.githubusercontent.com/assets/981076/16265718/2f585a3a-3896-11e6-9f39-30251730e538.png)

SafaAlfulaij:
Yes of course..
Anyway, it looks fine. But is it possible to mirror the other icons? It's just QIcon -> QPixmap -> QImage -> mirrored function -> QPixmap -> QIcon :blush:
Not all the icons, but the ones that are in the toolbar (New tab, Refresh and go) and "New folder" in the menu. Others won't be easily recognized that they aren't mirrored.

Thanks a lot!

tsujan:
@SafaAlfulaij 
Mirroring is possible but what isn't -- it wouldn't be worth the effort or the amount of code that should be added. Moreover, it wouldn't make much sense either because the other icons are good for RTL and if an RTL user wants different icons, he/she could use other icon sets or edit the current one.

Anyway, I don't understand why https://github.com/lxde/pcmanfm-qt/pull/366 didn't work for you while it works with `-reverse` and even with an on-the-fly change in the layout direction.

SafaAlfulaij:
Got it to work! (Somehow)
Thanks. It's ok leaving the others.

PR #366, opened 2016-06-21
Fix some icons for RTL

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/305.

It also responds to `QEvent::LayoutDirectionChange` but (1) changing the layout direction should be done outside PCManFM-Qt and for all apps, and (2) Qt has small bugs concerning on-the-fly change of layout direction. All in all, IMO, it's good to be thorough when the required changes are small.

### DIFF
--- a/pcmanfm/mainwindow.h
+++ b/pcmanfm/mainwindow.h
@@ -156,7 +156,7 @@ protected Q_SLOTS:
   void toggleMenuBar(bool checked);
 
 protected:
-  // void changeEvent( QEvent * event);
+  void changeEvent(QEvent *event);
   void closeTab(int index);
   virtual void resizeEvent(QResizeEvent *event);
   virtual void closeEvent(QCloseEvent *event);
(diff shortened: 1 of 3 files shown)


f2404:
Would it make sense to support RTL-LTR layout switch at runtime? This patch will change the icons only once, on start, as I understand.

tsujan:
> Would it make sense to support RTL-LTR layout switch at runtime?

Yes, but how?

f2404:
TBH, never thought about that regarding to desktops. On Android, a system notification is broadcasted on a locale change; is there something similar here?

tsujan:
@f2404 It's surely possible. I'll think about it tomorrow. Thanks for the suggestion!

pcmanfm/mainwindow.h
```diff
@@ -156,7 +156,7 @@ protected Q_SLOTS:
   void toggleMenuBar(bool checked);
 
 protected:
-  // void changeEvent( QEvent * event);
+  void changeEvent( QEvent * event);
```
paulolieuthier: `void changeEvent(QEvent *event);`

paulolieuthier:
GTM.

Issue #369, opened 2016-06-26
Location bar: Tilde for homedir?


jimbo1qaz:
Ubuntu Studio 16.04 with Cinnamon and Lubuntu Daily PPA:

Right now the location bar doesn't recognize `~` for homedir. Instead, pressing enter fails to perform navigation.

tsujan:
Yes, that would be a nice feature. I'll work on it.

Issue #201, opened 2015-07-28
ease completion in address bar


pmattern:
Right now there's no means to complete paths in address bar by hitting a single key, in particular `Tab` is keeping its usual functionality to jump between UI elements if the cursor is in address bar.
Also, the drop-down list presenting subdirectories of the directory that's actually in address bar gets displayed only if a trailing slash is suffixed.
Altogether this makes manually filing paths in address bar a rather cumbersome task: the first directory of both absolute and relative paths always has to be typed completely including the trailing slash. Typing following subdirectories can be avoided by picking them from the said drop-down menu only and still the following trailing slashes have to be typed manually.

All in all I'd suggest to make an exception and assign the `Tab` key a completion functionality similar to the one known from shells as soon as the cursor is in address bar. It would be good if this functionality could already be applied to the first directory of the path to be filed, too.

stefonarch:
+1, it's one of the things I miss mostly.

PCMan:
Fixed, finally. :-)
https://github.com/lxde/libfm-qt/commit/374abd283fe131ad2c9f1550fb9355770decc8eb

pmattern:
Unfortunately two problems remain which are pretty much impairing the UX, IMO:

Autocompletion doesn't comprise the trailing slash. E. g. filing `/us` + hitting `Tab` yields `/usr` instead of `/usr/`.
This makes paths comprising several levels of directories much more complicated to enter. Also, the list displaying subdirectories of the current one comes up only when the trailing slash is available (which can be considered expected behaviour).

This list is empty most of the time like in this screenshot
![pcmanfm-qt_addressbar-vs-tab_empty-list](https://cloud.githubusercontent.com/assets/9860801/12532547/fa9359a8-c216-11e5-848c-67d157d757f8.png)
generated after filing `/us`, hitting `Tab` and complementing the trailing slash manually.

While testing I stumbled upon one more thing which would be pretty good to have:
Sometimes it is necessary to interrupt filing a new path, e. g. if a user recalls he still has to look something up in the "old" directory before changing into the "new" one. In such cases it would be very nice if hitting `Esc` restored the old directory's path in the address bar.
Right now PCManFM-Qt keeps displaying the old directory in the actual file manager view but any remnants of a new directory's path that had already been filed in the address bar before hitting `Esc` as well.

[...]

PR #371, opened 2016-06-26
Replace tilde with home path

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/369.

As soon as the user types or pastes a path string starting with a tilde (in the form of `~` or `~/...`) in the location bar, the tilde is replaced by `QDir::homePath()`.

### DIFF
--- a/pcmanfm/mainwindow.cpp
+++ b/pcmanfm/mainwindow.cpp
@@ -382,6 +382,7 @@ void MainWindow::on_actionHome_triggered() {
 
 void MainWindow::on_actionReload_triggered() {
   currentPage()->reload();
+  pathEntry->setText(currentPage()->pathName());
 }
 
 void MainWindow::on_actionGo_triggered() {
--- a/pcmanfm/mainwindow.cpp
+++ b/pcmanfm/mainwindow.cpp
@@ -29,6 +29,7 @@
 #include <QToolButton>
 #include <QShortcut>
 #include <QKeySequence>
+#include <QDir>
 #include <QDebug>
 
 #include "tabpage.h"
@@ -128,6 +129,7 @@ MainWindow::MainWindow(FmPath* path):
   // path bar
   pathEntry = new Fm::PathEdit(this);
   connect(pathEntry, &Fm::PathEdit::returnPressed, this, &MainWindow::onPathEntryReturnPressed);
+  connect(pathEntry, &QLineEdit::textEdited, this, &MainWindow::onPathEntryEdited);
   ui.toolBar->insertWidget(ui.actionGo, pathEntry);
 
   // add filesystem info to status bar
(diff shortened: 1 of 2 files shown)


pmattern:
Not 100% sure whether it's convenient to replace the tilde at once or whether it would make for a better UX to not replace it and just treat the same as `/home/<user>/` or replace it right when when `Tab` is hit to auto-complete digits belonging to a subdirectory that were filed before.

But really not sure, just saying, other than that GTM.

f2404:
Yes, I agree with @pmattern - maybe it's better to leave `~` as is in some cases.

tsujan:
@pmattern and @f2404 
If the tilde was left or not replaced by the home path, auto-completion wouldn't work. Of course, there would be ways for making auto-completion work but they would require changes in libfm-qt and maybe libfm too. Replacing the tilde is the easiest and cleanest way, IMO.

f2404:
@tsujan Nemo is not replacing tilda unless Enter is pressed, and auto-completion is working with tilda.
But I'm ok with it being replaced if it makes the solution cleaner.

tsujan:
@f2404 Actually, I wanted to ask about others' opinions before making this PR but I couldn't see a problem with immediate replacement.
@all If you've found any real problem with that, please tell me -- I mean, a problem that is worth a deeper change in libfm-qt and, perhaps, libfm.

pmattern:
As said above I was only making proposals and certainly wouldn't be opposed to implementing branch `replace_tilde` as is, in particular if alternatives involve major tweaking of libfm[-qt].

tsujan:
@pmattern Now I'm working on https://github.com/lxde/pcmanfm-qt/issues/201, which will require changes in libfm-qt. I still don't know if the tilde could be kept without major tweaks. And your proposal was quite pertinent, of course.

tsujan:
I think this simple approach is practical for now and is better than nothing. I'm suspicious about any benefit coming from a more complex approach that would preserve the tilde. On the other hand, https://github.com/lxde/pcmanfm-qt/issues/201 doesn't seem clear enough to me (https://github.com/lxde/pcmanfm-qt/issues/201#issuecomment-228609406).

@pmattern, should I merge this?

paulolieuthier:
I agree not replacing the tilde is the better way, but if it requires more complexity, we can accept this solution for now.

paulolieuthier:
Does it work well if one types the text very quickly? I can't test right now. If there is lag, we could use a timer to trigger the replacement.

tsujan:
@paulolieuthier 
There's no lag at all. It makes no difference whether you type or paste it. And the tilde should come first to be replaced, of course.

>I agree not replacing the tilde is the better way, but if it requires more complexity, we can accept this solution for now.

Yes, it needs a more complex approach, while this PR consists of a few additions.

paulolieuthier:
GTM.

tsujan:
Thanks! If a simple way of keeping the tilde comes to my mind, I'll make another PR.

PR #374, opened 2016-07-10
I18N: Fix plural string

SafaAlfulaij:
The selected tabs label's string isn't treated as plural.
Just fixed the placeholder according to Qt docs.

### DIFF
--- a/pcmanfm/tabpage.cpp
+++ b/pcmanfm/tabpage.cpp
@@ -434,7 +434,7 @@ void TabPage::onSelChanged(int numSel) {
     else {
       goffset sum;
       GList* l;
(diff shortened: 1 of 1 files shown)


pcmanfm/tabpage.cpp
```diff
@@ -434,7 +434,7 @@ void TabPage::onSelChanged(int numSel) {
     else {
       goffset sum;
       GList* l;
-      msg = tr("%1 item(s) selected", NULL, numSel).arg(numSel);
+      msg = tr("%n item(s) selected", NULL, numSel).arg(numSel);
```
paulolieuthier: Please, use `nullptr` instead of `NULL`.

paulolieuthier:
Thanks!

Issue #379, opened 2016-07-30
Scrambled statusbar message for files with RTL names


tsujan:
This image explains everything:

![1](https://cloud.githubusercontent.com/assets/981076/17270283/665b03ec-5674-11e6-8b9d-db622e1f443d.png)

I couldn't find a way of showing the message correctly for both LTR and RTL layouts, especially with mixed strings (I even tried `0x202E` and `0x202D`).

SafaAlfulaij:
Please ping me in case you are stuck with RTL stuff ;)
This is an easy task:

LTR layout:
`LRM + "\"%1\"" + LRM + "(b...."`
RTL layout:
`RLM + "\"%1\"" + RLM + "(b...."`

For the file name, if `isRightToLeft` then you should prepend and append RLM to it, if it's not then LRM.
And that should fix it :)
RTL filename: `RLM + %1 + RLM`
LTR filename: `LTR + %1 + LTR`

While I'm not sure if we really need to prepend the RLM/LRM, need to be tested by all possibilities :)

SafaAlfulaij:
But we have to think to not make string puzzles for the translators, that is not a good i18n practise.

tsujan:
@SafaAlfulaij 
Thanks for the hint, but aren't `RLM` and `LRM` the same as `0x202E` and `0x202D` respectively? As I said above, I'd already tried them (in the same way you suggested). They worked with "pure" RTL strings but with mixed ones they caused a mess.

As for translation, I agree with you. However, the rest of tthe statusbar message doesn't seem translatble to me because of its source in libfm.

tsujan:
OK, `0x202E` and `0x202D` are RLO and LRO respectively, while RLM and LRM are `0x200F` and `0x200E`. I'll try them tomorrow.

tsujan:
@SafaAlfulaij
I reached a seemingly acceptable formula at https://github.com/lxde/pcmanfm-qt/pull/380. Would you please test it?

Please note that:

(1) This is for both LTR and RTL layouts;

(2) The "underscore problem" persists because of a problem like QTBUG-55012 for label texts; and

(3) The info part of message is always in English for now but a future translation for RTL languages is predicted in the commit.

tsujan:
Two screenshots, one for ltr layout:

![ltr](https://cloud.githubusercontent.com/assets/981076/17275051/5bde5b58-570f-11e6-998b-ac8f36957305.png)

And the other for rtl layout:

[...]

PR #380, opened 2016-07-31
Fix status message for RTL file names

tsujan:
Tries to fix https://github.com/lxde/pcmanfm-qt/issues/379.


tsujan:
Closed because a clean approach with a simple logic doesn't seem possible for now.

PR #381, opened 2016-08-02
Fixes two typos

tsujan:
On of them was old; the other was made recently (which commit?)

### DIFF
--- a/pcmanfm/tabpage.cpp
+++ b/pcmanfm/tabpage.cpp
@@ -420,7 +420,7 @@ void TabPage::onSelChanged(int numSel) {
       if(size_str) {
         msg = QString("\"%1\" (%2) %3")
                         .arg(QString::fromUtf8(fm_file_info_get_disp_name(fi)))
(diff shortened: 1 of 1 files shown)


tsujan:
Could someone confirm that `%1` was changed to `%n` by mistake? It was `%1` originally.

paulolieuthier:
GTM.

tsujan:
@paulolieuthier
It was https://github.com/lxde/pcmanfm-qt/commit/537bfc2b904fd9520fbb23f77a24e1bfb64bbb9d. `tr()` was changed to plural but the argument wasn't removed. Am I right?

paulolieuthier:
I wasn't sure about that, but after reading http://doc.qt.io/qt-5/i18n-source-translation.html#handling-plurals, I'm sure you are right.

tsujan:
> I wasn't sure about that, but after reading http://doc.qt.io/qt-5/i18n-source-translation.html#handling-plurals, I'm sure you are right.

Yes, that was exactly what I read too because I wasn't sure either ;)

PR #382, opened 2016-08-02
Fix direction of statusbar message

tsujan:
Event for LTR layouts, a file can have an RTL name or an LTR name with an RTL component. So, the direction of statusbar text should be corrected with LRM/RLM and LRE/RLE; otherwise the message would be scrambled in such cases.

### DIFF
--- a/pcmanfm/tabpage.cpp
+++ b/pcmanfm/tabpage.cpp
@@ -408,6 +408,17 @@ void TabPage::onOpenDirRequested(FmPath* path, int target) {
   Q_EMIT openDirRequested(path, target);
 }
 
+/* This is for creating texts with correct directions and is also needed
+   with LTR layouts and texts because LTR texts can contain RTL strings. */
+/*static*/ void TabPage::directedText(QString &txt, QChar startMark,QChar endMark) {
+  // mark (LRM/RLM) and embedding (LRE/RLE) directions
+  QString mDir = txt.isRightToLeft() ? QChar(0x200F) : QChar(0x200E);
+  QString eDir = txt.isRightToLeft() ? QChar(0x202B) : QChar(0x202A);
+  txt = (startMark.isNull() ? QString() : mDir + QString(startMark))
+         + eDir + txt
+         + (endMark.isNull() ? QString() : mDir + QString(endMark));
+}
+
 // when the current selection in the folder view is changed
 void TabPage::onSelChanged(int numSel) {
   QString msg;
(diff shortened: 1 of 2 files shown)


SafaAlfulaij:
Why didn't it get landed? :)

tsujan:
>Why didn't it get landed? :)

Do you confirm that the problem is fixed by it *as far as possible*?

SafaAlfulaij:
> Do you confirm that the problem is fixed by it as far as possible?

If it's possible to tell me where is the "bytes" string, I can give you a final word! ;)

tsujan:
All is done in `TabPage::directedText()` but my question was about the *result*, as I explained in https://github.com/lxde/pcmanfm-qt/issues/379#issuecomment-236907524 and the comments following it.

SafaAlfulaij:
A final word: **commit it**!
Until I find a clean solution, this can be really enough!
Thanks! :)

tsujan:
Although, IMO, this is a clean and logical solution, I prefer not to merge it until it's accepted by others too.

PCMan:
@tsujan Great work! I'm not an RTL user but I really want to understand the problem better.
@SafaAlfulaij Would you please help provide a screenshot to demonstrate the differences?
Thank you.

tsujan:
@PCMan 
As far as I remember, the problem was about mixed RTL and LTR and this commit fixed it. @SafaAlfulaij's opinion was a little different.

I should examine the problem again (when I have time) to tell you exactly how this commit fixes it with some screenshots.

tsujan:
@PCMan 
https://github.com/lxde/pcmanfm-qt/issues/379 has a few screenshots. The problem was about strings that had both RTL and LTR parts. Specifically, the statusbar text can be mixed when a file name is RTL. As my first screenshot on that page shows, the text was scrambled before. This commit fixes the problem by using LRM/RLM and LRE/RLE characters. All situations are considered here, as far as I remember.

Anyhow, I think an RTL user should confirm the fix before this PR can be merged. The branch has a minor conflict too.

PCMan:
@tsujan @SafaAlfulaij any news about this one?
RTL support is a must-have, but inserting the unicode markers everywhere manually creates problems for developers. It's surprising that Qt does not handle this automatically since according to the API doc this kind of things should be handled. :-(
http://doc.qt.io/qt-5/internationalization.html

tsujan:
@PCMan 
I agree that Qt should have handled this. Unfortunately, it doesn't. I also agree that inserting unicode markers is just a workaround. We could leave this PR open until Qt's problem is fixed or a better method is found.

In short, the current commit works but I don't like it either.

tsujan:
Closing this; it's too old.

PCMan:
@tsujan If no other better solutions are found so far and the current one works, maybe we can merge this PR. I also took some time to dig into the Qt RTL issues, but I did not find any more elegant ways to workaround this.

PCMan:
@tsujan  If the workaround works correctly, even if it's ugly, I think it's acceptable.

tsujan:
@PCMan I really don't think this was needed in the first place. If someone reports a bug, I'll make an up-to-date PR.

PR #385, opened 2016-08-03
Use LXQtCompilerSettings cmake module

luis-pereira:
We already have an build time dependency on liblxqt. So, let's take all the
advantages we can.


paulolieuthier:
GTM.

tsujan:
GTM

PR #386, opened 2016-08-03
Fix -Wswitch warning

luis-pereira:
As we are only handling two enumeration cases, just replace it by an if
statement.


paulolieuthier:
GTM.

tsujan:
GTM.

luis-pereira:
Closed by https://github.com/lxde/pcmanfm-qt/commit/523739b093db1cc29c78baac85e2abce558ed8d5

PR #388, opened 2016-08-09
Add Catalan translations

rbuj:


PCMan:
@rbuj Would you please separate the *.ts file and *.desktop file according to the guide here?
https://github.com/lxde/lxqt/wiki/Translation
We moved all *.ts files to lxqt-l10n repo instead.
Thank you.

rbuj:
@PCMan There is no TS file in this PR. There are only two files, and both are desktop files.
Catalan translation of TS file was merged in https://github.com/lxde/lxqt-l10n/commit/b0650bcc2555c2df338fb1e8f3a8184efde6e751

I'm sorry, but I don't understand you. Please, can you tell me what is wrong?

Thanks in advance.

PCMan:
Sorry, I mistakenly mixed this one with another PR which contains a ts file.
This one looks good and there's no problem. Thank you.

PR #391, opened 2016-08-24
Per folder settings: a cleaner and simplified implemention.

PCMan:
This new implementation supersedes the old PR #343 with the following changes.

1. Encapsulation with PCManFM::FolderSettings class and move most of the handling into PCManFM::TabPage rather than adding duplicated pieces of code in MainWindow.
2. Remove the code in onAboutToQuit since it does not work. For example, folders on removable devices  or remote filesystems might have their folder temporarily unavailable, but their per-folder settings should not be deleted.
3. Do not distinguish between per-folder sort settings and per-folder view settings. Separating them is too fine-grained and makes the code complicated. The gtk+ version of pcmanfm uses only one option to control them and this works well enough.
4. Re-use Fm::FolderConfig internally to store the settings.
5. Remove tear-down menus which is an out-dated UI style inconsistent with the rest parts of the DE.
6. Do not change the default view mode when you change the view mode of the current tab. This is a "feature", not a bug. The default view mode is configured in the preferences dialog.

@tsujan I merged your code last week, but after some more code review I found some issues listed above. So later I decided to do some refactor and rewrites.

### DIFF
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -349,24 +349,6 @@ void Application::onUserDirsChanged()
 
 void Application::onAboutToQuit() {
   qDebug("aboutToQuit");
-  // remove non-existent folders from the list of custom folders
-  QString configFile = QString("%1/perfolder-settings.conf").arg(settings_.profileDir(settings_.profileName()));
-  QSettings folderSettings(configFile, QSettings::IniFormat);
-  QHash<QString, QVariant> customFolders = folderSettings.value("customFolders").toHash();
-  if(!customFolders.isEmpty()) {
-    //qDebug() << "Custom folders: " << customFolders.keys();
-    QHash<QString, QVariant>::iterator i = customFolders.begin();
-    while (i != customFolders.end()) {
-      QDir dir(i.key());
-      if(!dir.exists()) {
-        i = customFolders.erase(i);
-      }
-      else ++i;
-    }
-    folderSettings.setValue("customFolders", customFolders);
-  }
-  if(customFolders.isEmpty())
-    folderSettings.clear();
   settings_.save();
 }
 
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -349,6 +349,24 @@ void Application::onUserDirsChanged()
 
 void Application::onAboutToQuit() {
   qDebug("aboutToQuit");
+  // remove non-existent folders from the list of custom folders
+  QString configFile = QString("%1/perfolder-settings.conf").arg(settings_.profileDir(settings_.profileName()));
+  QSettings folderSettings(configFile, QSettings::IniFormat);
+  QHash<QString, QVariant> customFolders = folderSettings.value("customFolders").toHash();
+  if(!customFolders.isEmpty()) {
+    //qDebug() << "Custom folders: " << customFolders.keys();
+    QHash<QString, QVariant>::iterator i = customFolders.begin();
+    while (i != customFolders.end()) {
+      QDir dir(i.key());
+      if(!dir.exists()) {
+        i = customFolders.erase(i);
+      }
+      else ++i;
+    }
+    folderSettings.setValue("customFolders", customFolders);
+  }
+  if(customFolders.isEmpty())
+    folderSettings.clear();
   settings_.save();
 }
 
--- a/pcmanfm/main-win.ui
+++ b/pcmanfm/main-win.ui
@@ -145,6 +145,9 @@
      <string>&amp;View</string>
     </property>
     <widget class="QMenu" name="menuSorting">
+     <property name="tearOffEnabled">
+      <bool>true</bool>
+     </property>
      <property name="title">
       <string>&amp;Sorting</string>
      </property>
@@ -159,16 +162,29 @@
      <addaction name="separator"/>
      <addaction name="actionFolderFirst"/>
      <addaction name="actionCaseSensitive"/>
+     <addaction name="separator"/>
+     <addaction name="actionPreserveSorting"/>
+    </widget>
+    <widget class="QMenu" name="menu_View_2">
+     <property name="tearOffEnabled">
+      <bool>true</bool>
+     </property>
+     <property name="title">
+      <string>&amp;View</string>
+     </property>
+     <addaction name="actionIconView"/>
+     <addaction name="actionThumbnailView"/>
+     <addaction name="actionCompactView"/>
+     <addaction name="actionDetailedList"/>
+     <addaction name="separator"/>
+     <addaction name="actionPreserveView"/>
     </widget>
     <addaction name="actionReload"/>
     <addaction name="actionShowHidden"/>
     <addaction name="separator"/>
-    <addaction name="actionIconView"/>
-    <addaction name="actionThumbnailView"/>
-    <addaction name="actionCompactView"/>
-    <addaction name="actionDetailedList"/>
-    <addaction name="separator"/>
+    <addaction name="menu_View_2"/>
     <addaction name="menuSorting"/>
+    <addaction name="separator"/>
     <addaction name="actionFilter"/>
     <addaction name="actionMenu_bar"/>
    </widget>
@@ -531,6 +547,14 @@
     <string>&amp;Folder First</string>
    </property>
   </action>
+  <action name="actionPreserveSorting">
+   <property name="checkable">
+    <bool>true</bool>
+   </property>
+   <property name="text">
+    <string>&amp;Preserve sorting for this folder</string>
+   </property>
+  </action>
   <action name="actionNewTab">
    <property name="icon">
     <iconset theme="window-new">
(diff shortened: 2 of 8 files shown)


PCMan:
@tsujan this might still contain some bugs, but the code is now organized in a more OO way.
Would be glad if you can help review the code. Thank you.

tsujan:
@PCMan
I'll study and test the code as soon as I have time.

However:

> Remove the code in onAboutToQuit since it does not work

Without it, a lot of unused entries may be left after a while, and that could affect performance. A config file that grows indefinitely is not a good idea, IMHO. On the other hand, users have no urgent need to customize folders related to removable devices or remote filesystems, exactly because they're temporary.

> Remove tear-down menus which is an out-dated UI style inconsistent with the rest parts of the DE.

I use my previous PR since I made it and have seen the need for that tear-off. Without it, the user should open the menu twice, which would make customization painful - unless you've done it differently (I haven't tried your code yet).

> Do not change the default view mode when you change the view mode of the current tab. This is a "feature", not a bug.

I'm afraid it could be confusing to the user when he/she changes the view mode and sees that nothing has changed with other folders -- and not even with same folder later (it was confusing to me). Being confusing, it can't be seen as a feature, IMO.

PCMan:
@tsujan : 
1. Windows does not clean its per-folder settings, either. We can try to find a better way to remove the dead entries. For example, record the last used time and set an expiration date. Otherwise provide a way for the users to remove the dead entries in the preferences dialog.
In addition, mounted remote filesystems, such as SFTP, might not be temporary. They might become unavailable simply because you don't have an internet connection.

2. onAboutToQuit() is not always called. It's only called if the program is terminated by SIGTERM. However, the program is mostly terminated by SIGKILL from killing the Xorg server.
Anyway, this can be solved in later releases. I don't see this an urgent issue. As these per-folder settings are added one by one manually, during normal daily use, it should not grow quickly.

3. Tera-off menus are removed from gtk+ a while ago. Other modern platforms or DEs do not have this UI style, either. For consistency reason, we don't use it.
FYI: http://doc.qt.io/qt-5/qmenu.html#actions  The Qt API doc stressed that this style is not used in Windows and some users may not be familiar with it. "Consider using a QToolBar instead."

4. The setting of "default view mode" is used by "Thunar",  the FM of XFCE and the older versions of Nautilus. Since this is used by two of the leading DEs, it's some what common.
Sometimes you only want to change the view of current tab temporarily, but this affect all new tabs, which can be annoying and confusion, too.

stefonarch:
I like the simple setting as implemented here, but I'm missing an "include subfolders" checkbox, best example my ~/foto/yyyy/mm directories, it  would be perfect  to be able  alltogether to "thumbnail view" with to clicks.

tsujan:
@PCMan
I tested the PR and encountered 6 issues. I'm afraid some of them can't be fixed here.

(1) `~/.config/libfm` isn't created (WARNING **: cannot save /home/pedram/.config/libfm/dir-settings.tmp: Failed to create file '/home/pedram/.config/libfm/dir-settings.tmp.K36MMY': No such file or directory).

(2) The list is saved only on quitting, which can't be good to you for the same reason you removed the previous code in onAboutToQuit.

(3) There is no distinction between profiles ("default" and "lxqt").

(4) `Folder First` isn't and can't be selected for non-customized folders.

(5) The `Folder First` property (not its checkbox) changes randomly, i.e., files come between folders randomly.

(6) I'm not sure about this one but I may have seen a sort of lagging in loading a folder.

There may be other problems too -- these were what I encountered in a few minutes.

palinek:
Just a note: 

> onAboutToQuit() is not always called. It's only called if the program is terminated by SIGTERM. However, the program is mostly terminated by SIGKILL from killing the Xorg server.

I don't think, that violent kill of X server should be considered as normal shutdown. Yes, there was a problem in lxqt-session which got fixed in https://github.com/lxde/lxqt-session/commit/302843e86512b38f9809c2119f15e02ffb6c3629. So normaly the process should be stopped correctly.

PCMan:
@tsujan I updated the PR.
1. ~/.config/libfm dir is created if it does not exist.
2. it's not only saved on quitting, but also saved when the settings of a tab page is changed.
3. yes, there are no distinctions for profiles. we can add that support to libfm later.
4 & 5. folder first is fixed.
6. I don't encounter performance problems here.
Thanks.

tsujan:
@PCMan Thanks! I'll test the changes soon.

tsujan:
@PCMan 
The problems about the libfm folder and "folder first" are fixed now.

I completely agree about the cleanness of this approach on the coding level and appreciate your work. However, I can't ignore these issues:

(1) Customizing folders is like a cumbersome ritual because menu tear-offs are removed. I don't get your point: Just because gtk+ removed them or MS Windows can't have them, we shouldn't use them?! Gnome3 developers (who develop GTK too) have removed many useful features from GTK; this isn't the only one. And why everything "modern" (new) should be good?

(2) *Practically*, the list is saved only on quitting -- saving on changing tab settings is not practical because the user may never do so. But *I* have no problem with that because, as @palinek said above,  killing of X server shouldn't be considered as a normal shutdown. That being said, I'd prefer immediate saving, which is immune to crash

(3) I still think that an ever-growing config file is a flaw. IMO, MS Windows is not a good model to follow. Providing a way for the users to remove the dead entries is not an option; such things should be done automatically. I don't get the importance you gave to customizing mounted remote filesystems and why we should have an ever-growing file just because of them.

tsujan:
I forgot to add that my concern is not about the *file* size but about the *list* size. On entering every folder, pcmanfm-qt should see if it's among the customized folders. With that human-readable format, I think an ever-growing list could have a negative impact on performance.

tsujan:
My last comment on this.
Just this tear-off makes a *huge* difference in practice:
tear-off

PCMan:
@tsujan I go some more answers for you.
1. Tear-off menu is not bad. It's just about UI consistency across different parts of the DE. If this UI style is used widely in LXQt, we use it. Otherwise the only component using this feature can look like an alien. If something is frequently accessed, add a tool button/menu instead.

2. The settings are saved on changes, not on closing.
For example:
https://github.com/lxde/pcmanfm-qt/pull/391/files#diff-a3ef47f1639253aa169b271696f97e75R584
When the view settings of a tab is changed, the changes are saved.

3. Hash table lookup is an O(1) operations which do not grow linearly to the file size. The really time consuming stuff is hash table creation in this case. Try with the following benchmarks by yourself.
  a. generate a huge config file by listing all dirs in your /usr/share:
`
for path in `find /usr/share -type d`; do
  echo -e "[$path]\nViewMode=detail\nSortOrtder=ascending\nShowHidden=flse\nSortFolderFirst=true\nSortCaseSensitive=false\n\n" >> test_keyfile.conf;
done
`
I have 10852 entries in the config file by doing this, which is quite a lot.

Then, test with the following code.
`
#include <glib.h>
int main(int argc, char** argv) {
        GKeyFile* kf = g_key_file_new();
        g_key_file_load_from_file(kf, argv[1], 0, NULL);
        g_key_file_free(kf);
}
`
Compile:
> gcc `pkg-config glib-2.0 --cflags --libs` test_keyfile.c
Drop all filesystem caches:
# echo 3 > /proc/sys/vm/drop_caches
Run a.out without the input file to get "glib" cached in the filesystem. Otherwise most of the time will be spent on loading the library, not on loading the config file.

Then, test it:
> time ./a.out test_keyfile.conf
`
[pcman@arch-pc ~]$ time ./a.out test_keyfile.conf 

real	0m0.087s
user	0m0.053s
sys	0m0.013s
[pcman@arch-pc ~]$ 
`
Then, tell me, how can this affect "performance"?

We are computer scientists. In terms of performance, do measurements and complexity analysis rather than assuming.

Thank you.

PCMan:
In the future, we can still add some mechanisms to remove dead entries for the users. Right now, just ship it first.

tsujan:
@PCMan
OK. It's your program after all. My only intention has been making it better and handier.

> We are computer scientists.

I don't see myself as a computer scientist. My background is in mathematics and physics; because of that, I found computer programming so familiar years ago. Hence my following the principle of least action ;)

PCMan:
@tsujan Sorry. I'm not trying to be offensive. I know you're here to help and really appreciate that.
It's always a good thing that people care about performance. Increased amount of code almost always have some kind of impact. So it's better to get a feel of how much the impact really is and we don't waste time on micro-optimizations.
I agree with you that a config file should not grow indefinitely. I just tried to show you that it's not very urgent since a huge config file does not affect much, so we can improve it later.

tsujan:
> Sorry. I'm not trying to be offensive.

No offence taken! We just had very different opinions on this matter.

> I know you're here to help

I have a good nose for good apps and this is one them. It's fun to make it better.

tsujan:
@PCMan
Just clearing up a mutual misunderstanding after reading the code:

By saving I meant *saving to file*. It's done only here:
````
void Application::onAboutToQuit() {
  qDebug("aboutToQuit");
  settings_.save();
}
````
Now I inderstand why you told me that settings were saved on changes: you meant in the memory, while I meant to the config file (as in the previous PR).

As I said before, IMO, this isn't a problem at all.

PR #396, opened 2016-09-06
wallpaper: Add helper logic for LXQt

palinek:
Store "wallpaperChanged" value into LXQt config
($XDG_CONFIG_HOME/lxqt/lxqt.conf) if wallpaper was not changed by LXQt
theme.

With this help the lxqt-config-appearance can do the decision to (not)
change the wallpaper when changing the LXQt theme.

Needed by lxde/lxqt-config#99

### DIFF
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -53,6 +53,7 @@
 
 #include "xdgdir.h"
 #include <QFileSystemWatcher>
+#include <QSettings>
 
 namespace PCManFM {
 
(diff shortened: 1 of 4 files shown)


paulolieuthier:
Ideally, pcmanfm-qt should continue DE-agnostic, but I think this is OK for now, as we need this fix. GTM.

paulolieuthier:
Ping @tsujan @PCMan.

tsujan:
This may not make pcmanfm-qt dependent on LXQt but can be seen as the first step in that direction. IMHO it's better to keep pcmanfm-qt completely DE-agnostic. (Right now, I'm using it under Enlightenment.)

tsujan:
I know this is for fixing a problem and the cod is OK in itsef but I can't help seeing it as an "impurity" in pcmanfm-qt's code. I think @PCMan should make the final decision.

paulolieuthier:
The problem is that there is no standardization for wallpapers. We should try to make it in the near future.

tsujan:
I'm not familiar with lxqt-config's code; so sorry if this seems silly:

Can't lxqt-config simply read `~/.config/pcmanfm-qt/lxqt/settings.conf`, see if the active wallpaper is inside a theme directory, and only if it is, change it on changing themes?

Issue #417, opened 2016-12-03
Recursive Permissions Possible?


FlailAway:
New user of PCmanFM (I have even un-installed Nautilus!) and so far I think it is **really** good.

Is there some way to ripple a Permission change throughout a folder structure to all sub folders and files?

I just copied some stuff from an old backup and all the folders are locked to "user-read" and I need to change them to "user read/write" but I am having to go through all of the folders and files to individually change the Permissions. That's getting boring really fast. :-1: 

Is there some option or plugin that I am not seeing that will do this for me?

Thanks

tsujan:
Do you mean pcmanfm-qt or pcmanfm? They're different. This site is for the former and there's still no way to change the permissions of all contents of a folder from the GUI.

tsujan:
Anyhow, this is a good idea to implement, IMO.

FlailAway:
Thanks for the reply, not sure, but the "Help/About" says 
"PCManFM 1.2.0 ... Developed by Hon Jen Yee (PCMan)"
With a link to this page
https://sourceforge.net/projects/pcmanfm/
I installed it in Ubuntu 14-4 using the Software Center.

tsujan:
It's pcmanfm, not pcmanfm-qt.

Although you came to the wrong site, please don't close this thread because the idea is good.

FlailAway:
OK, thanks, I will install the "-qt" version and wait for the feature. :)

tsujan:
This has been possible for a long time (file properties dialog asks about recursiveness).

PR #421, opened 2016-12-08
Make dialogue "About" point out it's the Qt version of PCManFM

pmattern:
So far the dialogue is only stating "PCManFM" which may trick users into thinking the GTK version was running. Happened to crop up some days on the tracker as well, see #417.

So this PR is simply proposing to state "PCManFM-Qt" in order to avoid such confusion.


tsujan:
GTM

agaida:
GTM

PR #423, opened 2016-12-12
Wallpaper Slide Show

tsujan:
Fixes https://github.com/lxde/lxqt/issues/608.

The new options are added to Desktop Preferences. The slide-show interval is between 5min and 24h,55min. An option is also added for randomizing the slide show.

### DIFF
--- a/pcmanfm/desktop-preferences.ui
+++ b/pcmanfm/desktop-preferences.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>534</width>
-    <height>500</height>
+    <height>434</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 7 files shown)


agaida:
Compiles and works good - GTM

pmattern:
Works as expected here, too. Some additional thoughts:

Minimum intervals <5min would be good - and be it for testing only. Default could still stay 5min, though.

The minimum size of dialogue "Desktop Preferences" has become a bit large due to the additional settings. It's barely fitting on the desktop at 1024x768.
I'd suggest summarizing all settings affecting the background in a new tab "Background" (leaving all the rest in tab "General" and tab "Advanced" altogether untouched).
A new tab like this should probably best be available as argument of command line switch `--desktop-pref`, too.

Right now only files in the chosen folder itself are considered. Wouldn't it be good to recursively search through subfolders it may have, too?
If so, this should certainly be an option, of course.

When random selection is disabled the slideshow always starts over at the first image when a new LXQt session is begun. Wouldn't it be nice if the image shown last in the previous session was memorized and this image or its successor was used as first in the upcoming session?
(Random selection doesn't seem to be affected but is a bit nasty to test right now due to the ≥5min. intervals.)

tsujan:
> Minimum intervals <5min would be good - and be it for testing only. Default could still stay 5min, though.

When I tested the code, I had to manually change it to seconds. If already not done, I'll reduce it to a single variable for testing and will tell you what it is.

> I'd suggest summarizing all settings affecting the background in a new tab "Background"

OK.

> Right now only files in the chosen folder itself are considered. Wouldn't it be good to recursively search through subfolders it may have, too?

I'll think about it.

> When random selection is disabled the slideshow always starts over at the first image when a new LXQt session is begun.

Yes! I thought random selection was enough for not seeing the first wallpaper on login, but you're right; it's better to remember the last shown wallpaper and show its following wallpaper on the next login when randomizing is disabled. KDE does so too.

Thank you for the suggestions!

tsujan:
@pmattern 

While implementing your ideas, I also found some bad codes and corrected them. I'll push the changes after one day of testing.

BTW, recursively searching for wallpapers may be CPU-intensive because we don't know how much deep folders are and how many images they contain. It's also overkill, IMO.

pmattern:
> BTW, recursively searching for wallpapers may be CPU-intensive because we don't know how much deep folders are

Couldn't we just pick one file after the other, then proceed to the next folder and go on, so without scanning the whole directory all at once in the first place?

> It's also overkill, IMO.

Maybe. Also, I couldn't even tell whether other implementations of slideshows are providing this.
The idea was that a feature like this would allow for using a preexisting image folder. If on the other hand a single folder is scanned only this folder would certainly have to be maintained.

tsujan:
@pmattern 

I'm sure that I added a comment but it isn't shown by QupZilla. Maybe I wrote it somewhere else!

Anyhow, would you please test again? I did as you told -- except for recursive search, which can be CPU intensive (also see below)

> Couldn't we just pick one file after the other

The order (after, before) will be determined only if we know about all images; hence the scanning.

pmattern:
After setting
```
#define MIN_SLIDE_INTERVAL 60000 // 1 min
```
in `./pcmanfm-qt/pcmanfm/desktopwindow.cpp` the shortest intervals eligible in tab "Slide Show" of dialogue "Desktop Preferences" still are 5 and 0 min respectively here.

As for dialogue "Desktop Preferences" I really think all settings dealing with the background should be summarized in that new tab, so section "Background" of current tag "General", too. That new tab should probably be called "Background" then, see [above](https://github.com/lxde/pcmanfm-qt/pull/423#issuecomment-277555202).
Also, a counterpart of arguments "general" and "advanced" of CLI switch `--desktop-pref` seems still to be missing.

The image shown last when an LXQt session is quit is now memorized and the slide show will chose the next image when the next session is launched.
This applies as well when the slide show is restarted within a running session, e. g. by temporarily disabling it. I had never thought about this use case but I think the current behavior is absolutely okay. So just saying.

Little weirdness: When the slide show is enabled by checking the box and hitting "Apply" the next image is chosen when the dialogue is quit by hitting "OK".
Not a bummer but I think getting this fixed would improve the UX.

If you say recursive usage of an image folder is by all means involving too high CPU load then I find it a pity but it's okay.

tsujan:
> the shortest intervals eligible in tab "Slide Show" of dialogue "Desktop Preferences" still are 5 and 0 min respectively here.

That's intentional. MIN_SLIDE_INTERVAL is just for testing.

> As for dialogue "Desktop Preferences" I really think all settings dealing with the background should be summarized in that new tab.

When the user sets background, he/she doesn't need to see more complex settings of slide show. IMO, the current layout is the best: 1st tab is for basic settings; the 2nd for slide show; and the 3rd is redundant ;)

> Also, a counterpart of arguments "general" and "advanced" of CLI switch --desktop-pref seems still to be missing.

I've never used `pcmanfm-qt --desktop-pref`. IMO, it's redundant and buggy (`pcmanfm-qt --desktop-pref advanced` should be `pcmanfm-qt --desktop-pref Advanced`, ...). But I'll look into it.

> When the slide show is enabled by checking the box and hitting "Apply" the next image is chosen...

KDE does so too. When apply is clicked, everything is updated. Slide show just isn't an exception.

Thanks for testing!

pmattern:
> When the user sets background, he/she doesn't need to see more complex settings of slide show. IMO, the current layout is the best: [...]

Disagree. The settings of slide show aren't exactly rocket science either so I certainly wouldn't call them "complex". But the slide show and the background are closely related as the latter is replacing the former if enabled.
I do agree that a single item doesn't warrant a tab on its own.

> KDE does so too. When apply is clicked, everything is updated. 

Probably a misunderstanding. The point is not what happens when "Apply" is clicked. The point is that another image is chosen when the dialogue is left by clicking "OK".
So check box "Enable Slide Show" when it isn't checked / the slide show isn't active and hit "Apply" → slide show starts with some image. Leave dialogue "Desktop Preferences" by hitting "OK" → the next image is chosen.
And the latter doesn't make sense, does it?

tsujan:
> The settings of slide show aren't exactly rocket science

It isn't but it has a checkbox that enables/disables it. I think our disagreement is just a matter of taste -- we can't convince each other.

> The point is not what happens when "Apply" is clicked. The point is that another image is chosen when the dialogue is left by clicking "OK"....

KDE does the same thing and no one has found it disturbing. OK and Apply both **update** settings. Cancel does nothing.

pmattern:
> I think our disagreement is just a matter of taste

True.

> KDE does the same thing

No, it doesn't.
When you configure a slide show from the context menu's topic "Configure Desktop" and hit "Apply" the slide show starts. When you leave the dialogue *right after this* by clicking "OK" the dialogue is closed but clicking "OK" does not trigger the display of the following image.
This is the default behavior of virtually all dialogues I've seen so far. After settings have been applied by hitting "Apply" hitting "OK" without tweaking any additional settings closes the dialogue but does not trigger anything else.

tsujan:
The ok and apply buttons both call `Application::updateDesktopsFromSettings()` → `DesktopWindow::updateFromSettings()`. On the other hand, when desktop starts, the same function is called. This is the code structure that existed BEFORE this PR. EVERYTHING IS UPDATED on clicking ok and apply -- you just see the slide-show part **now**.

If this is a problem, it isn't related to this PR. In another PR, we could decide whether OK should update anything after Apply.

BTW, you made me log out and log into KDE and see it doesn't do so anymore but I'm sure it did before beause I always used slide show with KDE.

tsujan:

Issue #431, opened 2016-12-19
Regression: Filter is erased on tab change with path buttons


tsujan:
When path buttons are used, the filter string of a tab is erased on changing tabs and returning to it.

The filter string is preserved when path-edit is used.

PR #432, opened 2016-12-19
Prevent wrong clearing of filter-bar with path buttons

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/431.

With path buttons, filter-bar was cleared on changing the current tab because libfm-qt uses `QPushButton::toggled` as the signal and so, `MainWindow::chdir()` was always called when path buttons changed their toggled state. This patch prevents redundant calls to `MainWindow::chdir()`.


agaida:
GTM

Issue #428, opened 2016-12-16
Tab Drag-and-Drop


tsujan:
Tab detachment and DND is a feature that I think I can implement. Does anyone support this feature? Is it useful to you? I ask because I can live without it ;)

tsujan:
This is what I mean: [packaged video](https://www.dropbox.com/s/9scitbspke2ql3u/video.tar.gz?dl=0)

agaida:
detachable Tabs are nice - not that i would use them often, but from time to time - like detachable tabs in browsers, one don't miss them until they would be quite handy in some situations ...

To be true, i would like them, but i would like a second pane more for the beginning :D

tsujan:
> To be true, i would like them, but i would like a second pane more for the beginning

To me, tab DND was easy to implement because I'd done it in [FeatherPad](https://github.com/tsujan/FeatherPad) ;) I just wonder why none of the file managers that I've seen has it.

I never used a second pane in Dolphin or Nemo, so I don't understand why people like it. From what you said, I guess it's more usable that tab DND. Is it? If so, I'll start to think on it.

agaida:
to be true - i don't know - if i compare/sync directories i do it with beyond compare - so i really have not so much use cases for a second pane. but a lot of people seems to use dual pane in dolphin, don't know the reasons

tsujan:
>  i really have not so much use cases for a second pane but a lot of people seems to use dual pane in dolphin, don't know the reasons.

Then it may be just a matter of habit.

Anyway, since its logic is simple, I'll make a PR for tab DND after some tests and after that, I'll think on a second pane, although it can't be as easy to implement as DND is.

agaida:
ok - in that case i would like to suggest that we leave out DND for this release and push the button again when it is ready - to be honest, i would prefer this. The main reason for this point release is/should be introducing the build tools - and we should not put so much in it.

And i have no objections to push the big release button for the related packages mid of January again. What do you think?

tsujan:
> ok - in that case i would like to suggest that we leave out DND for this release and push the button again when it is ready 

@agaida, I don't have any objection at all. Frankly speaking, I don't see releases but the continuity of the work ;) Being "release-agnostic", it makes no difference to me which feature is in which release.

agaida:
less complex and overloaded things reduce the chance of regressions - or make it easy to find them - i'm not that release agnostic, but hey - i run sid, that will never be released and the latest git :D

[...]

PR #434, opened 2016-12-22
Tab detachment and DND

tsujan:
This is about https://github.com/lxde/pcmanfm-qt/issues/428. IMO, users will find various applications for tab DND, as I did.

Tabs can be detached and also moved to another window by drag-and-drop. The page of the dropped tab is moved from the first window to the second one as a new tab without being deleted/recreated.

### DIFF
--- a/pcmanfm/mainwindow.cpp
+++ b/pcmanfm/mainwindow.cpp
@@ -31,6 +31,7 @@
 #include <QKeySequence>
 #include <QDir>
 #include <QSettings>
+#include <QMimeData>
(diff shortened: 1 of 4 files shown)


agaida:
@tsujan Do you want the DND things in the 0.11 series or should we shift it in 0.12?

tsujan:
@agaida 
Although I'm using it and it has no problem, it isn't a simple patch and should be reviewed by @PCMan. So, I don't think it'll be ready for 0.11.

tsujan:

PR #441, opened 2017-01-11
Add a workaround for the Qt5 bug which causes broken wallpaper background

PCMan:
This is to workaround Qt bug 54384 which affects Qt >= 5.6
https://bugreports.qt.io/browse/QTBUG-54384
Setting a QPixmap larger then the screen resolution to the background of a list view won't work.
So we did a hack here: Disable the automatic background painting.
Then paint the background of the list view ourselves by hook into its paint event handling method with a event filter.
Since this is a "hack", it's better to do more test to ensure it does not cause side effects. I really need helps on this. @tsujan please help test this if you have time since you're good at finding bugs. :-)

### DIFF
--- a/pcmanfm/desktopwindow.cpp
+++ b/pcmanfm/desktopwindow.cpp
@@ -26,7 +26,7 @@
 #include <QFile>
 #include <QPixmap>
 #include <QPalette>
-#include <QBrush>
+
 #include <QLayout>
 #include <QDebug>
 #include <QTimer>
@@ -36,6 +36,7 @@
 #include <QShortcut>
 #include <QDropEvent>
 #include <QMimeData>
+#include <QPaintEvent>
 
 #include "./application.h"
 #include "mainwindow.h"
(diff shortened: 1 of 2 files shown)


PCMan:
@agaida Your review is also wanted. The change is small, but IMHO it deserves a new minor release. Without the workaround, the desktop wallpaper might be broken in multi-head setup and when screen resolution is high, which I consider a serious UX problem. Thanks a lot.

luis-pereira:
Can't test it. It really deserves a minor release as many users reported it.

agaida:
i can review, but not test it - but if it works we should release it asap - and we can do this thanks to our latest changes without problems

agaida:
@luis-pereira at least we can test and make sure that no regressions are introduced - and i would consider this also as very important. other question - is it possible to simulate  a big hires  or dual screen in kvm or virtualbox?

agaida:
Build, install and run fine - tested on Virtualbox 2560x1600. Background is fine, no regressions in other installations visible.

PCMan:
@agaida The PR is merged and the master branch now contains the fix.
Would you mind help make a new release? Thanks a lot.

agaida:
@PCMan - yes - we can release it today if you want to.

PCMan:
@agida yes, please ship it!  :)

agaida:
@PCMan  done :)

PR #442, opened 2017-01-17
Use const iterators

luis-pereira:
Avoid assignment of a non-const pointer to a const pointer and the
consequent container detach.


luis-pereira:
@PCMan BTW, What's the reasonig behind ?
``` cpp
    QList<QAction*>::const_iterator last_it = actions.end() - 2;
```

tsujan:
>BTW, What's the reasonig behind ?

@luis-pereira, there are 2 items after the separator on the bookmarks menu. `loadBookmarksMenu()` will add the separator too.

luis-pereira:
@tsujan Thanks for the explanation. Just pushed https://github.com/lxde/pcmanfm-qt/pull/444 for caution.

tsujan:
GTM.

tsujan:
@luis-pereira, please rebase!

PR #444, opened 2017-01-27
Checks bookmarks iterators validity

luis-pereira:
After doing iterator arithmetic we must check if the iterator is still
valid.


tsujan:
Frankly, I think it's too cautious, but GTM.

luis-pereira:
It might be, but, better safe than sorry, IMO.

PR #446, opened 2017-01-28
Fixes tiling of wallpapers after https://github.com/lxde/pcmanfm-qt/c…

tsujan:
Fixes tiling of wallpapers after https://github.com/lxde/pcmanfm-qt/commit/de24628202b2789d8c11954669a5b86f35b2c346.

### DIFF
--- a/pcmanfm/desktopwindow.cpp
+++ b/pcmanfm/desktopwindow.cpp
@@ -534,7 +534,10 @@ void DesktopWindow::paintBackground(QPaintEvent *event) {
     painter.fillRect(event->rect(), QBrush(bgColor_));
   }
   else {
-    painter.drawPixmap(event->rect(), wallpaperPixmap_, event->rect());
+    if(wallpaperMode_ == WallpaperTile)
+      painter.drawTiledPixmap(event->rect(), wallpaperPixmap_);
+    else
+      painter.drawPixmap(event->rect(), wallpaperPixmap_, event->rect());
   }
 }
 


pcmanfm/desktopwindow.cpp
```diff
@@ -534,7 +534,10 @@ void DesktopWindow::paintBackground(QPaintEvent *event) {
     painter.fillRect(event->rect(), QBrush(bgColor_));
   }
   else {
-    painter.drawPixmap(event->rect(), wallpaperPixmap_, event->rect());
+    if(wallpaperMode_ == WallpaperTile)
```
palinek: Shouldn't be handled all cases here?
```c++
  enum WallpaperMode {
    WallpaperNone,
    WallpaperStretch,
    WallpaperFit,
    WallpaperCenter,
    WallpaperTile,
    WallpaperZoom
  };
```

tsujan: What do you mean by that? After @PCMan's workaround, tiling should be done by drawTiledPixmap().

tsujan:
@pmattern, if you have dual or hdpi monitor, could you please test this?

pcmanfm/desktopwindow.cpp
palinek: Sorry... it's done [here](https://github.com/lxde/pcmanfm-qt/blob/master/pcmanfm/desktopwindow.cpp#L297)

tsujan: It's OK.

palinek:
I've made a test... and with this patch, the tiling has the same "bad" symptoms as described in the original issue lxde/lxqt#1027. We probably need to render the tilling manually -> with this [commit](https://github.com/lxde/pcmanfm-qt/commit/bedd02ca8141), the tilling is OK for me.

tsujan:
@palinek 
Thanks for the test! Your approach seems good but please read my comment at https://github.com/lxde/pcmanfm-qt/commit/bedd02ca8141#commitcomment-20661937

tsujan:
There may still be a problem and that's HDPI. `QPainter::drawTiledPixmap()` is not equivalent to drawing a pixmap several times.

palinek:
> QPainter::drawTiledPixmap() is not equivalent to drawing a pixmap several times.

From the doc -> http://doc.qt.io/qt-5/qpainter.html#drawTiledPixmap
"Calling drawTiledPixmap() is similar to calling drawPixmap() several times to fill (tile) an area with a pixmap, but is potentially much more efficient depending on the underlying window system."

tsujan:
Yes, I wanted to quote it too. I mean this part: "is potentially *much more efficient* depending on the underlying window system."

Let me first try your solution with small images and compare it to drawTiledPixmap()!

tsujan:
Superseded by https://github.com/lxde/pcmanfm-qt/pull/448

Issue #445, opened 2017-01-27
Path bar switches automatically to "Path buttons" mode after reboot. 


githubtefo:
Hello,
since v0.11.3, when I restart the lxqt- session, the Path bar changes from Location mode to buttons mode.
Is there any way to prevent this?
Thanks in advance!

tsujan:
Check View → Path Bar → Location!

githubtefo:
tsujan, thanks for your answer. I changed to Location mode but every time I restart the session the Path bar returns to Buttons Mode.

tsujan:
I just changed that setting, logged out, logged in again and the new setting was preserved. How do you restart the lxqt-session?

githubtefo:
By rebooting the whole system (Arch/lx-qt).

tsujan:
Confirmed!

This bug isn't in pcmanfm-qt (although a workaround is possible in pcmanfm-qt). The bug may be in lxqt-session. @luis-pereira, @palinek ?

tsujan:
pcmanfm-qt saves its settings on quitting. It seems that rebooting the system dosn't make it quit. I don't see a crash either.

luis-pereira:
@palinek I'm away from my dev machine. I can certainly look at it Monday.

luis-pereira:
I did a logout and a login and was not able to reproduce it.

tsujan:
@luis-pereira, it's about shutdown/reboot, not logout/login.

palinek:
I can confirm, that the lxqt-session isn't properly stopped in case of reboot here.

You can check/confirm it by enabling the debug output for lxqt-session (by setting the environment variable `QT_LOGGING_RULES="lxqt-session.debug=true"` in e.g. `startlxqt`). In case of reboot [these messages](https://github.com/lxde/lxqt-session/blob/master/lxqt-session/src/lxqtmodman.cpp#L357) aren't shown in the output.

Ending the session was fixed here -> lxde/lxqt-session/@302843e86512b38f9809c2119f15e02ffb6c3629. But now it doesn't seem to work any more.

That leads to the question how is the session stopped by its master/parent/display manager/systemd?

palinek:
I'm on lightdm... the parent of lxqt-session is `lightdm --session-child 14 21` in my case.

[...]

PR #447, opened 2017-01-28
Quit gracefully with some signals

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/445 and similar issues.


tsujan:
I close this because pcmanfm-qt already has a SIGTERM handler and the other signals aren't important.

PR #448, opened 2017-01-30
DesktopWindow: Render tilled wallpaper manually

palinek:
Fixes tiling of wallpapers after https://github.com/lxde/pcmanfm-qt/commit/de24628202b2789d8c11954669a5b86f35b2c346.

### DIFF
--- a/pcmanfm/desktopwindow.cpp
+++ b/pcmanfm/desktopwindow.cpp
@@ -300,7 +300,17 @@ void DesktopWindow::updateWallpaper() {
     QImage image;
     if(wallpaperMode_ == WallpaperTile) { // use the original size
       image = QImage(wallpaperFile_);
-      pixmap = QPixmap::fromImage(image);
+      // Note: We can't use the QPainter::drawTiledPixmap(), because it doesn't tile
+      // correctly for background pixmaps bigger than the current screen size. This
+      // may be also the reason for wrong rendering by QWidget::setAutoFillBackground(true)
+      const QSize s = size();
+      pixmap = QPixmap{s};
+      QPainter painter{&pixmap};
+      for (int x = 0; x < s.width(); x += image.width()) {
+          for (int y = 0; y < s.height(); y += image.height()) {
+              painter.drawImage(x, y, image);
+          }
+      }
     }
     else if(wallpaperMode_ == WallpaperStretch) {
       image = loadWallpaperFile(size());


tsujan:
OK. I tested this with a 200x150 image and a 1920x1080 screen and couldn't see any difference in CPU usage. Apparently Qt devs have exaggerated about the efficiency of `drawTiledPixmap()` ;)

IMO, it's better to wait for a week and then merge it in case there's no feedback.

tsujan:
The question arises as to why not doing a similar thing with `WallpaperStretch` too? I mean:
````
    ...
    else if(wallpaperMode_ == WallpaperStretch) {
      const QSize s = size();
      pixmap = QPixmap(s);
      QPainter painter(&pixmap);
      image = loadWallpaperFile(s);
      painter.drawImage(0, 0, image);
    }
    ...
````
In this way, `paintBackground()` would only draw an empty pixmap, which seems better because paintBackground()` is called frequently. I think that should work with dual monitors too but I can't test.

palinek:
> The question arises as to why not doing a similar thing with WallpaperStretch too? I mean:

Why would we do that? Isn't the `QPixmap::fromImage` good for some reason?

> In this way, paintBackground() would only draw an empty pixmap,

Empty pixmap?

tsujan:
Never mind! I was working on two different things at the same time; hence those words, which are strange to me too ;)

foxpup:
I thought I'd jump in with my observations here.  I am running the latest version of lxqt at least as late as the repositories update to. (0.11.1 on ubuntu zesty)  I'm also running three 4K monitors in panoramic wide mode. (11520x2160) Until recently I was unable to span an image that big across the entire panorama, but it now works.  Strangely, repeated tiling of smaller images does not.  I only get the one image in the upper left corner of the leftmost monitor.  This used to work but is now broken. Overall this is progress since I simply need to generate a 11520x2160 jpg file with any desired content to get whatever look I want and I wasn't able to do that before.  That said, I'm sure others would like to see the repeating tile feature working again.  I am willing to try things and report the effects on my system.  It is my main system so I don't want to do things that are too extreme but I'm willing to do minor stuff and report the results.  (also, I'm using an ASUS NVidea G-Force GTX 960 -  370.28 driver - pushing out to three Dell P2815Q 30Hz 4k monitors)

palinek:
> Strangely, repeated tiling of smaller images does not. I only get the one image in the upper left corner of the leftmost monitor. This used to work but is now broken.

Yes, this is the problem, we are trying to solve here.

tsujan:
> It is my main system so I don't want to do things that are too extreme but I'm willing to do minor stuff and report the results

If you can compile from the source, please tell us whether https://github.com/lxde/pcmanfm-qt/tree/tiled_wallpaper2 fixes the tiling issue with your gigantic display and without any extra CPU usage!

palinek:
@tsujan @PCMan how about this?

tsujan:
@palinek 
Oh, we forgot this. Merge it!

PR #450, opened 2017-02-08
Adapt to the changes of the new libfm C++-11 APIs

PCMan:
This is related to the latest libfm-qt changes.
https://github.com/lxde/libfm-qt/pull/63

### DIFF
--- a/pcmanfm/desktopitemdelegate.cpp
+++ b/pcmanfm/desktopitemdelegate.cpp
@@ -64,16 +64,18 @@ void DesktopItemDelegate::paint(QPainter* painter, const QStyleOptionViewItem& o
     else {
         iconMode = QIcon::Disabled;
     }
-    QPoint iconPos(opt.rect.x() + (opt.rect.width() - opt.decorationSize.width()) / 2, opt.rect.y());
-    QPixmap pixmap = opt.icon.pixmap(opt.decorationSize, iconMode);
-    painter->drawPixmap(iconPos, pixmap);
+    QPoint iconPos(opt.rect.x() + (opt.rect.width() - option.decorationSize.width()) / 2, opt.rect.y());
+    QPixmap pixmap = opt.icon.pixmap(option.decorationSize, iconMode);
+    // in case the pixmap is smaller than the requested size
+    QSize margin = ((option.decorationSize - pixmap.size()) / 2).expandedTo(QSize(0, 0));
+    painter->drawPixmap(iconPos + QPoint(margin.width(), margin.height()), pixmap);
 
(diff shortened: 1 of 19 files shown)


pcmanfm/desktopitemdelegate.cpp
```diff
@@ -32,172 +32,178 @@
...
-  // draw some emblems for the item if needed
-  // we only support symlink emblem at the moment
-  Fm::FileInfo file = static_cast<FmFileInfo*>(index.data(Fm::FolderModel::FileInfoRole).value<void*>());
-  if(!file.isNull()) {
-    if(file.isSymlink()) {
-      painter->drawPixmap(iconPos, symlinkIcon_.pixmap(opt.decorationSize / 2, iconMode));
+    QPoint iconPos(opt.rect.x() + (opt.rect.width() - opt.decorationSize.width()) / 2, opt.rect.y());
+    QPixmap pixmap = opt.icon.pixmap(opt.decorationSize, iconMode);
+    painter->drawPixmap(iconPos, pixmap);
+
```
tsujan: All instances of `opt.decorationSize` should be changed to `option.decorationSize` and the above three lines should be changed as in `folderitemdelegate.cpp` because otherwise, the alignment of small icons/thumbnails and their text rectangles would be incorrect on desktop.

I don't why this issue didn't show up before your PR but if you want, I can fix it with a commit.

palinek: > I don't why this issue didn't show up before your PR but if you want, I can fix it with a commit.

IMO, go ahead and push the needed commits.... (should @PCMan not agree with them, they can be amended or purged later)

tsujan: OK.

tsujan:
I mean this:



which allows 4 lines of text and causes artifacts on refreshing desktop. The commit corrects it as this:

agaida:
looks good, uploaded

agaida:
The problem is - if there are to much warnings about nonsense one tend to ignore warnings at all - or even worse - at some point some guru change some compiler settings and declare them bugs.

agaida:
Items can be moved again - build and uploaded to snapshots, thanks

BTW: It slowly becomes a mess - we should merge now.

tsujan:
> BTW: It slowly becomes a mess - we should merge now.

I agree but prefer a merge after @palinek makes https://github.com/lxde/libfm-qt/pull/68 complete. Also there's a conflict with this commit https://github.com/lxde/pcmanfm-qt/commit/8854165753334185a0c51c2f06affc30e5f0a46a

PR #471, opened 2017-04-08
Inline Renaming for Desktop

tsujan:
This depends on and follows https://github.com/lxde/libfm-qt/pull/67. It implements inline renaming for LXQt desktop.

A screenshot:

inline


tsujan:

Issue #477, opened 2017-04-18
No "pointing hand" cursor for desktop items and...


tsujan:
This is a small issue but deserves a fix, especially when the fix is so easy: when the cursor is on a desktop item, it doesn't turn into a "pointing hand"; desktop items don't obey the auto-selection setting either.

PR #478, opened 2017-04-18
Fix cursor on hovering desktop items

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/477.

Desktop items also obey auto-selection setting now.

### DIFF
--- a/pcmanfm/desktopwindow.cpp
+++ b/pcmanfm/desktopwindow.cpp
@@ -948,7 +948,7 @@ bool DesktopWindow::eventFilter(QObject* watched, QEvent* event) {
             break;
         }
     }
-    return false;
+    return Fm::FolderView::eventFilter(watched, event);
 }
 
 void DesktopWindow::childDropEvent(QDropEvent* e) {


pcmanfm/desktopwindow.cpp
```diff
@@ -948,7 +948,7 @@ bool DesktopWindow::eventFilter(QObject* watched, QEvent* event) {
             break;
         }
     }
-    return false;
+    return Fm::FolderView::eventFilter(watched, event);
```
palinek: Are you skipping the direct parent's function intentionally?
(right, there is no override of `View::eventFilter()` now, but it can be added in the future...)

tsujan: Parent's function?! DesktopWindow → View → Fm::FolderView.

palinek: Yes, **direct** parent....and leave the compiler handle overrides.

tsujan: I'm not fond of hairsplitting; I enjoy coding because it's fun. Sometimes I feel that you make no hierarchy of things according to their relative importance in your mind. That's OK but I *do*. Why should I even think about adding a `View::eventFilter()` and override it while I can just change a line to fix a bug?! That's not my style.

palinek: No, you **should not** add new override -> implement `View::eventFilter()`. You need to

> just change a line to fix a bug

in this way
```c++
return View::eventFilter(watched, event);
```
to protect against (a potential) new bug in the future.

> I feel that you make no hierarchy of things according to their relative importance in your mind.

Of course I do. Everyone does. But I always prefer to make a "clean" (for the future) fix instead of  an "it works somehow" fix. With this I'm not saying this patch is one of "it works somehow".

I just tried to point that the solution can be made more robust for the future. That's all.

tsujan: "Protection", "clean", "robust" and fear of future bugs become very subjective concepts when it comes to hairsplitting. We could discuss about our likes and dislikes for hours and reach no conclusion because there is nothing to conclude. In contrast, discussing about objective facts is always fruitful, even if long. You and I have had fruitful discussions -- and I hope we'll have many in future -- but this isn't one of them.

PR #483, opened 2017-04-20
Use const iterators (Fix a FTBFS in debug mode)

luis-pereira:
Avoid assignment of a non-const pointer to a const pointer and the
consequent container detach.

We now enforce it in LXQtCompilerSettings:
https://github.com/lxde/lxqt-build-tools/commit/8679fbacd09fb3afb316cd8d5ed19bc0d0ba6990


palinek:
GTM

agaida:
GTM

Issue #495, opened 2017-04-30
Sporadically crashes when moving files on desktop via pcmanfm-qt instance


agaida:
* steps to reproduce - erm, be patient, don't have a set of actions to trigger this. even more annoying, wasn't able to reproduce this crashes via gdb. 
* how to track otherwise: running systemd-coredump and wait

```
[New LWP 16586]
[New LWP 16587]
[New LWP 29867]
[New LWP 16588]
[New LWP 16592]
[New LWP 28384]
[New LWP 16589]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `pcmanfm-qt --desktop --profile=lxqt'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fc3f4ecd94c in ?? () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
[Current thread is 1 (Thread 0x7fc3f6020900 (LWP 16586))]
(gdb) 
(gdb) 
(gdb) 
(gdb) bt
#0  0x00007fc3f4ecd94c in ?? () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#1  0x00007fc3f4ecdc53 in ?? () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#2  0x00007fc3f4d2cc52 in QString::fromUtf8_helper(char const*, int) () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#3  0x000055f12f46fa5d in QString::fromUtf8 (size=<optimized out>, str=0x41 <error: Cannot access memory at address 0x41>)
    at /usr/include/x86_64-linux-gnu/qt5/QtCore/qstring.h:543
#4  QString::fromStdString (s=...) at /usr/include/x86_64-linux-gnu/qt5/QtCore/qstring.h:1311
#5  PCManFM::DesktopWindow::onRowsAboutToBeRemoved (this=0x55f12fbd2e50, parent=..., start=<optimized out>, end=<optimized out>)
    at ./pcmanfm/desktopwindow.cpp:430
#6  0x00007fc3f4e9d95e in QMetaObject::activate(QObject*, int, int, void**) () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
[...]

PR #496, opened 2017-04-30
Fix crash on removing desktop items with custom position

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/495.


agaida:
i have to try harder to produce a crash - hm - maybe a short video might help

tsujan:
> i have to try harder to produce a crash

The same backtrace as the last time (https://github.com/lxde/pcmanfm-qt/issues/495#issuecomment-298237145)?

Issue #498, opened 2017-05-01
"Stick to current position" works only for one desktop item at a time


tsujan:
When multiple selected desktop items are right-clicked and "Stick to current position" is checked, only the first item sticks to its position.

The expected behavior is that the checkbox should work for all selected items.

PR #499, opened 2017-05-01
Make custom pos checkbox work for multiple items

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/498.

The state of the "Stick to current position" checkbox is corrected for multiple selections. Also when multiplely selected desktop items are right-clicked and the box is checked, all of them stick to their positions; if the box is unchecked, all of them will be sorted according to the sorting order.

### DIFF
--- a/pcmanfm/desktopwindow.cpp
+++ b/pcmanfm/desktopwindow.cpp
@@ -384,15 +384,15 @@ void DesktopWindow::prepareFileMenu(Fm::FileMenu* menu) {
     menu->insertSeparator(menu->separator2());
     menu->insertAction(menu->separator2(), action);
 
+    bool checked(true);
     auto files = menu->files();
-    // select exactly one item
-    if(files.size() == 1) {
-        auto file = menu->firstFile();
-        if(customItemPos_.find(file->name()) != customItemPos_.cend()) {
-            // the file item has a custom position
-            action->setChecked(true);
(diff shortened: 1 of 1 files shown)


tsujan:
@agaida 
I think the purpose of this PR is misunderstood. Let me explain it:

There are some items that have custom positions. When you right click one of them, "Stick to Current Position" is checked. Other items don't have custom positions but are placed according to the sorting method. For them, the checkbox in question isn't checked, of course.

Previous to this PR, when you right-clicked *multiple items with custom positions*, that checkbox wasn't checked -- it worked only when a single item. That was definitely a bug, which is fixed by this PR.

tsujan:
@agaida 
You might ask, "why do some items have custom/fixed positions?" The answer is because the user has given custom positions to them by either (1) dragging and dropping them to some other places, or (2) right clicking on them and checking "Stick to Current Position".

tsujan:
Yet another explanation regarding your screen-cast: When you copy items to desktop, they don't have custom positions yet, so they're positioned according to the sorting order. The case of drag-n-drop was different, so I made https://github.com/lxde/pcmanfm-qt/pull/504 for it.

BTW, I have a fix for those ugly item jumps that can be seen in your screen-cast too. Waiting for these PRs to be merged because I hate unneeded rebasing ;)

Issue #500, opened 2017-05-01
Positions when moving files to desktop or renaming files on desktop


agaida:
This is a follow up for #495 and lxde/lxqt#1030.

Imho the current handling of moving or renaming files is false in the sense that it don't match users expectations. When i move a file (or a bunch of) to the desktop per drag and drop i expect that it lands on the mouse position. When i rename a file it should stay in the position where it was before - when the desktop is set to use custom positions instead of auto-sorting.

Sample of that behaviour (Win 10):  https://youtu.be/0H9p4XqO6dA

@tsujan: feel free to change title or wording - i don't know how to describe it better.

tsujan:
@agaida The wording is good enough.

As for renaming, it's possible only when files are renamed on deskop itself -- and I think that's sufficient.

I also think that this should be added as an option in Desktop Preferences.

The needed code seems to be more or less straightforward.

tsujan:
This is my plan for now (you may disagree with it, of course):

(1) The auto-sorting will remain intact so that when a file is pasted or downloaded to desktop, it will appear according to the chosen sorting.

(2) When files are put on desktop by dragging and dropping, they will (almost) land on the mouse position (but optionally).

(3) When a file is renamed from inside desktop, its position won't change (but optionally).

fulalas:
@tsujan, sound great. But my dream is an option to disable auto-sorting, like Windows, for example. In other words, it'll add icons like a stack (from top left to bottom right).

tsujan:
>  it would add icons like a stack (from top left to bottom right).

That means all items are listed, instead of them being sorted. it can be implemented but please let me first implement the above-mentioned properties in an experimental PR. Then it could be tested and more properties could be added step by step. I think the current desktop structure has still potential.

tsujan:
On second thought, I should say that auto-sorting cannot be removed *logically*. Suppose that we put items on desktop *only* based on a growing list (which means stacking). Then, yes, we could add new items to the end of the list but where on desktop? Stacking doesn't mean that there are no gaps between the real items on desktop -- and the gaps can be large. Should we put the new item immediately after the last one? The answer is no because the last item may be positioned at the right of dekstop.

So, I stick with my previous plan for now.

fulalas:
> Should we put the new item immediately after the last one?

No. You put on the first found gap -- unless user created/moved/copied an icon to a specific place, of course.

OK, no need to hurry. :)

tsujan:
[...]

PR #504, opened 2017-05-07
Place dropped item at drop position

tsujan:
Fixes (a part of) https://github.com/lxde/pcmanfm-qt/issues/500.

For a more natural "desk-top" experience, dropped items are positioned successively (top→bottom + left→right), starting with the drop position (rectangle).

IMO, there's no need to make this optional because most users expect such a behavior and if a user doesn't want it, he/she could use the context menu easily (see https://github.com/lxde/pcmanfm-qt/pull/499).


agaida:
Works as expected, very nice, imho GTM

PR #529, opened 2017-07-09
Replace customized DesktopItemDelegate hacks with the standard Fm::FolderItemDelegate provided by libfm-qt.

PCMan:
This PR moves the text shadow painting code from PCManFM::DesktopItemDelegate to Fm::FolderItemDelegate. So pcmanfm-qt can reuse the class directly without adding its own item delegate.
Finally this piece of duplicate code is eliminated.


tsujan:
Merged after https://github.com/lxde/libfm-qt/pull/100

Issue #533, opened 2017-07-12
pcmanfm-qt doesn't remember if hidden files are shown


tsujan:
##### Expected Behavior
When `Ctrl+H` is used, pcmanfm-qt should remember if hidden files are shown or not.

##### Current Behavior
When hidden files are shown, if another folder is loaded or the window is closed and reopened, hidden files will be hidden again.

##### Steps to Reproduce (for bugs)

1. Use `Ctrl+H` to show hidden files
2. Change the directory or close and open the window
3. Hidden files will be hidden again

##### System Information

* Everything is the latest version

Please note:

I don't have this issue with my patched pcmanfm-qt and libfm-qt. I saw it once I started to use the default, unpatched sources. So, a confirmation is needed. If confirmed, I should see which one of my unpublished patches fixes the issue.

tsujan:
@agaida Could you confirm this?

tsujan:
Found it, after a long walk under rain :) The story is that I use the original per-folder settings patch I'd proposed in an old PR (which was replaced by @PCMan's method); in that patch, I'd fixed this issue.

Will wait for confirmation.

MggMuggins:
I actually like it hiding the hidden files again. Could that be a configuration option?

tsujan:
When the user presses `Ctrl+H`, he/she wants to see hidden files. Making them hidden again when the directory is changed is an annoying bug not a feature. `Ctrl+H` could be pressed again if hidden files aren't needed to be shown anymore. Sane file managers like Dolphin behave so (Dolphin can remember it with a single setting but that's another matter).

tsujan:
A fix for this and related bugs is at https://github.com/lxde/pcmanfm-qt/pull/535.

PR #535, opened 2017-07-14
Fix showing/hiding hidden files

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/533.

The bugs that are fixed by this patch:

(1) The status message wasn't changed when hidden files were shown/hidden by context menu (in contrast to `Ctrl+H`).
(2) Showing hidden files wasn't remembered globally.
(3) Sometimes, the setting for showing hidden files wasn't saved for customized folders either.

Current behavior:

(1) If hidden files are shown/hidden inside a folder *without* customization, the setting will be remembered globally for all folders *without* customization.
(2) If hidden files are shown/hidden inside a folder *with* customization, the setting will be remembered only for that folder and not globally.

The behavior implemented by this patch is usually expected for hidden files, as is the case with well-known file managers.

IMHO, this behavior is better to be generalized to other view settings too and it's unfortunate that pcmanfm-qt has no way to inform the user that a folder is customized other than by toggling checkboxes in menus -- but that's another matter to discuss.

### DIFF
--- a/pcmanfm/mainwindow.cpp
+++ b/pcmanfm/mainwindow.cpp
@@ -489,6 +489,9 @@ void MainWindow::on_actionFolderProperties_triggered() {
 void MainWindow::on_actionShowHidden_triggered(bool checked) {
     currentPage()->setShowHidden(checked);
     ui.sidePane->setShowHidden(checked);
+    if(!currentPage()->hasCustomizedView()) {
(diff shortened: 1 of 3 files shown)


agaida:
GTM

tsujan:
So, there's no disagreement on this? I mean, IMO, this is the rational behavior. Does anyone think I'm irrational?

agaida:
Erm, now that you ask .. :P

Issue #534, opened 2017-07-13
Counterintuitive scrolling and selection on reloading


tsujan:
##### Expected Behavior
When a folder is reloaded, the previous child folder shouldn't be selected or scrolled to.

##### Current Behavior
On reloading, pcmanfm-qt selects and scrolls to the child folder from which the directory was changed.

##### Possible Solution
Will make a PR.

##### Steps to Reproduce

1. Go from a folder to its parent folder (by going up).
2. Deselect the child folder (which was correctly selected).
3. Reload the current (parent) folder. 
4. The child folder will be selected again and scrolled to, which is totally counterintuitive.

##### System Information

* Everything is up-to-date

PR #536, opened 2017-07-14
Don't select or scroll to child folder on reloading

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/534.

In addition, this patch restores the scrollbar position on reloading.


agaida:
GTM

PR #544, opened 2017-07-20
Inline renaming with F2

tsujan:
This depends on and follows https://github.com/lxde/libfm-qt/pull/98. It implements inline renaming with `F2` for single files.


agaida:
tested and work like charm

tsujan:
Thanks for the fast test! I should have made this PR sooner.

PR #546, opened 2017-07-22
Wallpaper Slide Show

tsujan:
Fixes https://github.com/lxde/lxqt/issues/608.

The new options are added to Desktop Preferences in a new tab. The slide-show interval is between 5min and 24h55min. An option is also added for randomizing the slide show.

Unusual cases -- like when a wallpaper is removed, the folder doesn't exist, etc. -- are dealt with appropriately.

P.S.
(1) I also fixed what @pmattern saw as a minor issue in an old discussion; now, pressing OK in the the Desktop Preferences doesn't change the current slide unless needed (KDE has also fixed this in its slide show).
(2) This patch is not new -- I use it for half a year.

### DIFF
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -679,11 +679,11 @@ void Application::updateFromSettings() {
     }
 }
 
-void Application::updateDesktopsFromSettings() {
+void Application::updateDesktopsFromSettings(bool changeSlide) {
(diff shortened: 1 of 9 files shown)


tsujan:
A picture is always worth a thousand words:

preferences

tsujan:
Please test and merge this -- I've used it for a long time.

Issue #563, opened 2017-09-17
Default search settings of pcmanfm-qt aren't practical


tsujan:
As the title says.

##### Possible Solution
There can be 2 solutions:

(1) A better set of settings; or
(2) Saving settings on closing the search dialog.

The solution (1) has this potential problem that there might be no consensus about the best settings. So, I might make a PR for saving the search settings.

stefonarch:
No idea why in file-search.ui the settings do not work:
```
<item row="1" column="0" colspan="2">
           <widget class="QCheckBox" name="checkBox">
            <property name="text">
             <string>Case insensitive</string>
            </property><item row="1" column="0" colspan="2">
           <widget class="QCheckBox" name="checkBox">
            <property name="text">
             <string>Case insensitive</string>
            </property>
            <property name="checked">
             <bool>true</bool>
            </property>
           </widget>
          </item>
            <property name="checked">
             <bool>true</bool>
            </property>
           </widget>
          </item>
```
![schermata-18-17-57-50](https://user-images.githubusercontent.com/10681413/30551603-49157164-9c9b-11e7-8484-2b5fc1004459.png)

tsujan:
> No idea why in file-search.ui the settings do not work:

I didn't understand what you meant. Did you try [this](https://github.com/lxde/libfm-qt/pull/116) and [this](https://github.com/lxde/pcmanfm-qt/pull/564) (in this same order)?

stefonarch:
I tried just to set "checked" in file-search.ui and it did not work, only then I noticed your commit.  
Now I  compiled both, testing, seems working fine recording the previous settings, I think that resolves the issue, not everyone uses the same settings, I had always to check "subfolders" and "case insensitive" 2 clicks to much for a lazy man like me  ;)

tsujan:
@stefonarch Thanks for testing!

> I had always to check "subfolders" and "case insensitive" 2 clicks to much for a lazy man like me ;)

I may be lazier. A good app design should care for all lazy users ;)

Joking aside, when the search dialog is used a lot, checking those boxes becomes a pain in the a...

fulalas:
Since LXDE by default checks both 'subfolders' and 'case insensitive', I vote for keeping this and also giving the possibility to save what user sets.

palinek:
> Since LXDE (and Windows) by default checks both 'subfolders' and 'case insensitive', I vote for keeping this and also giving the possibility to save what user sets.

[...]

PR #564, opened 2017-09-18
Save and restore search settings

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/563.

Also removed a redundant file that was left since libfm-qt was separated from pcmanfm-qt.

This PR follows and depends on https://github.com/lxde/libfm-qt/pull/116.


agaida:
GTM

PR #568, opened 2017-09-22
Disable menubar actions that cannot be used

tsujan:
Follows https://github.com/lxde/libfm-qt/pull/121, disabling menubar Edit actions when they cannot be performed.

Also a typo is corrected.


agaida:
GTM

Issue #569, opened 2017-09-26
status bar issues


tsujan:
The status bar has 2 problems:

(1) When a folder is selected, if dir is changed into it, the status message won't be updated.
(2) As @PCMan has commented in the code, the status message may change too frequently and that's 
inefficient.

##### Expected Behavior
Status message should be updated appropriately and not, for example, 100 times in a second.

##### Possible Solution
Will fix both issues in one PR. (1) is about a small mistake (a missing line) but (2) needs a custom statusbar.

##### Steps to Reproduce (for bugs)

1. Select a folder and see the status message;
2. Press enter to change dir into that folder while looking at the status message;
3. The message won't be updated.

##### System Information

* As always, everything is the last git.

PR #570, opened 2017-09-26
Update status message appropriately

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/569 and https://github.com/lxde/pcmanfm-qt/issues/571

This patch fixes:

(1) A bug, because of which, the status text wasn't updated when the folder was selected before changing dir into it;
(2) Too frequent and redundant text updates; and
(3) A Qt "feature", because of which, the status text disappeared on focusing menubar items.

### DIFF
--- a/pcmanfm/CMakeLists.txt
+++ b/pcmanfm/CMakeLists.txt
@@ -4,6 +4,7 @@ set(pcmanfm_SRCS
     mainwindow.cpp
     tabpage.cpp
     tabbar.cpp
+    statusbar.cpp
     view.cpp
     launcher.cpp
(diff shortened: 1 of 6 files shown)


tsujan:
Fixed a mistake of mine and rebased. It should be OK now.

tsujan:
Too long texts and newline in file names are taken into account and handled in the last commit. Also, the text is elided when there isn't enough space (in case a file with very long name is selected, for example).

No other scenario comes to my mind. I'm using this but more tests would be really appreciated.

agaida:
GTM

Issue #562, opened 2017-09-17
Replace 20×20 with 22×22 in display settings of pcmanfm-qt


tsujan:
Most icon themes include separate 22px icons but no 20px or 12px ones. Separate icons are important with such small sizes because if they're missing, tiny icons might seem blurry (there are well-known methods for making small icons sharp). Moreover, 20 and 12 aren't standard icon sizes anywhere in Linux, as far as I know.

So, I suggest replacing 20 with 22 and removing 12 in pcmafm-qt display settings. Please tell me about your opinion!

palinek:
agreed

tsujan:
This isn't as trivial as it may seem. A unified approach should be taken in `Settings` and `PreferencesDialog`.

PR #572, opened 2017-09-27
Fix icon sizes

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/562

Non-standard small icon sizes are removed because tiny icons might be blurry with them. Non-standard sizes for relatively big icons are kept though.

Also icon sizes are corrected if manually put into config file, so that the sizes used by Preferences dialog are always respected.

### DIFF
--- a/pcmanfm/settings.cpp
+++ b/pcmanfm/settings.cpp
@@ -266,10 +266,10 @@ bool Settings::loadFile(QString filePath) {
     shadowHidden_ = settings.value("ShadowHidden", false).toBool();
 
     // override config in libfm's FmConfig
-    bigIconSize_ = settings.value("BigIconSize", 48).toInt();
-    smallIconSize_ = settings.value("SmallIconSize", 24).toInt();
-    sidePaneIconSize_ = settings.value("SidePaneIconSize", 24).toInt();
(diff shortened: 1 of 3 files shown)


pcmanfm/settings.cpp
```diff
@@ -439,6 +439,33 @@ bool Settings::saveFile(QString filePath) {
     return true;
 }
 
+QList<int> Settings::iconSizes(IconType type) const {
+    QList<int> sizes;
+    switch(type) {
+    case Big:
+        sizes << 96 << 72 << 64 << 48 << 32;
```
palinek: How about using `constexpr`'s (and return the `const QList<int> &` here?

tsujan: No objection. Would you please add a commit?

palinek: OK... the `QList` can't be `constexpr`, but also the `static const` should perform better... pushed the commit.

tsujan: Yes, I saw it. It's good. Thanks!

tsujan:
@agaida Please consider this for the next release.

agaida:
GTM

PR #576, opened 2017-10-18
Set Version

agaida:
Bump tools version to 0.4.0


tsujan:
GTM, after https://github.com/lxde/libfm-qt/pull/122.

PR #586, opened 2017-10-29
Tab DND

tsujan:
Closes https://github.com/lxde/pcmanfm-qt/issues/428.

Tabs can be detached or moved to another window by drag-and-drop. Also windows can be unified with tab DND.

### DIFF
--- a/pcmanfm/mainwindow.cpp
+++ b/pcmanfm/mainwindow.cpp
@@ -1291,12 +1291,7 @@ void MainWindow::dropTab() {
     // its page to a new tab in the second window
     TabPage* dropPage = lastActive_->currentPage();
     if(dropPage) {
-        disconnect(dropPage, &TabPage::titleChanged, lastActive_, &MainWindow::onTabPageTitleChanged);
-        disconnect(dropPage, &TabPage::statusChanged, lastActive_, &MainWindow::onTabPageStatusChanged);
-        disconnect(dropPage, &TabPage::openDirRequested, lastActive_, &MainWindow::onTabPageOpenDirRequested);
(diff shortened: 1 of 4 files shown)


tsujan:
A screencast: https://www.dropbox.com/s/p7c214hln69p8zz/tab_DND.mkv

pcmanfm/mainwindow.cpp
```diff
@@ -1256,6 +1267,80 @@ void MainWindow::focusPathEntry() {
...
+void MainWindow::detachTab() {
+    if (ui.stackedWidget->count() == 1) { // don't detach a single tab
+        ui.tabBar->finishMouseMoveEvent();
+        return;
+    }
+
+    // close the tab and move its page to a new window
+    TabPage* dropPage = currentPage();
+    if(dropPage) {
+        disconnect(dropPage, &TabPage::titleChanged, this, &MainWindow::onTabPageTitleChanged);
```
palinek: Are you disconnecting all signals here? If so, you can use on of the `disconnect`'s overload to do it with one call.

tsujan: Just for the sake of caution: one-by-one disconnecting means we exactly know what's happening.

However, feel free to do it at once by adding a commit for both cases here.

tsujan: @palinek  Never mind! On second thought, that "caution" seems too much here. Will change it later, today.

Long story short, I had to disconnect signals in FeatherPad one by one and did the same thing here without it being really needed

tsujan:
All done.

agaida:
Works fine. GTM

Issue #587, opened 2017-10-30
Change folder view by icon


Ygarr:
Change folder view by icon for fast access like nautilus or nemo fm as most frequent action

##### Expected Behavior
Fast click by mouse on prefered folder view - thumbnail, compact, detailed, icon

##### Current Behavior
Currently need to manually to go to View> View > <select one of 4 views>

##### Possible Solution
Visually clear guessable buttons on pane. Hotkeys, shortcuts if  there was such a need previously

 

##### Context
 It is very need for convenient user experience

##### System Information
any

![2017-10-30-222500_283x587_scrot](https://user-images.githubusercontent.com/14309871/32193856-3c9361b6-bdc1-11e7-92de-04dfd71adb7f.png)

tsujan:
I agree.

tsujan:
Is anyone against this for any reason?

agaida:
no

MggMuggins:
If someone is there is always the option to add a config option to disable it in the UI.

tsujan:
I think a config for this would be an overkill. I'm going to implement it without config, although an option could be added later (but for what?)... Since there can be no shortcut for views -- all good shortcuts are already in use (see https://github.com/lxde/pcmanfm-qt/issues/558) --  this should have been done sooner, IMO.

tsujan:
I chose Dolphin's style and generic `QStyle` icons (exactly as in the LXQt file dialog). The look and grouping of buttons depends on the active style. For example, with a Kvantum theme:

![kvbeige](https://user-images.githubusercontent.com/981076/32354819-d2118278-c040-11e7-891e-088a4ee9529d.png)

And with Fusion:

![fusion](https://user-images.githubusercontent.com/981076/32354824-d859f4c6-c040-11e7-88ea-33946e75d8c9.png)

Will make a PR soon.

tsujan:
BTW, if Fusion is used as the style, the icons for the icon and thumbnail views will be chosen from the hard-coded icons of Qt itself, both in the file dialog and here. This is a Fusion issue and isn't related to LXQt. I guess Fusion devs will never recognize it as a bug.

Issue #558, opened 2017-08-05
Adding shortcuts to View menu


artmg:
Hi,

I was thinking of modifying **pcmanfm-qt/pcmanfm/main-win.ui** to add shortcuts in to make it work the same as original PCManFM

![pmmanfm-view-menu-shortcuts](https://user-images.githubusercontent.com/6281320/28994371-8548fefc-79c4-11e7-9f33-0e2fcc0cd009.png)

Can anyone let me know if Ctrl-1, Ctrl-2, Ctrl-3 and Ctrl-4 would be _unsuitable_ choices, before I submit a PR. Thanks

artmg:
Doh! Of course, Ctrl+digit switches to the _nth_ tab open, just like in a browser window.

So is there any blueprint suggesting standards for implementing keyboard shortcuts in LxQt, to help guide the choice of which key combination to use?

tsujan:
There's still a problem: `Alt+Right`, `Alt+Up`, ... are used for going forward, up,.... ;)

tsujan:
Closed by https://github.com/lxde/pcmanfm-qt/commit/4a96c1c87498e517a5cda9cd37bbbc40151d9b10 because the view buttons are on the toolbar now.

PR #589, opened 2017-11-04
View tool-buttons

tsujan:
Closes https://github.com/lxde/pcmanfm-qt/issues/587 and https://github.com/lxde/pcmanfm-qt/issues/558

EDIT: Preserved the original action names because the toolbar style is icon-only.


agaida:
GTM

PR #593, opened 2017-11-06
Add Spanish desktop entries

jairlopez:


pcmanfm/translations/pcmanfm-qt_es.desktop
```diff
@@ -0,0 +1,4 @@
+#Translations
+Name[es]=Gestor de archivos PCManFM
```
palinek: PCManFM-qt ?

jairlopez: I thought the same, but I didn’t know whether it was appropriate because this same issue was present in [`pcmanfm/pcmanfm-qt.desktop.in`](https://github.com/lxde/pcmanfm-qt/blob/8ba67fdfc8d848e1882f992ee37eb8d795e857dd/pcmanfm/pcmanfm-qt.desktop.in#L3) since 2014. Well, here you go then ...

palinek:
@lxde/lxqt IMO it is correct to use the PCManFM-qt... why wasn't it used in desktop files?

agaida:
@palinek - overlooked? and only nitpicking - shouldn't it be better PCManFM-Qt ?

palinek:
> @palinek - overlooked? 

No, I asked for explanation from LXQt team (I may have overlooked something)... but didn't receive any :)

tsujan:
"PCManFM-Qt" is good.

jairlopez:
Ok, so what should I do?
1. Amend last commit.
2. Add a new one with `PCManFM-Qt` included.
3. Do something else.

palinek:
@jairlopez Please, amend the  47c657a

PR #598, opened 2017-11-12
Really cancel multiple renaming on cancelling

tsujan:
This follows and depends on https://github.com/lxde/libfm-qt/pull/130. It cancels shortcut/menubar renaming of multiple files when Cancel/Esc is pressed.


agaida:
GTM, works fine

Issue #599, opened 2017-11-12
Bulk renaming


tsujan:
An option for bulk renaming (in contrast to the ordinary renaming of multiple files) is missing from pcmanfm-qt.

##### Expected Behavior
Bulk renaming would choose a base name and rename the selected files by adding ascending numbers to the base name.

tsujan:
This is what I mean:

![1](https://user-images.githubusercontent.com/981076/32701994-f181116c-c7f4-11e7-870c-d0230ec50e69.png)
![2](https://user-images.githubusercontent.com/981076/32701998-f9bc619c-c7f4-11e7-969a-5dee34999eac.png)
![3](https://user-images.githubusercontent.com/981076/32702001-fef1fd48-c7f4-11e7-86d7-fd32050fdb7b.png)

stefonarch:
Was about to open an issue but it can go here.

> Bulk renaming would choose a base name and rename the selected files by adding ascending numbers to the base name.

Actually it   removes  the extension or you have to  re-insert the extension manually:
<newname>#<extension>

So this could be improved to be more user friendly.

agaida:
@stefonarch - please file a new bug - nobody cares about bugs that are closed one and a half year ago

stefonarch:
ehm some months ;) 
Will do

PR #600, opened 2017-11-15
Basic bulk rename

tsujan:
Closes https://github.com/lxde/pcmanfm-qt/issues/599.

Follows and depends on https://github.com/lxde/libfm-qt/pull/136.

"#" is used for inserting the ascending numbers at any positions(s) of the base name but if omitted, the numbers will still be added appropriately.

The code could be extended later if needed.

### DIFF
--- a/pcmanfm/bulkrename.cpp
+++ b/pcmanfm/bulkrename.cpp
@@ -1,0 +1,93 @@
+/*
+    Copyright (C) 2017 Pedram Pourang (Tsu Jan) <tsujan2000@gmail.com>
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+#include "bulkrename.h"
+#include <QTimer>
+#include <QPushButton>
+#include <QMessageBox>
+
+#include <libfm-qt/utilities.h>
+
+namespace PCManFM {
+
+BulkRenameDialog::BulkRenameDialog(QWidget* parent, Qt::WindowFlags flags) :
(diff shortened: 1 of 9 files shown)


pcmanfm/bulkrename.cpp
```diff
@@ -0,0 +1,93 @@
...
+            end = baseName.size();
+        }
+        baseName.insert(end, QLatin1Char('#'));
+    }
+    uint failed = 0;
+    for(auto& file: files) {
+        auto fileName = QString::fromStdString(file->name());
+        QString newName = baseName;
+        newName.replace(QLatin1Char('#'), QString::number(start));
+        if (newName == fileName || !Fm::changeFileName(file->path(), newName, nullptr, false)) {
```
palinek: `Fm::changeFileName()` is blocking operation, so calling it in main thread can make the whole application "hang" (you also can get unresponsive desktop while long running renaming). This normally will not be a problem for (few) local files, but can be problematic with remote...

tsujan: Thanks for mentioning that! I'll investigate it tomorrow -- today, I don't have time. However, for local files, I tested with > 1000 files and the renaming was instant; that was the reason I didn't include a progressbar in this.

tsujan: In the meanwhile, feel free to add any commit.

tsujan:
Added a progress dialog, which can be used for aborting bulk renaming (single renaming can't be canceled in libfm-qt yet).

pcmanfm/bulkrename.cpp
```diff
@@ -68,17 +69,26 @@ BulkRenamer::BulkRenamer(const Fm::FileInfoList& files, QWidget* parent) {
...
-        newName.replace(QLatin1Char('#'), QString::number(start));
+        newName.replace(QLatin1Char('#'), QString::number(start + i));
         if (newName == fileName || !Fm::changeFileName(file->path(), newName, nullptr, false)) {
             ++failed;
         }
-        ++start;
+        ++i;
     }
-    if(failed == files.size()) {
+    progress.setValue(i);
```
palinek: Shouldn't this be in the above loop?

tsujan: No it shouldn't.

MggMuggins: Line 76 has the same call.

tsujan: > Line 76 has the same call.

Not the same. Line 90 is after the loop, when `i` is `files.size()`, which mean 100%.

MggMuggins: I thought he was asking why a call that should be made every time the loop executes is being made at the end. It makes sense why you did it.

agaida:
GTM

PR #601, opened 2017-11-21
Wait for events to be processed before chdir

tsujan:
Fixes https://github.com/lxde/lxqt/issues/1410

In all cases of changing directory, a `QTimer::singleShot()` is used to prevent crashes when the action is triggered too fast.

### DIFF
--- a/pcmanfm/mainwindow.cpp
+++ b/pcmanfm/mainwindow.cpp
@@ -287,12 +287,14 @@ MainWindow::~MainWindow() {
 }
 
 void MainWindow::chdir(Fm::FilePath path) {
-    TabPage* page = currentPage();
-    if(page) {
-        ui.filterBar->clear();
-        page->chdir(path, true);
-        updateUIForCurrentPage();
-    }
+    QTimer::singleShot(0, [this, path] {
+        TabPage* page = currentPage();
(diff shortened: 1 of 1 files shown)


tsujan:
BTW, `singleShot()` is needed only in `MainWindow::chdir()` but, with it, the other functions could be safely used in ways other than signal-slot calls.

pcmanfm/mainwindow.cpp
```diff
@@ -287,12 +287,15 @@ MainWindow::~MainWindow() {
...
 
 void MainWindow::chdir(Fm::FilePath path) {
-    TabPage* page = currentPage();
-    if(page) {
-        ui.filterBar->clear();
-        page->chdir(path, true);
-        updateUIForCurrentPage();
-    }
+    // wait until queued events are processed
+    QTimer::singleShot(0, [this, path] {
```
palinek: IMO the `void QTimer::singleShot(int msec, const QObject *context, Functor functor)` overload should be used for the sake of correctness (doc: _If **context** is destroyed before the interval occurs, the method will not be called._)

tsujan: Good point.

PR #606, opened 2017-11-26
Fix direction of statusbar message

### LINKED ISSUE #379: Scrambled statusbar message for files with RTL names

SafaAlfulaij:
We are displaying the filename at the first in the statusbar, which means that if it's RTL and the locale is LTR, the whole text will get messed up, resulting in hard-to-read and wrong display of the name and the information.

Fixes #379

### DIFF
--- a/pcmanfm/tabpage.cpp
+++ b/pcmanfm/tabpage.cpp
@@ -466,8 +466,8 @@ void TabPage::reload() {
 // 202b RIGHT-TO-LEFT EMBEDDING
 // 202c POP DIRECTIONAL FORMATTING
 QString TabPage::encloseWithBidiMarks(const QString& text) {
-    QChar bidiMark = text.isRightToLeft()? QChar(0x200f) : QChar(0x200e);
-    QChar embedBidiMark = text.isRightToLeft()? QChar(0x202b) : QChar(0x202a);
+    QChar bidiMark = text.isRightToLeft() ? QChar(0x200f) : QChar(0x200e);
+    QChar embedBidiMark = text.isRightToLeft() ? QChar(0x202b) : QChar(0x202a);
     return embedBidiMark+text+bidiMark+QChar(0x202c);
 }
 
@@ -486,14 +486,14 @@ void TabPage::onSelChanged() {
                       .arg(encloseWithBidiMarks(Fm::formatFileSize(fi->size(), fm_config->si_unit))) // FIXME: deprecate fm_config
                       .arg(encloseWithBidiMarks(fi->mimeType()->desc()))
                       .arg(QString::fromUtf8(
-                          (!layoutDirection() == Qt::LeftToRight) ? "\u200f" : "\u200e"));
+                          (layoutDirection() == Qt::RightToLeft) ? "\u200f" : "\u200e"));
             }
             else {
                 msg = QString("%3\"%1\" %3%2")
                       .arg(encloseWithBidiMarks(fi->displayName()))
                       .arg(encloseWithBidiMarks(fi->mimeType()->desc()))
                       .arg(QString::fromUtf8(
-                          (!layoutDirection() == Qt::LeftToRight) ? "\u200f" : "\u200e"));
+                          (layoutDirection() == Qt::RightToLeft) ? "\u200f" : "\u200e"));
             }
             /* FIXME: should we support statusbar plugins as in the gtk+ version? */
         }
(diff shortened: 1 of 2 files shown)


pcmanfm/tabpage.cpp
```diff
@@ -470,15 +475,19 @@ void TabPage::onSelChanged() {
...
+                      .arg(encloseWithBidiMarks(fi->mimeType()->desc()))
+                      .arg(QString::fromUtf8(
+                          (!layoutDirection() == Qt::LeftToRight) ? "\u200f" : "\u200e"));
             }
             else {
-                msg = QString("\"%1\" %2")
-                      .arg(fi->displayName())
-                      .arg(fi->mimeType()->desc());
+                msg = QString("%3\"%1\" %3%2")
+                      .arg(encloseWithBidiMarks(fi->displayName())(
```
tsujan: This should be:
```cpp
.arg(encloseWithBidiMarks(fi->displayName()))
```

pcmanfm/tabpage.cpp
```diff
@@ -470,15 +475,19 @@ void TabPage::onSelChanged() {
...
+                      .arg(encloseWithBidiMarks(fi->mimeType()->desc())
```
tsujan: This should be:
```cpp
.arg(encloseWithBidiMarks(fi->mimeType()->desc()))
```

tsujan:
Apart from the above typos, this is what I get when I rename a file by first typing "ها" an then "ha", with a space between them (and with LTR layout/locale):

ha

Please see the difference between the file name and what is shown on the status-bar. I think that such names were shown correctly with my closed PR but I may be wrong because it was a long time ago.

tsujan:
In my opinion, if the above typos are corrected, this will be a good approach. The above-mentioned problem is negligible if one sees the same thing in Dolphin, for example ;)

pcmanfm/tabpage.h
```diff
@@ -200,6 +200,10 @@ class TabPage : public QWidget {
 
     void setCustomizedView(bool value);
 
+protected:
+    // Adds bidi marks (RLM/LRM) around the text for the statusbar.
+    QString encloseWithBidiMarks(QString text);
+
```
tsujan: This can be added to the `private:` as:
```cpp
QString encloseWithBidiMarks(const QString& text);
```

SafaAlfulaij:
Sorry, by mistake!

I think now even the cases that we said not to support are supported now :D

tsujan:
> I think now even the cases that we said not to support are supported

Yes, the code seems to do that; will test it later.

pcmanfm/tabpage.h
```diff
@@ -217,6 +217,9 @@ protected Q_SLOTS:
     void freeFolder();
     QString formatStatusText();
 
+    // Adds bidi marks (RLM/LRM/RLE/LRE/POP) around the text for the statusbar.
+    QString encloseWithBidiMarks(QString& text);
+
```
tsujan: A `const` is missing from the argument, which results in:
```
error: binding reference of type ‘QString&’ to ‘const QString’ discards qualifiers
```
Please always try to compile the branch before committing! Thanks!

SafaAlfulaij: I did, but forgot about this pointers thing!
_I actually can't understand pointers to the deep :)_

tsujan:
Now it solves the problem completely. GTM.

The same conclusion, here: https://github.com/lxde/pcmanfm-qt/pull/382/files, but this is cleaner.

Apparently, wherever quotes, parentheses, brackets, etc. are used inside a potentially LTR-RTL mixed string, this method should be used. Dolphin has a similar issue.

pcmanfm/tabpage.h
tsujan: It's good. I'll fix small problems after merging but I can't find any now.

tsujan: Oh, found one, which isn't small. This:
```cpp
(!layoutDirection() == Qt::LeftToRight) ? "\u200f" : "\u200e"
```
You mean the following; don't you?
```cpp
layoutDirection() == Qt::RightToLeft ? "\u200f" : "\u200e"
```
Please fix them and test the result!

PR #608, opened 2017-12-04
Handle cmake cmp0071

luis-pereira:


luis-pereira:
Part of the fix to https://github.com/lxde/lxqt/issues/1420

tsujan:
I don't know the story behind `SKIP_AUTOGEN` but this works as intended.

tsujan:
We already have a policy conditional for CMP0063. Doesn't another one for CMP0071 work?

luis-pereira:
@tsujan 
`SKIP_AUTOGEN` works fine, I know. I expected that it woudn't be needed.
The documentation says that:
> If any of the files should be  processed, set CMP0071 to NEW.

But none of our files should be processed. They were already processed. Setting CMP0071 to NEW will apply to all files. We don't want that.

CMP0063 is a different case. We want the NEW policy to be applied. The OLD policy was not what we wanted anyway.

tsujan:
@luis-pereira I'm not good at cmake. Thanks for the explanation!

luis-pereira:
By the way ```CMP0063``` should not be there. It's set in `LXQtCompilerSettings` module. It's an artifact from the times when `pcmanfm-qt` did not use `lxqt-build-tools`. Adressed in https://github.com/lxde/pcmanfm-qt/pull/610.

agaida:
GTM

PR #616, opened 2017-12-16
move config to /usr/share/pcmanfm-qt/lxqt

agaida:


tsujan:
GTM.

palinek:
While at it.... the handling of settings in pcmanfm-qt is IMO wrong -> https://github.com/lxde/pcmanfm-qt/blob/master/pcmanfm/settings.cpp#L149

We are trying to do by hand (and not fully compatible) the very same as the `QSettings` (properly used) does provide by itself.

tsujan:
> ... the handling of settings in pcmanfm-qt is IMO wrong...

It's not wrong but I agree that it might be better to rely on QSettings.

palinek:
> It's not wrong but I agree that it might be better to rely on QSettings.

By "wrong" I meant unneeded manual processing and not fully compatible (it loads the keys/values **only** from the first file found, not considering other existing files which may contain other key/value pairs).

tsujan:
I know and, as told above, I agree with you -- the word "wrong" just had an extra connotation.

Issue #636, opened 2018-01-01
Add symlink info to status bar


aramazano:
In pcmanfm-qt symlinks are represented with a different icon, but that's all. No other information is shown in the main (detailed list) view, nor in the status bar. I have to look into properties of a symlink to see which file it is pointing to.

In detailed view under the "file type" column, *destination* file type is shown for a symlink (e.g. "plain text file"), whereas it would be better if it read "symbolic link" or "symlink to plain text file" or something like this.

In the status bar it repeats the file name and destination file type, whereas it would be better to show the structure of the symlink, something like "sample-link.txt -> ../../somedirectory/real-file.txt"

##### System Information

* Distribution & Version: Debian Stretch (9.3)
* Kernel: 4.9.0-4-amd64
* Qt Version: 5.7.1
* lxqt-build-tools Version: ?
* libqtxdg Version: ?
* libfm-qt Version: libfm-qt3:amd64 0.11.2-1
* Package version: PCManFM-Qt 0.11.3

tsujan:
Symlinks are clearly distinguished from other files by their emblems and their file types are definitely those of their targets.

The only valid feature requested here is some symlink info on the statusbar -- although I'm not sure about cluttering the statusbar.

aramazano:
There are certain situations where I systematically change some symlinks in a directory. When I enter such a directory, I feel clueless regarding which symlink is pointing to which file lately. Then I have to check each symlink's properties to get a picture of the latest setup.

I believe showing symlink structure in the status bar is more of a usability enhancement than clutter.

aramazano:
A more generalized solution might be showing the output of file(1) command in the status bar, without any post-processing at all. Parameters to the command could be adjusted so that the longest output (of file command) can still fit in the space provided by status bar (in case a file has too long a name, then output should be trimmed, so some post-processing might still be needed after all).

For instance something like this could be used for a reasonably concise output:

file -e cdf -e compress -e elf -e tar $FILE

Additionally, full output of the "file" command could be added in the properties dialog. I think That would be cool. This would give better view/control to users, while not getting in the way, I think

So, info gathering part is delegated to the file command, and pcmanfm-qt does only the presentation part. Full info in the right-click dialog should not clutter user experience since it is neatly hidden away in normal use. Since only one file's properties processed at a time, it won't hog the CPU or memory.

[...]

PR #641, opened 2018-01-08
Add symlink target info to statusbar

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/636.


agaida:
Works fine and is GTM - but who is responsible for the translations showed?

tsujan:
> but who is responsible for the translations showed?

The translators. They modify/complete their translations once in a while, and that's the case with all localized apps.

PR #642, opened 2018-01-30
Drop usage of Fm::IconTheme

palinek:
..as it is not needed any more.

Note: must be merged before or right after lxde/libfm-qt#165


tsujan:
Please remove `libfm-qt/icontheme.h` from `autorundialog.cpp` and then GTM.

palinek:
> Please remove libfm-qt/icontheme.h from autorundialog.cpp and then GTM.

Done..

agaida:
GTM

PR #667, opened 2018-05-06
Migrate to the new libfm-qt Fm::FileLauncher API.

PCMan:
This needs https://github.com/lxqt/libfm-qt/pull/186

### DIFF
--- a/pcmanfm/launcher.cpp
+++ b/pcmanfm/launcher.cpp
@@ -37,7 +37,7 @@ Launcher::~Launcher() {
 
 }
 
-bool Launcher::openFolder(GAppLaunchContext* ctx, const Fm::FileInfoList& folderInfos, Fm::GErrorPtr& err) {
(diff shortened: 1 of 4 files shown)


tsujan:
@agaida I see them as an annoyance ;)

tsujan:
I don't want to start a discussion about this new trend -- I just dislike it -- but, for example, that "falls through blah blah" message made me do a wrong thing in libfm-qt PR 186 today. I realized that when searching for the cause of a real problem, in contrast to false problems some of these warnings create. Whether I'm stupid or these new warnings is another question ;)

tsujan:
"BTW", found a real problem to fix here: the quick execution option isn't respected.

agaida:
btw: some annoyances quickly mutate to real bugs - esp. when maintainers switch warnings to bugs in the compiler preset - and yes, i know exactly what i'm talking about. Bitten by debians presettings more than one time - do you remember the g_assert bug in libfm?

tsujan:
I don't deny the fact that some of them are not only useful but necessary. However, even things like usefulness or happiness have a limit, beyond which they would be senseless :D

agaida:
GTM

PR #675, opened 2018-05-19
Bumped minor version to 13

agaida:


tsujan:
Ignore this if irrelevant: Does pcmanfm-qt always depend on the latest libfm-qt?

agaida:
it is not irrelevant  - right now it should, but we never check it - not so clever
```
find_package(fm-qt REQUIRED)
```

agaida:
lxqt/pcmanfm-qt/pull/677 - would be possible, right now it is up to the package maintainers to make sure that the right™ version of lifbm-qt is used

tsujan:
Interesting. I thought the dependency was hidden somewhere.

agaida:
the problem with the most cmake scripts is that they are far to complex

Best example might be quassel - maybe i'm only to old fashioned, but i like the idea to have all dependencies in the "root" CMakeLists.txt instead in the components. Same for options, as far as possible. And i really do _not_ like magic in these files.

PR #677, opened 2018-05-20
Set required minimum versions

agaida:


tsujan:
I like this. Sometimes -- as recently -- pcmanfm-qt needs the latest libfm-qt, so it's better to make the dependency always explicit.

agaida:
bah - wrong :D

agaida:
we should take the lib versionion instead of the package version i guess

tsujan:
> we should take the lib versionion instead

OK.

PR #682, opened 2018-05-26
Prefer theme icons for view actions

tsujan:
Fixes https://github.com/lxqt/pcmanfm-qt/issues/618

Only for pcmanfm-qt -- the file dialog still uses style icons.

For a good look and feel, theme icons should be used as far as possible, unless all icons are hard-coded, which isn't the case for pcmanfm-qt.


PCMan:
GTM

PR #690, opened 2018-06-05
Correctly set the required version of libfm-qt.

PCMan:
Now pcmanfm-qt requires libfm-qt API 0.14.0 (ABI version / so version: 6.0.0)

I noted that previously the ABI version (so version) "5.0.0" is mistakenly written into the libfm-qt cmake config version module while API version "0.13.1" should be used.
So I fix this, but this caused the version number returned by the find libfm-qt cmake module decreased from 5.0.0 to 0.14.0.
Not sure if this will cause side effects.

This PR is for the changes in https://github.com/lxqt/libfm-qt/pull/203


luis-pereira:
@PCMan This will mess things up. 
Will take a look into https://github.com/lxqt/libfm-qt/pull/203.

luis-pereira:
Hmmm packaging wasn't using `find_package()` version number. So this only affect us.
Making some additional checks

PR #692, opened 2018-06-06
Fix several FTBFS related to libfm dependecy drop

luis-pereira:
Just remove the libfm headers and include libfm-qt ones.


PCMan:
@luis-pereira thank you. sorry that I missed these wrong headers.

luis-pereira:
@PCMan No problem. I uninstalled libfm so it was easy to spot them.

PR #694, opened 2018-06-10
Fix the functionality of "Connect to remote server"

yan12125:
By correcting handling of URIs

Depends on https://github.com/lxqt/libfm-qt/pull/210

Testing: install `gvfs` package and connect to an sftp server via Go -> Connect to remote

PS. This bug is originally reported at the Arhc Linux bug tracker: https://bugs.archlinux.org/task/58940


tsujan:
LGTM

PR #723, opened 2018-07-11
Fix DND and drop indicator on desktop

tsujan:
Fixes https://github.com/lxqt/pcmanfm-qt/issues/706 and https://github.com/lxqt/pcmanfm-qt/issues/722

(1) When items are dragged and dropped on any desktop position other than that of a directory, they will stick to it (and to its following positions), moving next sticky items if needed. When desktop items are dropped on themselves or on a directory, the DND menu will be shown.

(2) The drop rectangle (when dragging desktop items on desktop folders) is also fixed by drawing it explicitly.


agaida:
LGTM

tsimonq2:
LGTM.

PR #725, opened 2018-07-12
Fix closing tab on ejecting/unmounting

tsujan:
Fixes https://github.com/lxqt/pcmanfm-qt/issues/695

When unmounting/ejecting is done from the places view, we should wait for the process to be finished (and auto-destroyed) and only then, remove the tab page; otherwise, the process might be finished with error because its object is a child of the places view.

For the sake of certainty, the same thing is done in the case of changing directory to user's home (I've seen random crashes with various backtraces otherwise).

Please also note that `QObject::destroyed` is used instead of `MountOperation::finished` intentionally, for cases of multiple operations (which may be practically impossible).


agaida:
will have a look later

agaida:
GTM

PR #755, opened 2018-08-13
Consider detailed view switch with transient filter-bar

tsujan:
When the current view is switched to or from the detailed list, it's deleted and recreated. So, we should re-install our event filter in such cases; otherwise, the transient filter-bar won't work until the window is closed and reopened.


tsujan:
This was just missing from https://github.com/lxqt/pcmanfm-qt/commit/71c73c9eccf75a4540a5e80ad416a1929c8316c8 ; hence merging.

PR #760, opened 2018-08-26
Added split view

tsujan:
Closes https://github.com/lxqt/pcmanfm-qt/issues/104

The split view can be toggled by `F6`. Switching between view frames can be done by the `Tab` key. Also, when the single tab of a split view frame is closed, the split mode is changed to the simple one.

For now, the split view has only two frames but the code is written in such way that more frames will be possible later.

Please note that the "full-width tab bar" was incompatible with the code of split view and so, is removed. IMHO, preserving it would require an unmaintainable code.

### DIFF
--- a/pcmanfm/main-win.ui
+++ b/pcmanfm/main-win.ui
@@ -32,19 +32,6 @@
      <number>0</number>
     </property>
     <item>
-     <widget class="PCManFM::TabBar" name="tabBar" native="true">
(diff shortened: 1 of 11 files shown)


agaida:
Really nice - GTM

agaida:
i'm kind of sorting my company nextcloud a bit - so yes, i will do.

PR #763, opened 2018-09-01
Split-view fixes for inactive windows and tabbar

tsujan:
This patch fixes 2 issues in the split view:

 * When the active view frame of an *inactive* window was removed, the background of the other view didn't change to normal.
 * When the tab bar of an inactive view frame was clicked, its view wasn't focused.

In addition:

 * Instead of comparing whole palettes, which can be "slow", only their base colors are compared.
 * Tab bars are made non-focusable for better focusing of views.
 * An old comment (about hiding network) is removed because it wasn't pertinent anymore.

### DIFF
--- a/pcmanfm/mainwindow.cpp
+++ b/pcmanfm/mainwindow.cpp
@@ -60,6 +60,7 @@ ViewFrame::ViewFrame(QWidget* parent):
     vBox->setContentsMargins(0, 0, 0, 0);
 
     tabBar_ = new TabBar;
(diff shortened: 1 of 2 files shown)


tsujan:
The first bug is hard to detect. Choose `Preferences → Volume → Close tab containing removable media`, create a new folder, go into it from the active view frame and then, remove that folder with QTerminal, *while pcmnanfm-qt's window is inactive*. Previously, the split mode was removed but the background of the remaining view didn't change. Now, everything is OK.

The second issue should have been fixed before split views but was annoying with them.

PR #810, opened 2018-11-07
Optional Desktop shortcuts

tsujan:
Closes https://github.com/lxqt/lxqt/issues/812

They can be added/removed in `Desktop Preferences` → `Advanced` (the tab got a real gob, at last). By default, there is no shortcut but distros can change that easily.

The Trash shortcut can empty Trash and files can be dropped into it (the drop indicator is shown). Its icon and text are changed when needed.

NOTES:

  * When seen inside a pcmanfm-qt *window*, the shortcuts are like ordinary files but if deleted in any way, they will be recreated with the app restart.
  * If there are already desktop entries with the same file-names, they will be updated.
  * It isn't easy to test this patch under all possible circumstances. So, IMO, it should be tested for a while.

### DIFF
--- a/pcmanfm/desktopwindow.cpp
+++ b/pcmanfm/desktopwindow.cpp
@@ -78,7 +78,8 @@ DesktopWindow::DesktopWindow(int screenNum):
     desktopHideItems_(false),
     screenNum_(screenNum),
     relayoutTimer_(nullptr),
-    selectionTimer_(nullptr) {
+    selectionTimer_(nullptr),
+    trashMonitor_(nullptr) {
 
     QDesktopWidget* desktopWidget = QApplication::desktop();
     setWindowFlags(Qt::Window | Qt::FramelessWindowHint);
@@ -101,6 +102,8 @@ DesktopWindow::DesktopWindow(int screenNum):
     // Then we paint desktop's background ourselves by using its paint event handling method.
     listView_->viewport()->setAutoFillBackground(false);
 
+    Settings& settings = static_cast<Application* >(qApp)->settings();
+
     // NOTE: When XRandR is in use, the all screens are actually combined to form a
     // large virtual desktop and only one DesktopWindow needs to be created and screenNum is -1.
     // In some older multihead setups, such as xinerama, every physical screen
@@ -108,7 +111,6 @@ DesktopWindow::DesktopWindow(int screenNum):
     // In this case we only want to show desktop icons on the primary screen.
     if(desktopWidget->isVirtualDesktop() || screenNum_ == desktopWidget->primaryScreen()) {
         loadItemPositions();
-        Settings& settings = static_cast<Application* >(qApp)->settings();
 
         setShadowHidden(settings.shadowHidden());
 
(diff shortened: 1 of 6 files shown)


pcmanfm/desktopwindow.cpp
```diff
@@ -199,6 +207,206 @@ DesktopWindow::~DesktopWindow() {
...
+    }
+    g_key_file_set_string(kf, "Desktop Entry", "Name", name.toLatin1().constData());
+
+    auto path = Fm::FilePath::fromLocalPath(XdgDir::readDesktopDir().toStdString().c_str()).localPath();
+    auto trash_can = Fm::CStrPtr{g_build_filename(path.get(), "trash-can.desktop", nullptr)};
+    g_key_file_save_to_file(kf, trash_can.get(), nullptr);
+    g_key_file_free(kf);
+}
+
+void DesktopWindow::createHomeShortcut() {
```
agaida: Only a idea - if we pre-create these four files and put them somewhere™ (/usr/share/applications/lxqt/$foo.desktop or so) we could just copy them - the adventage would be: We could translate the desktop icons. If the idea sound nuts please ignore it.

tsujan: You don't need to create them; they will be created automatically if the new config key (`DesktopShortcuts` under the `[Desktop]` section) has them -- and it will if the checkboxes are checked. Distros can make it like this:
```
...
DesktopShortcuts=Home, Trash, Computer, Network
...
```
They will also be removed automatically if checkboxes are unchecked. So, the user doesn't need to worry about desktop entries.

You'll see it when you test it.

tsujan: As for localization, the desktop entries will be created according to new translatable strings. I've done it for Trash but forgot the other items. Good catch! Will add a commit very soon.

tsujan: Done! Now all names (what users see) are localized.

Even if a bad distro does the bad thing and adds dekstop entries to Desktop, they will be updated/overwritten once pcmanfm-qt starts -- in a localized way.

agaida:
@tsujan - we need this to be tested:
image

agaida:
this too:
grafik

agaida:
maybe this too:

grafik

btw, looks good :)

tsujan:
If I'm not wrong, it's the same thing I've seen for a month or 2: it happens when a root--owned file/directory is emptied. Is that it?

agaida:
deleted a root-owned thing from somewhere month ago/years ago, yes - it was the dev-things for my tegra

agaida:
grafik

tsujan:
For me, the trash emptied correctly after those error dialogs. Please open an issue for it! I thought I was the only one ;)

agaida:
image

tsujan:
BTW, could you translate the new strings to German offline and see what happens with your localization?

agaida:
so - it works fine, with a little but. GTM

Re translations: i can at least try it

tsujan:
> it works fine, with a little but.

Tell me more about the "but". Anything other than those messages? They existed before this commit.

agaida:
The but is mostly: I had emptied my trash bin on a regular base (mostly one time per month or so) - so i was a bit surprised when this things come up.

PR #812, opened 2018-11-08
Don't update desktop trash faster than 4 times every sec

tsujan:
Fixes https://github.com/lxqt/pcmanfm-qt/issues/811

Because too many writes can block input and the desktop isn't refreshed more frequently either.

The result was quite acceptable when I dropped ~5000 files (chromium cache) into desktop trash and then emptied it.

NOTE: A part of https://github.com/lxqt/pcmanfm-qt/issues/811 isn't related to desktop trash and should be reported again.


PCMan:
LGTM

PR #855, opened 2019-01-02
Change text string in main-win.ui

Atalanttore:
Change unnecessary upper case letter to lower case letter.


tsujan:
"this" is among words whose first letters should be capitalized with "title casing". So, the original string was correct.

PR #857, opened 2019-01-02
Change string in main-win.ui

Atalanttore:
Make text string consistent to other strings.


Atalanttore:
It is quite difficult to get information about possible mistakes if no developer answers questions on Weblate LXQt.

PR #871, opened 2019-01-20
Removed QDesktopWidget

tsujan:
Closes https://github.com/lxqt/pcmanfm-qt/issues/865

`QDesktopWidget` is obsolete and some of its methods don't work correctly. This patch removes it from the code by using `QScreen` methods directly.

As a result, problems with desktop update are also fixed, especially with virtual screens.

P.S. I made the PR because the patch was ready but it could wait until after 0.14.

### DIFF
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -24,7 +24,6 @@
 #include <QDBusConnection>
 #include <QDBusInterface>
 #include <QDir>
-#include <QDesktopWidget>
 #include <QVector>
 #include <QLocale>
 #include <QLibraryInfo>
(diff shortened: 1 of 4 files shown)


agaida:
erm - 0.14 will be around in debian stable for two years - so maybe it would be clever to put it into 0.14

tsujan:
> 0.14 will be around in debian stable for two years - so maybe it would be clever to put it into 0.14

I have nothing against that; just wanted to say that I had no haste. Here, it works well with or without virtual screens/multiple monitors.

agaida:
it applies, compile and install well - and the best is - it works well :D GTM

tsujan:
> and the best is - it works well :D

I'd tested it several times with 2 monitors with various alignments and it worked without problem. I'm using it for 2 weeks on my work laptop. So, if there's a problem, it should be very hidden; users will report it and I'll fix it :)

PR #877, opened 2019-01-21
Fixed desktop trash drop rectangle

tsujan:
Fixes https://github.com/lxqt/pcmanfm-qt/issues/876

Because of a previous commit, the drop indicator wasn't shown when the drag source was outside Desktop.

### DIFF
--- a/pcmanfm/desktopwindow.cpp
+++ b/pcmanfm/desktopwindow.cpp
@@ -1535,14 +1535,21 @@ void DesktopWindow::childDragMoveEvent(QDragMoveEvent* e) {
     // see DesktopWindow::eventFilter for an explanation
     QRect oldDropRect = dropRect_;
     dropRect_ = QRect();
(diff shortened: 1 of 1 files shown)


tsujan:
> The drop indicator should appear on "Benutzerverzeichnis" (user home) too.

But user home doesn't support DND; does it? DND is only for Trash (among desktop shortcuts).

tsujan:
Moreover, this fixes a regression; nothing more.

agaida:
hmm - $HOME should support DND i guess, it should behave like a "normal" directory folder. To be honest, never used it that way - but your explanation is sufficient for now, we can discuss this behaviour if nothing else is to do.

GTM

tsujan:
> we can discuss this behaviour if nothing else is to do.

OK :) Just this: when I added those shortcuts, I thought about that too but didn't see it worth the effort -- a decision that could be overridden.

PR #907, opened 2019-02-23
Point release pcmanfm-qt 0.14.1

agaida:


tsujan:
Good.

CHANGELOG
```diff
@@ -1,3 +1,17 @@
+pcmanfm-qt-0.14.0 / 2019-
```
Chiitoo: Should this be 0.14.1?

Also no date?  (Same for lxqt-config-0.14.1 it seems.)

agaida: it should - and i wanted to do some tests first before i set the date

agaida: thanks a lot, damn typos :D

Chiitoo: Alrighty!  Figured the date was just not set yet, but wasn't sure.  :]

PR #927, opened 2019-04-02
Tell users about dbus-launch

tsujan:
Because it may be needed for launching a root instance.

Also, use rich text to prevent confusion about quotation marks.

### DIFF
--- a/pcmanfm/preferences.ui
+++ b/pcmanfm/preferences.ui
@@ -704,11 +704,13 @@ only if there are more than one tab.</string>
             <item row="2" column="0" colspan="2">
              <widget class="QLabel" name="label_4">
               <property name="text">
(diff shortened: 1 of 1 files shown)


dariusity:
Function "Open As Root" not working in Manjaro. However it does work in Lubuntu.

Just to make sure the command is:
lxqt-sudo %s (original command)
lxqt-sudo --dbus-launch %s (revised command)

Both commands aren't working. Checked at Lubuntu it is using the original and it is working fine.

tsujan:
@dariusity 
It isn't `--dbus-launch` but `dbus-launch` ;) It works everywhere.

dariusity:
Ok, I think I got on the wrong bus then! Apologies.

tsujan:
It's OK; it was because of the title: "Root instance -- dbus-launch is...".

PR #929, opened 2019-04-03
WIP: Option to enable passing mousewheel to root window

dwrl:
Enables to not only pass middle and left click to the window manager for menus, but also to pass mouse wheel action (vertically and horizontally), to e.g. allow desktop switching.

Can be enabled/disabled via a separate checkbox in the settings window.

### DIFF
--- a/pcmanfm/desktopwindow.cpp
+++ b/pcmanfm/desktopwindow.cpp
@@ -58,6 +58,7 @@
 #include <QX11Info>
 #include <QScreen>
 #include <xcb/xcb.h>
+#include <xcb/xinput.h>
 #include <X11/Xlib.h>
 
 #define WORK_AREA_MARGIN 12 // margin of the work area
@@ -75,7 +76,8 @@ DesktopWindow::DesktopWindow(int screenNum):
     wallpaperTimer_(nullptr),
     wallpaperRandomize_(false),
     fileLauncher_(nullptr),
-    showWmMenu_(false),
+    passButtonsToRoot_(false),
+    passWheelToRoot_(false),
     desktopHideItems_(false),
     screenNum_(screenNum),
     relayoutTimer_(nullptr),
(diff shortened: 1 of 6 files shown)


pcmanfm/desktopwindow.cpp
```diff
@@ -1523,6 +1584,17 @@ bool DesktopWindow::eventFilter(QObject* watched, QEvent* event) {
...
+                // Forward wheel events we received to the root window.
+                // check if the user is over a blank area
+                QModelIndex index = listView_->indexAt(e->pos());
+                if(!index.isValid()) {
+                    forwardWheelEventToRoot(e);
+                    return true;
+                }
+            }
+            break;
             return true;
```
tsujan: Please correct this: `return true;` isn't reached when `passWheelToRoot_` is `false`. Long story short, `return true;` is needed with item DND.

dwrl: Done.

tsujan: Thanks!

pcmanfm/desktopwindow.h
```diff
@@ -175,7 +175,8 @@ protected Q_SLOTS:
     bool wallpaperRandomize_;
     QPixmap wallpaperPixmap_;
     Launcher fileLauncher_;
-    bool showWmMenu_;
+    bool passButtonsToRoot_;
```
tsujan: We'll get a FTBFS here; please restore `showWmMenu_` and remove `passButtonsToRoot_`.

dwrl: Done.

tsujan:
OK. When I find time, I should do a test about inline renaming with your code and might ask you to change it a little (will explain why later). All in all, it looks good.

dwrl:
Cool, I'll wait.
Thank you.

tsujan:
@dwrl  I did my tests but you may not like the result.

What I wanted to do was not limiting wheel forwarding to when the cursor was on an empty space but doing it only when there's no open editor for inline renaming. That's easy to do because our delegate has the method `hasEditor()`.

However, when I enabled Openbox's desktop switching by wheeling up/down on desktop, I saw that it was like Compiz: *It didn't need any forwarding on our part.*

So, my conclusion is what I said at https://github.com/lxqt/pcmanfm-qt/issues/928: "... an X11 WM is able to achieve it without any change to pcmanfm-qt." I don't like more X11 codes just for specific WMs like fluxbox and fvwm. It's not our job to do so.

I leave the final decision to @agaida.

PR #937, opened 2019-04-12
Removed WM menu from desktop

tsujan:
Because:
(1) Only some X11 WMs have desktop context menus, while LXQt is WM-agnostic;
(2) It hid our FM desktop context menu; and
(3) X11 codes should be kept only when really useful because they are an obstacle to preparing for Wayland.

NOTE: "Advanced" is preserved for the tab name because an extra item is added to it outside LXQt.

Closes https://github.com/lxqt/pcmanfm-qt/issues/931

### DIFF
--- a/pcmanfm/desktop-preferences.ui
+++ b/pcmanfm/desktop-preferences.ui
@@ -507,22 +507,6 @@ A space is also reserved for 3 lines of text.</string>
         </widget>
        </item>
        <item>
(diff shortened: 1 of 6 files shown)


tsujan:
@agaida Thanks for merging all pcmanfm-qt and libfm-qt PRs! Now, I can start to fix minor issues.

PR #941, opened 2019-04-16
Remove unneeded Qt version check

luis-pereira:
Qt 5.7.1 is the minimum. The check is always true.


tsujan:
It was left there from old codes. Thanks!

PR #942, opened 2019-04-18
Prevent future crash if a new sort column is added in libfm-qt

tsujan:
If a new sort column is added to detailed list view in libfm-qt but there's no corresponding action in pcmanfm-qt's sort menu, pcmanfm-qt will crash. This patch prevents that.

Such a scenario is possible in 2 ways:

(1) The change may have been made to libfm-qt but not yet to pcmanfm-qt.
(2) Although not likely, we might really not want to add an action corresponding to a new sort column.

Anyways, pcmanfm-qt shouldn't crash.

### DIFF
--- a/pcmanfm/mainwindow.cpp
+++ b/pcmanfm/mainwindow.cpp
@@ -1185,14 +1185,31 @@ void MainWindow::updateViewMenuForCurrentPage() {
         modeAction->setChecked(true);
 
         // sort menu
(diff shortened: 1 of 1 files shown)


tsujan:
Please note that this patch doesn't change anything in pcmanfm-qt. But, when I added deletion time to libfm-qt here, I was happy that pcmanfm-qt didn't crash with it on sorting trashed items by deletion time. (Adding deletion time will take time.)

PR #950, opened 2019-04-24
Enhanced wallpaper browsing dialog

tsujan:
Use LXQt file dialog directly, set it to thumbnail view mode, select and scroll to the current wallpaper in it, and remember its size and splitter position.

The slide-show dialog is still the default LXQt file dialog because it only shows folders.

Closes https://github.com/lxqt/lxqt/issues/391

### DIFF
--- a/pcmanfm/desktoppreferencesdialog.cpp
+++ b/pcmanfm/desktoppreferencesdialog.cpp
@@ -30,6 +30,7 @@
 #include <QRegExp>
 #include <QDebug>
 #include <QStandardPaths>
+#include <libfm-qt/filedialog.h>
 
 namespace PCManFM {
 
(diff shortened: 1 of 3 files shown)


tsujan:
@agaida I want to enhance this PR, so that LXQt file dialog is called directly in the thumbnail view mode (also see https://github.com/lxqt/libfm-qt/issues/412). After all, pcmanfm-qt depends on libfm-qt. Will tell when finished.

tsujan:
Finished.

Now, LXQt file dialog is directly used in thumbnail mode and the user can resize it and its splitter independently from the default file dialog, like this:

wallpapaer_dialog

agaida:
Really nice. GTM.

PR #957, opened 2019-05-10
Strings

luis-pereira:


pcmanfm/preferencesdialog.cpp
```diff
@@ -80,15 +80,15 @@ void PreferencesDialog::initIconThemes(Settings& settings) {
         // load xdg icon themes and select the current one
         QHash<QString, QString> iconThemes;
         // user customed icon themes
-        findIconThemesInDir(iconThemes, QString(g_get_home_dir()) % "/.icons");
+        findIconThemesInDir(iconThemes, QFile::decodeName(g_get_home_dir()) + QStringLiteral("/.icons"));
```
tsujan: Thank you very much for the hard work!

Isn't `QString::fromUtf8` better here? We haven't used `QFile::decodeName` anywhere else — in pcmanfm-qt or libfm-qt.

pcmanfm/desktopwindow.cpp
```diff
@@ -532,7 +532,7 @@ QImage DesktopWindow::loadWallpaperFile(QSize requiredSize) {
 
         // read info file
         QString origin;
-        info.setFileName(cacheFileName % ".info");
+        info.setFileName(cacheFileName + QStringLiteral(".info"));
```
tsujan: The are still other cases of ` % ` in `desktopwindow.cpp` and `settings.cpp`. Now that you've replaced them in some places, wouldn't it be better to replace them completely and then remove `#include <QStringBuilder>` from:
```
desktopwindow.cpp
preferencesdialog.cpp
settings.cpp
```

pcmanfm/preferencesdialog.cpp
tsujan: Actually, `QFile::decodeName` is (in Qt → `qfile.h`):
```c++
QString::fromUtf8(NAME).normalized(QString::NormalizationForm_C);
```
The `normalized` part should have been already done in `g_get_home_dir`.

PR #985, opened 2019-07-17
Added Alt+Enter in addition to Alt+Return for properties

tsujan:
To prevent confusion about Enter and Return


agaida:
Now i get fully what you mean :D 

Lucky me that there was no hidden cam connected to youtube or so. Lesson learned, Return is Return, Enter is Enter. So if i understand it right it should even work translated or am i wrong? (if someone translate the Keynames)

tsujan:
> Return is Return, Enter is Enter

For years, I haven't had any keyboard with a "Return" label ;) I think this confusion is historical. But, yes, it's as you said: Return is Return (= a main key), Enter is Enter (in the keypad section of the keyboard).

agaida:
GTM

PR #997, opened 2019-08-22
Wait: Always show display names on Desktop

tsujan:
…in contrast to full names. The reason is that desktop entries have a special usage on Desktop.

In this patch, Desktop (as what is drawn behind all windows) has its own folder model. So, https://github.com/lxqt/libfm-qt/pull/460 is needed for a proper functioning.

Also, two defaults are changed: full names are shown by default inside pcmanfm-qt windows; hidden icons are shadowed by default too.

Closes https://github.com/lxqt/pcmanfm-qt/issues/996

NOTE: Merge/use it only after https://github.com/lxqt/libfm-qt/pull/460 is merged/applied!

### DIFF
--- a/pcmanfm/desktopwindow.cpp
+++ b/pcmanfm/desktopwindow.cpp
@@ -115,8 +115,9 @@ DesktopWindow::DesktopWindow(int screenNum):
         setShadowHidden(settings.shadowHidden());
 
         auto desktopPath = Fm::FilePath::fromLocalPath(XdgDir::readDesktopDir().toStdString().c_str());
-        model_ = Fm::CachedFolderModel::modelFromPath(desktopPath);
-        model_->setShowFullName(settings.showFullNames());
+        model_ = new Fm::FolderModel();
+        model_->setFolder(Fm::Folder::fromPath(desktopPath));
+        model_->setShowFullName(false); // always show display name on Desktop
         folder_ = model_->folder();
         connect(folder_.get(), &Fm::Folder::startLoading, this, &DesktopWindow::onFolderStartLoading);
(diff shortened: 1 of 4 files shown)


tsujan:
@agaida Everything has worked smoothly here. IMO, it's safe to merge. But please test it yourself before merging.

tsujan:
I mean don't trust my "self-confidence" ;) I may have become too brave.

agaida:
I using it now for 4 days - and see no downsides - in other words - i love it.

Maybe a suggestion, should be a new issue:
Add an dialog to applications who need a restart after changed configurations and ask for the restart: I realized it for now with a simple line in the terminal
```
% sudo pkill --sig SIGHUP pcmanfm-qt
```
Second thing, also worth a new issue: Extend the quick file info with the real filename - would be very helpful esp. for desktop files on the desktop.

agaida:
GTM

tsujan:
> Add an dialog to applications who need a restart after changed configurations and ask for the restart

Later. I'm not sure we should do it.

> Extend the quick file info with the real filename

On the status-bar?

tsujan:
.... saw your reported issue. You mean tooltip.

tsujan:
> I using it now for 4 days - and see no downsides

That's more than enough for me. I may be cautious but don't have obsessive-compulsive disorder ;)

tsujan:
Off topic:

For some reason unknown to me, the memory usage of pcmanfm-qt has decreased after this -- I sometimes check it after a day of heavy work. It may be about how Qt/C++ handles memory; nothing to celebrate ;)

PR #1001, opened 2019-08-28
Respect view setting when showing selected file name on status bar

tsujan:
Show the real (full) name if the view shows it.


agaida:
GTM

PR #1006, opened 2019-09-05
Follow DE click setting if single-click is not chosen

tsujan:
This is the simplest and, IMO, best compromise between pcmanfm-qt's and LXQt's settings for single vs. double click. If single-click is chosen, it will be used; if not, the DE setting will be used (`QCommonStyle` gives it automatically), whether inside LXQt or not.

I also added a tooltip to the check-box.

NOTE 1: `QCommonStyle` is used intentionally because widget styles can enforce their click behavior (Kvantum does but I'll change that soon).

NOTE 2: Following the DE setting *completely* is not an option. A file manager should be able to control clicking because opening a file is more than activating a view-item.

→ https://github.com/lxqt/pcmanfm-qt/issues/309


agaida:
Cool, GTm

PR #1010, opened 2019-09-08
Added context menu item to create launchers on Desktop

tsujan:
The created desktop entries are trusted automatically.

### DIFF
--- a/pcmanfm/CMakeLists.txt
+++ b/pcmanfm/CMakeLists.txt
@@ -15,6 +15,7 @@ set(pcmanfm_SRCS
     connectserverdialog.cpp
     settings.cpp
     bulkrename.cpp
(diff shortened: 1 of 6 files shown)


stefonarch:
Works fine here. Wondering if this editor could be used by the quicklaunch widget, menu or else, creating a .desktop file in ~/.local/share/applications (maybe it could be offered in the right click menu there too).

tsujan:
> Wondering if this editor could be used by...

No, it's specific to pcmanfm-qt's Desktop and is for creating desktop entries (launchers) on it. A menu editor should be a stand-alone program and capable of creating, removing and editing menu entries. I might try to make one if my work area becomes so bugless and featureful that I have nothing to do for a while ;)

And thanks a lot for testing!

PR #1023, opened 2019-10-08
Do not show trash path in trashed file names

tsujan:
That's done by showing display names inside trash.

Also, show display names in computer:/// and network:///, like other file managers do. For example, `root.link` in computer:/// is a little weird but `File System` (its display name) is familiar. Real names can always be found in Properties dialogs.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1022


tsujan:
@agaida, I made it always show display names in special places (trash, computer and network in addition to menu) because (1) full names were very unusual in trash:///, computer:/// and network:/// and (2) other file managers do so.

If the user is curious about full names in those places, he could use the Properties dialog or take a look at path buttons/path edit. So, nothing will be missing.

PR #1043, opened 2019-12-03
Multimonitor wallpaper support

hmollercl:


pcmanfm/desktopwindow.cpp
```diff
@@ -214,6 +214,31 @@ DesktopWindow::~DesktopWindow() {
     }
 }
 
+bool DesktopWindow::isIndividual() {
+/*Check if individual wallpaper is needed and if screen is expanded or unified
+ based on initial point of sceen as used in fastmenu.cpp in lxq-config-monitor*/
```
tsimonq2: screen and lxqt are misspelled

hmollercl: solved

tsujan:
My first review:

OK, supposing that we forget about the original report and limit ourselves to the same wallpaper for all screens (which is acceptable to me), the current PR has the following problems — please fix all of them:

 1. The method `DesktopWindow::isIndividual()` and the variable `QList<QScreen*> screens` are redundant. Instead, inside the method `DesktopWindow::updateWallpaper()`, you could define `perScreenWallpaper` like:
 
```c++
...
     if(wallpaperMode_ != WallpaperNone) {  // use wallpaper
        Settings& settings = static_cast<Application* >(qApp)->settings();
        auto screen = getDesktopScreen();
        bool perScreenWallpaper(screen != nullptr && screen->virtualSiblings().size() > 1 && settings.individualWallpaper());
...
 ```
 
And use it like:
 
 ```
 ...
            if(perScreenWallpaper) {
                const QSize s = size();
                pixmap = QPixmap{s};
                QPainter painter{&pixmap};
                pixmap.fill(bgColor_);
                image = getWallpaperImage(); //when multiple files this need to go inside the for loop to change images.
                QImage scaled;
                const auto screens = screen->virtualSiblings();
                for(const auto& scr : screens) {
                    scaled = image.scaled(scr->size(), Qt::IgnoreAspectRatio, Qt::SmoothTransformation);
                    painter.drawImage(scr->geometry().x(), scr->geometry().y(), scaled);
                }
            }
...
 ```
 
And so on.
 
  2. The current diff for adding only a checkbox to `desktop-preferences.ui` isn't clean. Instead, you could just add these lines after line 333 (diffs should be clean as far as possible):

```xml
          <item row="4" column="0" colspan="2">
           <widget class="QCheckBox" name="individualWallpaper">
            <property name="text">
             <string>Individual wallpaper for each monitor</string>
            </property>
           </widget>
          </item>
```

 3. The places of `bool individualWallpaper() const` and `void setIndividualWallpaper(bool tr)` are wrong; they make the code untidy. Instead, they should come together appropriately. See other similar functions in `settings.h`.

 4. The default value for `individualWallpaper_` should be `false`, not `true`.

 5. The name `individualWallpaper` isn't good. Function/variable names should be so that the code reader could guess their meanings as easily as possible. Please change it to `perScreenWallpaper`.

tsujan:
@hmollercl Thanks for the changes! It's much better now. I'll review it again soon.

hmollercl:
> @hmollercl
> I just wanted to take a look at this without reviewing it (lack of time). This isn't what was requested at [lxqt/lxqt#1175](https://github.com/lxqt/lxqt/issues/1175); otherwise, I wouldn't say "it requires lots of changes to the current code".
> 
> Please re-read [lxqt/lxqt#1175 (comment)](https://github.com/lxqt/lxqt/issues/1175#issuecomment-557665888).

From that comment, the things that I see this doesn't have are:
1. different wallpapers for each screen
2. caching the wallpaper
The first one can be easily added, it need a ui definition though on how to handle the filepicking.
On the second one, I haven't seen the need for cache, probably I'm missing something but I really haven't seen the need for it, the wallpapers are created fast enough. That would also need a new cache file definition.
Are there anything else missing?

tsujan:
IMHO, implementation of different wallpapers for each screen isn't worth cluttering the current code.

As for caching, I'm not sure. I'll think about it.

> Are there anything else missing?

Apparently, no. Let me review it by paying attention to more details -- I'll do it as soon as I find the time. And thanks for the hard work!

pcmanfm/desktopwindow.cpp
```diff
@@ -633,39 +636,119 @@ void DesktopWindow::updateWallpaper() {
...
+            else {
+                if(wallpaperMode_ == WallpaperCenter) {
+                    image = getWallpaperImage(); // load original image
+                }
+                else if(wallpaperMode_ == WallpaperFit || wallpaperMode_ == WallpaperZoom) {
+                    // calculate the desired size
+                    QImageReader reader(wallpaperFile_);
+                    QSize origSize = reader.size(); // get the size of the original file
+                    if(reader.transformation() & QImageIOHandler::TransformationRotate90) {
+                        Settings& settings = static_cast<Application* >(qApp)->settings();
```
tsujan: @hmollercl 
Would you please remove the above line -- `settings` is already  defined -- and also include https://github.com/lxqt/pcmanfm-qt/pull/1045 in your PR (it's about adding a nullity check)? In that way, you won't need to re-base your PR and I'll merge it after testing it again.

hmollercl: removed that re-definition of settings and added #1045 code.

tsujan:
2 fixes in setting the image copy width and height (as well as a cleanup) were needed: https://github.com/lxqt/pcmanfm-qt/commit/15235f046725068651c2ae5f809c6d8bfde05156

PR #1045, opened 2019-12-07
Fix hang when the wallpaper is in tile mode and the file is missing

yan12125:
s.width() and s.height() are zero, so pcmanfm-qt hangs in an infinite
loop.

Noticed this when the partition with my wallpaper fails to mount.

### DIFF
--- a/pcmanfm/desktopwindow.cpp
+++ b/pcmanfm/desktopwindow.cpp
@@ -621,14 +621,16 @@ void DesktopWindow::updateWallpaper() {
         QImage image;
         if(wallpaperMode_ == WallpaperTile) { // use the original size
             image = getWallpaperImage();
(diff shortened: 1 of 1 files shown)


tsujan:
@yan12125 would you mind if I we put this into https://github.com/lxqt/pcmanfm-qt/pull/1043 to prevent re-basing of that PR?

yan12125:
> @yan12125 would you mind if I we put this into #1043 to prevent re-basing of that PR?

Sure, no problem. Please do, thanks!

tsujan:
Thank you! I'll leave it open until @hmollercl adds it to his PR and it's merged.

tsujan:
Done in https://github.com/lxqt/pcmanfm-qt/commit/dfdf1fa0db00330d927e664c8e6829c0835adfa7. Thank you very much for finding and eliminating the cause of a crash!

PR #1062, opened 2020-01-30
minor translatability enhancement

Fat-Zer:
Use plural-aware version of `tr()` to process a message for the message box.
And fix source English string to use plural-only for "files" here.


tsujan:
Thanks for fixing the typo.

PR #1075, opened 2020-02-15
Add option for transparent desktop background

0Ekho:
Allows for using other background manages or placing [status] windows
below the desktop and still being able to see them. Doing so is probably
not advised but seems to work OK.

Unfortunately know neither QT5 nor C++ so the patch probably needs some work / cleanup, apologies. Also probably a very niche use case, but being only ~20 lines it might still be worth adding (xfdesktop seems to have a "transparent" option looking at the UI but not sure if it's supposed to do quite the same thing).


0Ekho:
Unfortunately know neither QT5 nor C++ so the patch probably needs some work / cleanup, apologies. Also probably a very niche use case, but being only ~20 lines it might still be worth adding (xfdesktop seems to have a "transparent" option looking at the UI but not sure if it's supposed to do quite the same thing).

tsujan:
Setting `Qt::WA_TranslucentBackground` may have serious side effects when updating desktop, especially without a compositor.

0Ekho:
good to know, should it just be moved into only being set when transparent mode is selected and unset when deselected? or is it problematic to ever set at all?

tsujan:
I'm afraid, an option that breaks desktop under some circumstances is a bug, not a feature. Even if problems only happen without compositing (about which I'm not sure), we'll need to check compositing. That's possible but it requires X11 code, which we absolutely don't want.

All in all, the cost of this special usage is a damaged code, IMO.

0Ekho:
sounds completely reasonable, thank you for the time.

PR #1105, opened 2020-04-27
Fixed remounting of remote folders under special circumstances

tsujan:
Fixes https://github.com/lxqt/pcmanfm-qt/issues/1103

When a folder is unmounted and the option "Change folder in the tab to home folder" is selected, the folder shouldn't be freed but disconnected from all signals.


stefonarch:
Works fine now.

tsujan:
Thanks for the quick test! And keep on hunting bugs.

PR #1121, opened 2020-05-15
Safeguard against null images with multi-screen wallpapers

tsujan:
A nullity check for the wallpaper image is added in the case of individual wallpapers for multiple screens. In one case, a division by zero was possible without the check.

The nullity check already existed for other cases but an extra one is added for the sake of certainty.


tsujan:
There's nothing special here; only a few cases of `if(!image.isNull())` for covering unusual cases.

BTW, GitHub's diff view sucks...

PR #1131, opened 2020-06-02
Update German translation

ghostbunny:
Hi guys,

I updated the German translations to cover all current strings.

BR,

gb


tsujan:
Thanks.

PR #1133, opened 2020-06-07
Update pcmanfm-qt_he.ts

omeritzics:
I have been added translations for non-translated words.


tsujan:
Thanks.

tsujan:
@omeritzics 

I merged your PR but it seems you didn't use Qt Linguist. Something like `<translation type="פעולה"></translation>` is incorrect ("type" shouldn't be there) and might cause problems with your localization. Such problems won't arise if you use Qt Linguist correctly. Please fix it in another PR!

I sometimes check translations before merging but not always (I don't know the target languages most of the time).

PR #1172, opened 2020-07-29
Unminimize window when switching to new tab is enabled

tsujan:
When the user enables switching to newly created tab, the window shouldn't be left minimized when a new tab is added. This is mostly relevant to the single window mode but isn't limited to it.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1109


tsujan:
Any objection to merging this?

P.S. I don't want to take other devs' times by reviewing my libfm-qt/pcmanfm-qt PRs because they have other LXQt jobs. That's why I merge these PRs after testing them (usually for relatively long times if they're complex). But feel free to self-request a review if you have the time; I'd really appreciate that.

luis-pereira:
Looks good. Ship ir.

PR #1173, opened 2020-07-29
reopen last tabs feature like in ff or chrome

coding-moding:
Very convenient. No need to reopen tabs you working with.
did you know?)

PR #1174, opened 2020-07-30
reopen last tabs like in ff or chrome

coding-moding:


tsujan:
@coding-moding 
I'm confused. You made 2 PRs for the same feature, this and https://github.com/lxqt/pcmanfm-qt/pull/1173. Let me explain the basics of GitHub:

 1. The contributor makes one, and only one, PR for each change to each source.
 2. The PR is marked for a review and, after a while, its review starts.
 3. The review may require changes to be made by the coder. The coder may be asked several times to change/fix this or that part of his/her code.
 5. If everything goes well, the PR will be merged and becomes a part of the program.

Now, please tell me which PR of yours can be reviews and which one should be closed.

coding-moding:
#1173 is for 0.14.1 branch, and can be deleted
#1174 is for master branch

tsujan:
@coding-moding 
Thanks for your reply!

The development is done only on the master branch. LXQt doesn't have LTS; 0.14.1 is dead and 0.15.1 is old.

Before I review this, please fix the following issues in it (by adding new commits to it):

 * We shouldn't have different coding styles in the same source. Please take a look at the original coding style of the sources and respect it in your patch, exactly as it is.
 * Remove your commit under the title "give user a feedback when he changes prefs". Each PR should be about one subject; different changes shouldn't be mixed in one PR.

coding-moding:
> PR (it should be removed from here).

how do i remove it? it was added automatically. i wanted to create a different PR.

tsujan:
> how do i remove it?

What do you mean? You can remove lines by adding a commit that removes them.

> i wanted to create a different PR.

I you create another PR for every change, reviewing will be impossible.

tsujan:
> it was added automatically

No commit is added automatically.

tsujan:
The code style aside (I already mentioned it twice), 4 problems are visible at first glance:

 1. The patch causes crash when a tab is dragged and dropped out of the window (= detached).
 2. It creates empty tabs when some folders don't exist. With a file manager, empty tabs don't make sense at startup and without any command argument.
 3. It doesn't consider the split view and that there may be multiple tabs in each view.
 4. It especially gets ugly with password protected mounts.

tsujan:
Thanks for the main idea. Superseded by https://github.com/lxqt/pcmanfm-qt/pull/1177.

PR #1177, opened 2020-08-02
Added an option to open tabs of last window

tsujan:
This PR supersedes https://github.com/lxqt/pcmanfm-qt/pull/1174 because explaining all details there would take much more time and energy than making a clean PR.

When the new option is checked (it's unchecked by default), the following behaviors are implemented:

 1. The first window loads the tabs of the last closed window (and nothing else). Subsequent windows are opened in the usual way until the last window is closed and its tabs are remembered.

 2. If some folders are opened in the first window *explicitly*  (e.g., by clicking a desktop folder or adding paths to command-line), the last tabs will be ignored with it but they will be reopened in the next window that doesn't explicitly open a folder. That's intentional because the feature shouldn't interfere with opening of folders.

 3. If some folders of the last tabs are removed or inaccessible, libfm-qt's error messages will be shown. They inform the user about the non-existing folders without opening empty tabs. If all folders are inaccessible and there is no window yet, the current work directory will be opened.

### DIFF
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -249,16 +249,18 @@ bool Application::parseCommandLineArgs() {
         }
         else {
             if(!parser.isSet(desktopOption) && !parser.isSet(desktopOffOption)) {
+                bool reopenLastTabs = false;
                 QStringList paths = parser.positionalArguments();
                 if(paths.isEmpty()) {
                     // if no path is specified and we're using daemon mode,
                     // don't open current working directory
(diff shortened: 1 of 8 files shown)


tsujan:
@PCMan
Since yesterday, I assumed you're available for reviews. If not, feel free to ignore my review request(s).

tsujan:
> Maybe: Reopen last window tabs in a new window is clearer.

I agree; didn't like it either. Will change it to what you suggested before merging. Thanks!

PR #1179, opened 2020-08-18
Remove redundant Places keys

### LINKED ISSUE #1137: Remove redundant keys from the code
Under the `Places` section of pcmanfm-qt's settings file, there are 7 old keys that have no effect anymore. They were included for hiding specific `Places` in the `Lists` side-pane with older versions of libfm-qt but now, libfm-qt treats `Places` and `Devices` in the same way with regard to hiding, so that all the needed info is provided by the key `HiddenPlaces`.

Of course, removing them from the code doesn't mean they'll be removed from an already existing settings file but they won't be read/written.

keysemble:
Closes #1137

### DIFF
--- a/pcmanfm/settings.cpp
+++ b/pcmanfm/settings.cpp
@@ -124,13 +124,6 @@ Settings::Settings():
     showFullNames_(true),
     shadowHidden_(true),
     noItemTooltip_(false),
-    placesHome_(true),
-    placesDesktop_(true),
-    placesApplications_(true),
-    placesTrash_(true),
-    placesRoot_(true),
-    placesComputer_(true),
-    placesNetwork_(true),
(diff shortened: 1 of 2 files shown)


keysemble:
Would you folks mind if push another commit to clean the QList warnings?

tsujan:
> Would you folks mind if push another commit to clean the QList warnings?

It's too soon. The minimum Qt dependency is 5.10 and (some of) those warnings are about Qt 5.15. We don't want version checks in the code, as far as possible.

keysemble:
Glad to help. I've been looking for ways to improve my C++. Thought this issue was a good starter to get familiar with the code.

tsujan:
Indeed, it was a big help. Thanks!

PR #1182, opened 2020-08-22
Don't remove `@Invalid()` from config file

tsujan:
`@Invalid()`, as a value, is Qt's way of announcing an empty list in config files. I had removed keys with empty lists but that could damage code readability and maintenance. This patch restores it.


tsujan:
Straightforward; hence merging...

PR #1216, opened 2020-10-20
Added an option to open desktop folders in default file manager

tsujan:
The option is in the first tab of Desktop Preferences and is unchecked by default. It has effect when desktop folders are launched in the usual way, i.e., by left clicking, and the text label is explicit about that. Middle/right clicking triggers the default behavior.

NOTE: For user convenience, the home shortcut will also be opened by the default file manager if the option is checked but the other shortcuts are always opened by pcmanfm-qt because other file managers may not be able to handle them.

Closes https://github.com/lxqt/lxqt/issues/1823

### DIFF
--- a/pcmanfm/launcher.cpp
+++ b/pcmanfm/launcher.cpp
@@ -28,7 +28,8 @@ namespace PCManFM {
 Launcher::Launcher(PCManFM::MainWindow* mainWindow):
     Fm::FileLauncher(),
     mainWindow_(mainWindow),
-    openInNewTab_(false) {
+    openInNewTab_(false),
+    openWithDefaultFileManager_(false) {
 
     Application* app = static_cast<Application*>(qApp);
     setQuickExec(app->settings().quickExec());
@@ -38,12 +39,26 @@ Launcher::~Launcher() {
 
 }
 
-bool Launcher::openFolder(GAppLaunchContext* /*ctx*/, const Fm::FileInfoList& folderInfos, Fm::GErrorPtr& /*err*/) {
+bool Launcher::openFolder(GAppLaunchContext* ctx, const Fm::FileInfoList& folderInfos, Fm::GErrorPtr& /*err*/) {
     auto fi = folderInfos[0];
     Application* app = static_cast<Application*>(qApp);
     MainWindow* mainWindow = mainWindow_;
     Fm::FilePath path = fi->path();
     if(!mainWindow) {
+        // Launch folders with the default file manager if:
+        //   1. There is no main window (i.e., folders are on desktop),
+        //   2. The folders are not supposed to be opened in new tabs, and
+        //   3. The default file manager exists and is not PCManFM-Qt.
+        if(openWithDefaultFileManager_ && !openInNewTab_) {
+            auto defaultApp = Fm::GAppInfoPtr{g_app_info_get_default_for_type("inode/directory", FALSE), false};
+            if(defaultApp != nullptr
+               && strcmp(g_app_info_get_id(defaultApp.get()), "pcmanfm-qt.desktop") != 0) {
+                for(const auto folder : folderInfos) {
+                    Fm::FileLauncher::launchWithDefaultApp(folder, ctx);
+                }
+                return true;
+            }
+        }
         mainWindow = new MainWindow(std::move(path));
         mainWindow->resize(app->settings().windowWidth(), app->settings().windowHeight());
 
--- a/pcmanfm/desktop-preferences.ui
+++ b/pcmanfm/desktop-preferences.ui
@@ -207,6 +207,17 @@ A space is also reserved for 3 lines of text.</string>
         </widget>
        </item>
        <item>
+        <widget class="QCheckBox" name="defaultFileManager">
+         <property name="toolTip">
+          <string>By default, desktop folders will be opened in PCManFM-Qt if they
+are left clicked, even when it is not the default file manager.</string>
+         </property>
+         <property name="text">
+          <string>Open desktop folders in default file manager by left clicking</string>
+         </property>
+        </widget>
+       </item>
+       <item>
         <spacer name="verticalSpacer_2">
          <property name="orientation">
           <enum>Qt::Vertical</enum>
--- a/pcmanfm/desktoppreferencesdialog.cpp
+++ b/pcmanfm/desktoppreferencesdialog.cpp
@@ -129,6 +129,8 @@ DesktopPreferencesDialog::DesktopPreferencesDialog(QWidget* parent, Qt::WindowFl
   ui.vMargin->setValue(settings.desktopCellMargins().height());
   connect(ui.lockMargins, &QAbstractButton::clicked, this, &DesktopPreferencesDialog::lockMargins);
 
+  ui.defaultFileManager->setChecked(settings.openWithDefaultFileManager());
+
   resize(sizeHint()); // show it compact
 }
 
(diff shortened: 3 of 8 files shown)


luis-pereira:
Will test.

pcmanfm/launcher.cpp
```diff
@@ -38,12 +39,26 @@ Launcher::~Launcher() {
...
     auto fi = folderInfos[0];
     Application* app = static_cast<Application*>(qApp);
     MainWindow* mainWindow = mainWindow_;
     Fm::FilePath path = fi->path();
     if(!mainWindow) {
+        // Launch folders with the default file manager if:
+        //   1. There is no main window (i.e., folders are on desktop),
+        //   2. The folders are not supposed to be opened in new tabs, and
+        //   3. The default file manager exists and is not PCManFM-Qt.
+        if(openWithDefaultFileManager_ && !openInNewTab_) {
```
luis-pereira, line 52: openWithDefaultFileManager is always false.

luis-pereira (commented):
It's not working for me.

- Setup the default file manager to something else
- Checked the use default file manager
- Left clicked on a desktop folder. It opens with pcmanfm-qt.

Traced it to openWithDefaultFileManager being always false (launcher.cpp:52). Run out of time to find the cause.

pcmanfm/launcher.cpp
tsujan: Strange! It works here. But there is a possibility that I should investigate; will tell you soon. Thanks for testing and reviewing!

tsujan: @luis-pereira 
Apart from my installation, that works, I downloaded the branch ["default_file_manager"](https://github.com/lxqt/pcmanfm-qt/tree/default_file_manager), compiled it and tested it without installation on two machines. In all cases, desktop folders were opened by Dolphin when I checked "Open desktop folders in…" and set the default file manager to Dolphin. All changes were applied on-the-fly and correctly.

I have no idea why your test resulted in a different behavior.

tsujan: At the risk of stating the obvious, OK or Apply should be pressed for changes to take effect.

stefonarch:
It works here, with thunar set to default file manager in "file associations, default applications" tab.

> Middle/right clicking triggers the default behavior.

Was it supposed to open with pcmanfm-qt? Middle click on folders does nothing here.

luis-pereira:
@tsujan I'm I doing something wrong ? https://youtu.be/lRDpWY0btVA

tsujan:
> @tsujan I'm I doing something wrong ? https://youtu.be/lRDpWY0btVA

They say a video on YouTube is worth 10 thousand words ;) In the video, You click on a desktop entry (with `Exec=pcmanfm-qt computer:///`), not on folder. 

I didn't cover desktop entries because we can't suppose all file managers can open them (for example, a file manager may not recognize `computer:///`). The patch is only for folders.

> Was it supposed to open with pcmanfm-qt? 

@stefonarch, yes, it does open pcmanfm-qt here.

luis-pereira:
@tsujan Ok. Check this one https://youtu.be/T81GpIqXHbc

tsujan:
@luis-pereira 

Again, the second item you clicked in your second video is a desktop entry with `Exec=pcmanfm-qt /home/YOUR_USER`.

There is a risk I don't want to take: If I changed the exec commands of desktop entries and removed `pcmanfm-qt` from them, this new feature could damage the default behavior (because those entries could be at the mercy of another file manager, *even without the option*).

On the other hand, making another set of desktop entries just for this option would be too much.

In my opinion, this patch is what can be done *safely* and *frugally* for a feature that I consider unimportant. I'll be happy if the parch is NOT merged (but won't be sad if it is). However, I can't agree with more that this.

tsujan:
I suggest we don't merge it and, instead, wait to see if users of LXQt ≥ 0.16.0 will need the feature in question. If it's proved to be popular, it'll be worth more changes, which could be made to the current PR.

luis-pereira:
@tsujan Thanks for your thorough explanation.
Will think about it and come back on this.

luis-pereira:
@tsujan I would be "happy" if it's merged.
For the Trash, Computer and Network use pcmanfm is right, since not all file managers support it.
But for the Home folder we could use the default file manager.  What's your opinion ?
A very rough implementation at https://github.com/lxqt/pcmanfm-qt/pull/1219.

tsujan:
@luis-pereira 
I'm happy that you work on it too (https://github.com/lxqt/pcmanfm-qt/pull/1219). IMHO, this needs more code. We need to change all shortcuts whenever the option is checked/unchecked. Also, pcmanfm-qt's desktop can be used outside LXQt (I even tested it under Wayfire — it worked fine; the problems were Wayfire's). We should also do something about `computer:///` because, for example, Dolphin doesn't understand it; etc.

Let's do it after 0.16.0; we don't have time for tests now.

tsujan:
Oh, sorry, I forgot this:

> But for the Home folder we could use the default file manager. What's your opinion ?

Yes, that's a good plan and [your PR](https://github.com/lxqt/pcmanfm-qt/pull/1219) seems to do it right; `qtxdg-mat` is the best opion. We just need to overwrite the home shortcut if the option is changed while it exists.

tsujan:
@luis-pereira 
I added a commit and squashed. It passed my tests with Dolphin, Nautilus, Thunar and spacefm. Please test it!

tsujan:
BTW, in https://github.com/lxqt/pcmanfm-qt/pull/1219#issuecomment-718246131, I said, "... because a file manager should be able to open links". I was drowsy then ;) The link is handled by `libfm-qt` and its *folder* is opened by the default file manager. Therefore, the default file manager doesn't even need to know what a link is.

luis-pereira (approved):
Tested with spacefm, thunar, dolphin and.... get this... pcmanfm-qt ;)
Works as described.

tsujan:
@luis-pereira 
Thanks! Now that we agree on the home shortcut, what do you think about merging? Should we exclude it from 0.16.0?

stefonarch:
Tested too, works fine. I wonder if "Computer" folder  should open too in the default filemanager, to have consistent behaviour with "Home" folder.

tsujan:
> Tested too

Thanks!

> I wonder if "Computer" folder should open too

Dolphin doesn't understand `computer:///` and we can't assume other file managers do. The only shortcut that all file managers can handle is home because it gives an ordinary folder to them.

stefonarch:
Clear, didn't know this. Updating translations after merge please - weblate is sending mail notifications for new string to translate :)

tsujan:
> Updating translations after merge please

I'll make PRs for all LXQt components to update changelogs and translations, starting from tomorrow. Did the same thing for 0.15.0 and will do it before all major releases.

As for merging this PR, let's wait for @luis-pereira's reply.

luis-pereira:
> As for merging this PR, let's wait for @luis-pereira's reply.

@tsujan I approved after testing it (12 hours ago). 
We can continue to work on it.But the current status is an good improvement  and works fine. Ship it.

tsujan:
> I approved after testing it (12 hours ago).

Yes, I knew that, of course. My question was about the lack of time for more tests or, in other words, "feature freeze" (https://github.com/lxqt/lxqt/issues/1848#issuecomment-714323140).

If something goes wrong, a point release is possible.

luis-pereira:
> Yes, I knew that, of course. My question was about the lack of time for more tests or, in other words, "feature freeze" ([lxqt/lxqt#1848 (comment)](https://github.com/lxqt/lxqt/issues/1848#issuecomment-714323140)).

You tested, I tested. No obvious issue.
If something is wrong we will get reports. 
 
> If something goes wrong, a point release is possible.

Of course.

tsujan:
> You tested, I tested

True. @stefonarch tested it too. My cautiousness may seem irrational but I've seen too many hidden bugs.

Anyway, we agreed on point releases in April :)

PR #1219, opened 2020-10-28
Use default file manager for Home under LXQt

luis-pereira:
It makes the user experience more coherent.

### DIFF
--- a/pcmanfm/application.h
+++ b/pcmanfm/application.h
@@ -96,6 +96,10 @@ public:
         return profileName_;
     }
 
+    bool lxqtRunning() {
+        return lxqtRunning_;
+    }
+
 protected Q_SLOTS:
     void onAboutToQuit();
     void onSigtermNotified();
@@ -153,6 +157,10 @@ private:
     char** argv_;
 };
 
+#if defined(pcmanfmApp)
+#undef pcmanfmApp
+#endif
+#define pcmanfmApp (static_cast<PCManFM::Application *>(qApp))
 }
 
 #endif // PCMANFM_APPLICATION_H
--- a/pcmanfm/desktopwindow.cpp
+++ b/pcmanfm/desktopwindow.cpp
@@ -313,7 +313,11 @@ void DesktopWindow::createTrashShortcut(int items) {
 void DesktopWindow::createHomeShortcut() {
     GKeyFile* kf = g_key_file_new();
     g_key_file_set_string(kf, "Desktop Entry", "Type", "Application");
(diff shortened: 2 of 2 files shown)


tsujan:
IMHO, the home shortcut should be changed depending on the value of `settings.openWithDefaultFileManager()`. Otherwise, its behavior will be different from that of desktop folders, which is the same problem that https://github.com/lxqt/pcmanfm-qt/pull/1216 has.

`qtxdg-mat` is a good idea under LXQt but, outside LXQt, we still have that option and need to do the job without `qtxdg-mat`. A way of doing it is changing the shortcut to a link.

luis-pereira:
> IMHO, the home shortcut should be changed depending on the value of `settings.openWithDefaultFileManager()`. Otherwise, its behavior will be different from that of desktop folders, which is the same problem that #1216 has.

I'm aware of that. That's why it's still a draft.

luis-pereira:
> `qtxdg-mat` is a good idea under LXQt but, outside LXQt, we still have that option and need to do the job without `qtxdg-mat`. A way of doing it is changing the shortcut to a link.

or use `xdg-open`.
If qtxdg-mat proves to be stable and managed to get it xdg, one single `xdg-open` would handle both cases.

tsujan:
I'm thinking about something like this under `DesktopWindow::createHomeShortcut()`:

```c++
    if(settings.openWithDefaultFileManager()) {
        g_key_file_set_string(kf, "Desktop Entry", "Type", "Link");
        g_key_file_set_string(kf, "Desktop Entry", "URL", Fm::FilePath::homeDir().toString().get());
    }
    else {
        g_key_file_set_string(kf, "Desktop Entry", "Type", "Application");
        g_key_file_set_string(kf, "Desktop Entry", "Exec", Fm::CStrPtr(g_strconcat("pcmanfm-qt ", Fm::FilePath::homeDir().toString().get(), nullptr)).get());
    }
```

It needs a little more code in another place but should work because a file manager should be able to open links.

tsujan:
@luis-pereira 

It worked fine with some changes to `topWindow::trustOurDesktopShortcut()`: the home was opened by Dolphin and the changes were applied on-the-fly.

I'll add it to https://github.com/lxqt/pcmanfm-qt/pull/1216 after more tests tomorrow; here, it's past midnight.

luis-pereira:
> I'll add it to #1216 after more tests tomorrow; here, it's past midnight.

That's great. It's a good solution.

PR #1231, opened 2020-11-12
Close tooltip on key pressing and directory/filter/sorting change

tsujan:
Closes https://github.com/lxqt/pcmanfm-qt/issues/1229


tsujan:
Because of https://github.com/lxqt/libfm-qt/commit/868f0d36979c6930efcc8be20ed355cc1205723c, a point release for libfm-qt is needed. It's better to have the same version for pcmanfm-qt and libfm-qt. So, I merge this small patch to have a meaningful change in pcmanfm-qt for its simultaneous release.

PR #1232, opened 2020-11-12
Natural keyboard navigation on Desktop

tsujan:
With this patch, the arrow keys follow the visible arrangement of items, from top to bottom and right to left. `Home`, `End`, PgDn` and `PgUp` are covered too.

Also, `Shift` adds to the selection and `Ctrl` changes the current item without changing the selection.

Closes https://github.com/lxqt/pcmanfm-qt/issues/577

### DIFF
--- a/pcmanfm/desktopwindow.cpp
+++ b/pcmanfm/desktopwindow.cpp
@@ -41,6 +41,7 @@
 #include <QClipboard>
 #include <QWindow>
 #include <QRandomGenerator>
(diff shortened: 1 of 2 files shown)


tsujan:
Merging to be tested and fixed if there is a bug.

Please test it! Also note that Qt rules for Home/End/Arrow are respected (e.g., if the first item has focus, it will NOT be selected by Home — the item with focus is always selected by Space).

PR #1246, opened 2020-11-19
translations: Use .yaml for .desktop translation

palinek:


tsujan:
@stefonarch, can this be merged? Please merge it if yes.

tsujan:
Oops! **Please don't merge it before 0.16.1!** I didn't know about the new requirements.

stefonarch:
> I didn't know about the new requirements.

What requirements? Shouldn't influence on nothing this change. But more testing in weblate doesn't hurt.

tsujan:
> What requirements?

The latest git lxqt-build-tools. I don't want to make a point release of lxqt-build-tools just for libfm-qt/pcmanfm-qt.

tsujan:
Qt 5.15.2 came here and, contrary to my expectation, libfm-qt 0.16.0 was compiled well with it. So, there is no need to a point release. If compilation fails with a future version of Qt, distro patches could be used until 0.17.0.

I removed the draft mode from the PR. Feel free to merge it (I don't know about translations).

stefonarch:
IMO all of those PR's for using .yaml files can be merged now, tests in weblate are ok.
But I prefer @palinek doing it.

PR #1255, opened 2020-12-05
cleanup *.yaml.desktop files

stefonarch:


stefonarch:
Merging to avoid weblate issues.

PR #1264, opened 2020-12-21
Don't show busy cursor before folder loading

tsujan:
Closes https://github.com/lxqt/pcmanfm-qt/issues/1263


tsujan:
Oh, this is wrong! I've explained the reason in https://github.com/lxqt/pcmanfm-qt/issues/1263#issuecomment-748702561

PR #1266, opened 2020-12-22
Changed a label and added a tooltip to Preferences

tsujan:
`Bookmarks:` is changed to `Bookmarks menu:` because it isn't about bookmarks on the side-pane.

Also added a tooltip for "Select newly created files" to inform the user that it also works with renamed files.

### DIFF
--- a/pcmanfm/preferences.ui
+++ b/pcmanfm/preferences.ui
@@ -130,7 +130,7 @@
             <item row="3" column="0">
              <widget class="QLabel" name="label_2">
               <property name="text">
-               <string>Bookmarks:</string>
+               <string>Bookmarks menu:</string>
               </property>
              </widget>
             </item>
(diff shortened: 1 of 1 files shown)


stefonarch:
I suppose this needs lxqt-transupdate, before running it we need to merge all recent  stuff from weblate and lock translations before merging updates sources I guess.

tsujan:
Before Weblate, I used `lxqt-transupdate` in all sources once in a while, usually before releases. Now, I prefer leave it to you because you know about these things much more than I do.

stefonarch:
I'd do it after merging the background image settings.

stefonarch:
Forget the above, I got confused with qterminal's bookmarks.
Merged the updated translations now and unlocked translations.
I'll keep an eye on lxqt-transupdate, it's nice to ship releases mostly translated.

tsujan:
> I'll keep an eye on lxqt-transupdate, it's nice to ship releases mostly translated.

I think it's enough to use it once a few weeks, or even a month. It's hard to keep track of new or removed strings in all LXQt components.

However, I'll tell you whenever I change a string in my PRs.

PR #1269, opened 2020-12-23
Added a Tool menu-item to open current directory as admin

tsujan:
Closes https://github.com/lxqt/pcmanfm-qt/issues/1265

### DIFF
--- a/pcmanfm/main-win.ui
+++ b/pcmanfm/main-win.ui
@@ -207,10 +207,13 @@
    </widget>
    <widget class="QMenu" name="menu_Tool">
     <property name="title">
-     <string>&amp;Tool</string>
+     <string>&amp;Tools</string>
     </property>
     <addaction name="actionOpenTerminal"/>
+    <addaction name="separator"/>
+    <addaction name="actionOpenAsAdmin"/>
     <addaction name="actionOpenAsRoot"/>
+    <addaction name="separator"/>
     <addaction name="actionCopyFullPath"/>
     <addaction name="actionFindFiles"/>
    </widget>
@@ -698,7 +701,7 @@
   </action>
   <action name="actionOpenAsRoot">
    <property name="text">
-    <string>Open as &amp;Root</string>
+    <string>Open Tab in &amp;Root Instance</string>
    </property>
   </action>
   <action name="actionEditBookmarks">
@@ -907,6 +910,11 @@
   <action name="actionHiddenShortcuts">
    <property name="text">
     <string>Hidden &amp;Shortcuts</string>
+   </property>
+  </action>
+  <action name="actionOpenAsAdmin">
+   <property name="text">
+    <string>Open Tab in &amp;Admin Mode</string>
    </property>
   </action>
  </widget>
--- a/pcmanfm/mainwindow.cpp
+++ b/pcmanfm/mainwindow.cpp
@@ -1394,6 +1394,8 @@ void MainWindow::updateUIForCurrentPage(bool setFocus) {
         ui.actionGoBack->setEnabled(tabPage->canBackward());
         ui.actionGoForward->setEnabled(tabPage->canForward());
 
+        ui.actionOpenAsAdmin->setEnabled(tabPage->path() && tabPage->path().isNative());
+
         updateViewMenuForCurrentPage();
         updateStatusBarForCurrentPage();
     }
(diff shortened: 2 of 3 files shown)


tsujan:
@stefonarch 
Please test it when you have time.

It'll need a translation update after it's merged.

stefonarch:
Should this add a new menu item? If so it's not present here, the usual 4 only. 
"Open as root" had never worked here too,   `lxsudo dbus-launch  pcmanfm-qt %U` is needed here.
At the moment I get this on a selected file or folde at "open as root" and nothing happens after authentication:

schermata-09-13-20

the command for a file should be different. But I've to check more.

tsujan:
> Should this add a new menu item? If so it's not present here, the usual 4 only.

You may have not restarted pcmanfm-qt or.... The menu item is added explicitly in the patch:

admin

> "Open as root" had never worked here too, lxsudo dbus-launch pcmanfm-qt %U is needed here.

That's explained in `Preferences → Advanced` but it isn't related to this PR. **The whole point of the PR is NOT opening a root instance.**

stefonarch:
There was something wrong with my pcmanfm-gt checkout, on a fresh clone it works, sorry.

But I don't full understand what's the difference between admin:/// and a root instance,  it does  the same but without the red  "root instance" warning?

And off topic: I never really got what this command in pref > advanced was meant for...
By  "Changing user" I didn't think of root, but of other users and didn't think further, and anyway I never use a filemanager in root mode.

GTM

tsujan:
> ... what's the difference between admin:/// and a root instance,

"Admin" isn't an instance but a mode: the app is still running as user but the user can transfer root files. 

In the admin mode, files can't be opened by clicking, unless the opening app supports admin. No other Qt app that I know of supports it but I added a simple support to git FeatherPad.

Under Wayland, only the admin mode works; a root instance is impossible with Wayland.

tsujan:
> And off topic: I never really got what this command in pref > advanced was meant for.

To open a root instance. It works here. But if doesn't work for you, "Discussions" will be a good place to discuss it.

stefonarch (approved):

stefonarch:
Somehow related: maybe a general rewording is needed as some items refer to the folder and only "copy full path" to a selected file or folder

```
Tool > Tools
Open terminal here
Open position in admin mode
Open position as root
```


schermata-26-12-11:38

tsujan:
Yes, the texts could be better:

 * Since the admin mode is for the current directory, its text could be "Open in Admin Mode".
 * The text for root could be "Open in Root Instance".

In this way, the user knows that the admin mode is inside the current window while the root is about another instance of the app.

"Position" is a misleading word, IMO. The current folder is opened in both cases; so, there's no need to use longer texts like "Open Current Folder...."

@stefonarch. what do you think?

stefonarch:
The problem IMO is that it's not clear if those items are about selected files or not - "copy path" and "find" work on selections too, the others not.
Not sure if you saw my edit about moving copy-path to context menu.

stefonarch:
Thinking of it also "find files" isn't correct, as it searches also inside files. Just "Search"?

tsujan:
> The problem IMO is that it's not clear if those items are about selected files or not - "copy path" and "find" work on selections too, the others not.

You're completely right; I hadn't thought about that.

So, what about "Open Tab in Admin Mode" and "Open Tab in Root Instance"?

tsujan:
> Thinking of it also "find files" isn't correct, as it searches also inside files. Just "Search"?

I agree. Will change it in this PR. Thanks!

stefonarch:
> So, what about "Open Tab in Admin Mode" and "Open Tab in Root Instance"?

That's clear and short.

>Will change it in this PR!

Remember also "Tool > Tools" ;)

stefonarch:
Noticed  only now: open terminal/F4 is broken here now, please check.

tsujan:
> Remember also "Tool > Tools" ;)

Good point! Yes, it should be "Tools". Thanks!

tsujan:
@stefonarch 

Done.

Only "Find Files" was OK and I didn't change it because, in all cases, files are found — even when the search is done in their contents.

stefonarch:
Looks fine all now!
The admin mode will be new to many people, saw just now that help > about is quite empty...

tsujan:
> saw just now that help > about is quite empty...

See https://github.com/lxqt/pcmanfm-qt/issues/550. Never found time to write it but it's very very needed. pcmanfm-qt is an advanced file manager but many users may not know about custom actions and several other things — I can't use a file manager without them.

PR #1270, opened 2020-12-24
shortened text

stefonarch:
In many translations this text is way too long and shortens selectors in GUi.
It's obvious that 0 =disabled.
Added missing colon.

### DIFF
--- a/pcmanfm/preferences.ui
+++ b/pcmanfm/preferences.ui
@@ -99,8 +99,11 @@
               <property name="enabled">
                <bool>false</bool>
               </property>
+              <property name="toolTip">
+               <string>Set to zero to disable auto-selection.</string>
+              </property>
               <property name="text">
-               <string>Delay of auto-selection in single click mode (0 to disable)</string>
+               <string>Delay of auto-selection in single click mode:</string>
               </property>
              </widget>
             </item>
@@ -124,6 +127,9 @@
               </property>
               <property name="singleStep">
                <double>0.100000000000000</double>
+              </property>
+              <property name="toolTip">
+               <string>Set to zero to disable auto-selection.</string>
               </property>
              </widget>
             </item>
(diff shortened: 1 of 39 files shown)


tsujan:
> It's obvious that 0 =disabled.

I don't think so. It wasn't obvious to me when I came to LXQt and I was happy that it could be disabled in that way.

IMO, we shouldn't sacrifice clarity for the width of a dialog ;)

tsujan:
With the shorter text, please add this tooltip to it:
```
Set it to zero to disable auto-selection
```

stefonarch:
Thinking about it maybe you're right: it isn't that obvious.
But as it's greyed out if single click  mode is not selected the text could be:
`Delay of auto-selection mode (0 to disable):`

tsujan:
> But as it's greyed out if single click mode is not selected

Oh, I think we have a bug here! If "single click" is unchecked, the DE's setting will be used (as its tooltip says) but it can be single-click again; therefore, the spin-box and its text shouldn't be disabled.

I'll check the code soon but, because of the above, "single click mode" shouldn't be omitted. Please use a tooltip ("Set it to zero to disable auto-selection")!

stefonarch:
Setting delay >0 works also if greyed out. 
But in mouse setting > use single click: I wonder if there is there anything besides files that use double clicks?

tsujan:
> But in mouse setting > use single click: I wonder if there is there anything besides files that use double clicks?

There is, in Qt's item-views and if their click behavior isn't set by their apps.

tsujan:
Check Configuration Center as an example; we haven't overridden its click behavior.

tsujan (approved):
GTM

@stefonarch, please merge it if you don't want to change translations. The GUI changes are fine.

stefonarch:
Thinking about I like much better the text:
`Delay of auto-selection (0 to disable):`
tooltip: `Delay of auto-selection in single click mode `

tsujan:
@stefonarch, the current texts (for which I added a "GTM") are clear and professional: the main text shows user what the option is for and the tooltip gives him additional info about it.

I know that your goal is shortening the text but please don't sacrifice usability!

stefonarch:
ok, but I have to update it first otherwise it's likely that weblate will have rebase errors.

tsujan:
Yes, I'd guessed that you might want to update translations. That's why I didn't merge it myself.

PR #1274, opened 2020-12-27
Added "invert selection" shortcut

stefonarch:
There is no easy way to unselect after ctrl+A using the  keyboard, I was missing that.
ctrl+shift+a inverts selection.
Could be also another instead of this one.

### DIFF
--- a/pcmanfm/main-win.ui
+++ b/pcmanfm/main-win.ui
@@ -617,6 +617,9 @@
    <property name="text">
     <string>&amp;Invert Selection</string>
    </property>
+   <property name="shortcut">
+    <string>Ctrl+Shift+A</string>
+   </property>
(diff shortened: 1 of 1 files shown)


tsujan:
GTM

tsujan:
> Could be also another instead of this one.

It's usually `Ctrl+I` but it's used by the filter bar (and that was the reason why "Invert Selection" had no shortcut) but `Ctrl+Shift+A` is acceptable.

stefonarch:
I see that on the desktop this has no effect (ctrl+A is working). But no big thing.

tsujan:
Menu shortcuts are for windows. Desktop defines its own shortcuts in its code and no, it has no shortcut for selection inversion.

PR #1303, opened 2021-02-03
Completed the support for creation time

tsujan:
Follows and depends on https://github.com/lxqt/libfm-qt/pull/627

NOTE: The settings for the hidden columns of the detailed list mode may change (because a new column is added). If you have hidden some columns, you might need to change them after this patch.


tsujan:
Oh, of course, GitHub's checks for this PR fail because it depends on a libfm-qt PR (I received 2 emails about that).

PR #1305, opened 2021-02-10
Workaround an issue with glibc 2.33 on old Docker engines

yan12125:
The same as https://github.com/lxqt/lxqt-panel/pull/1562


yan12125:
Merging myself as this patch only touches CI.

PR #1307, opened 2021-02-13
Handle GVFS' Admin separately and appropriately

tsujan:
Since Admin is peculiar (e.g., its mount isn't visible and it asks for the password after being mounted), it's treated separately for preventing another Polkit prompt after the first one is cancelled.

The patch mostly consists of long comments because, although its code is very simple, the logic behind it is not.

I also found and worked around a bug in the GVFS admin backend, which could cause an infinite loop in the old code (without the current patch and in a very rare case). Unfortunately, our old code can still be vulnerable to upstream bugs but that's a separate topic.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1306


tsujan:
@stefonarch 
The patch works fine here but would you please test too?

It should cover all cases, e.g., also when `admin:///` is *directly* entered into the path-bar or an Admin bookmark is opened after pcmanfm-qt's process is started: no double prompt will be shown if the first prompt is cancelled.

tsujan:
Any objection to merging this?

PR #1326, opened 2021-03-02
Made max thumbnail size spinbox in MiB

tsujan:
And set its max to 2048 MiB, with a precision of 0.1 MiB.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1316


tsujan:
I close this to make another PR for having separate options for built-in and external thumbnailers.

PR #1328, opened 2021-03-03
Added an option for max. file size with external thumbnailers

tsujan:
Follows and depends on https://github.com/lxqt/libfm-qt/pull/644

The external thumbnailers shouldn't follow the size limit of our built-in image thumbnailer. We allocate an amount of memeory equal to the file size when making internal thumbnails of images but an external thumbnailer can easily create the thumbnail of a 1-GiB video. In other words, images and videos shouldn't have the same size limit.

Moreover, this patch is backward-compatible: users won't see a difference with external thumbnails (of videos, PDF,…) and the behavior of internal thumbnails will be mostly unchanged.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1316

### DIFF
--- a/pcmanfm/preferences.ui
+++ b/pcmanfm/preferences.ui
@@ -122,14 +122,14 @@
               <property name="enabled">
                <bool>false</bool>
               </property>
+              <property name="toolTip">
+               <string>Set to zero to disable auto-selection.</string>
+              </property>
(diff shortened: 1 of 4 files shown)


tsujan:
GitHub's check will fail at first, because of a dependency on a libfm-qt PR.

PR #1333, opened 2021-03-13
updated *.ts files

stefonarch:


tsujan (approved):
GTM.

Merge it whenever you see fit -- I didn't only because I didn't know what would happen in Weblate.

stefonarch:
Before running lxqt-transupdatein weblate pending commits are pushed, merged and then I lock the component.

PR #1337, opened 2021-03-25
Port to Qt 5.15 and slow down compiler warnings

adjamhub:
This should be last patch in the "port to Qt 5.15" and "move to Qt5 signal/slot syntax" series :)

### DIFF
--- a/pcmanfm/settings.cpp
+++ b/pcmanfm/settings.cpp
@@ -448,8 +448,12 @@ bool Settings::saveFile(QString filePath) {
     settings.endGroup();
 
     settings.beginGroup(QStringLiteral("Places"));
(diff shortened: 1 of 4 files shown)


pcmanfm/settings.cpp
```diff
@@ -443,7 +448,7 @@ bool Settings::saveFile(QString filePath) {
     settings.endGroup();
 
     settings.beginGroup(QStringLiteral("Places"));
-    QStringList hiddenPlaces = hiddenPlaces_.toList();
+    QStringList hiddenPlaces = hiddenPlaces_.values();
```
tsujan, line 451: Please do here what you did at https://github.com/lxqt/libfm-qt/pull/659

tsujan (approved):
Thanks!

PR #1343, opened 2021-04-05
Pre-release changes

tsujan:


tsujan:
CI failed where it shouldn't. It can't see libfm-qt 0.17.0, which was updated minutes ago.

PR #1344, opened 2021-04-06
updated *.ts files

stefonarch:


stefonarch:
Should be ok but better double check.

tsujan (approved):
It's perfect.

PR #1356, opened 2021-04-22
Character removal in translation

bugcrazy:
Character entered incorrectly in translation.


stefonarch:
All translations should be done[ using  weblate](https://github.com/lxqt/lxqt/wiki/Translation):
https://translate.lxqt-project.org/translate/lxqt-desktop/pcmanfm-qt/pt_BR/?checksum=69b928c62a27eb11&q=%25Ferramentas&sort_by=-priority%2Cposition

And it's not to remove, it's a tipo, should be `&Ferramentas`

Thanks for reporting!

PR #1361, opened 2021-04-26
Better text colors in inactive frame of split view

tsujan:
In rare cases, the colors of view text and window text may be very different from each other (see https://github.com/lxqt/lxqt-qtplugin/issues/69). So, they're distinguished from each other in the inactive frame of the split view. In most cases, there will be no difference.


tsujan:
I'm merging this; it'll be needed by users who set strange palettes after https://github.com/lxqt/lxqt-qtplugin/commit/0ae303ae2a32d308d84f303978b5b4cfd0dd7dcd

PR #1365, opened 2021-04-28
Move to Qt5.15 as Q_MINIMUM_VERSION

adjamhub:
also, 2 small C++ fixes: one missing ovverride clause and one reference
used to not copy objects


tsujan (approved):
GTM (will read it again later).

PR #1373, opened 2021-05-10
Fixed inconsistent item selection with transient filter-bar

tsujan:
Since the transient filter-bar replaces Qt's auto-selection when some text is typed inside the view, the patch ensures that the first item is always selected after a filtering of this kind.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1372


retrotails:
I think one negative change this brought is selecting items based on the first letters. for example I have folders "ab" and "bb", if I type "b" I expect "bb" to be selected because it begins with what I'm typing, which was the old behavior. but now it will select "ab" first because it contains the letter "b".

PR #1381, opened 2021-05-23
Allow recursive customization of folders

tsujan:
Closes https://github.com/lxqt/pcmanfm-qt/issues/1056

### DIFF
--- a/pcmanfm/main-win.ui
+++ b/pcmanfm/main-win.ui
@@ -149,6 +149,14 @@
      <addaction name="actionShowFilter"/>
      <addaction name="actionUnfilter"/>
     </widget>
+    <widget class="QMenu" name="menuView_Settings">
+     <property name="title">
+      <string>Customized View Settings</string>
+     </property>
+     <addaction name="actionPreserveView"/>
+     <addaction name="actionPreserveViewRecursive"/>
+     <addaction name="actionGoToCustomizedViewSource"/>
+    </widget>
     <addaction name="actionReload"/>
     <addaction name="separator"/>
     <addaction name="actionShowHidden"/>
@@ -156,7 +164,7 @@
     <addaction name="actionSplitView"/>
     <addaction name="menuSorting"/>
     <addaction name="menu_View_2"/>
-    <addaction name="actionPreserveView"/>
+    <addaction name="menuView_Settings"/>
     <addaction name="separator"/>
     <addaction name="actionSidePane"/>
     <addaction name="menuToolbars"/>
@@ -573,14 +581,6 @@
     <string>&amp;Hidden Last</string>
    </property>
   </action>
-  <action name="actionPreserveSorting">
-   <property name="checkable">
-    <bool>true</bool>
-   </property>
-   <property name="text">
-    <string>&amp;Preserve sorting for this folder</string>
-   </property>
-  </action>
   <action name="actionNewTab">
    <property name="icon">
     <iconset theme="window-new">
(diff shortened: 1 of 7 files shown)


tsujan:
@stefonarch
I made the PR as a draft for you and I to test it and find its probable problems. There's absolutely no rush.

Please also check if there's a tangible difference in CPU usage when you change deep directories quickly. I saw none with my usual directories.

stefonarch:
Thanks, I'm testing it on both machines.
* Couldn't find delays or exceptional CPU usage also on the old laptop is (which I consider on the limit uf usability nowadays). Made a hierarchy of >190 folders with 5 files each, clicking trough it is smooth and CPU goes up max to 10% on the old machine.
* A folder which already has set some customized view will not follow the recursive setting
* Maybe the existing text should be changed from ` Preserve sorting for this folder` to `Preserve settings for this folder`, downside: will break existing translations which could be correct. Also texts should be similar , not sure about "recursively" which can be hard to translate. "Preserve setting for this folder and subfolders"?
* not really related but looking at `~/.config/pcmanfm-qt/lxqt/dir-settings.conf` I noticed may "dead" entries, the file has ~500 lines and is 8k. Could it be too cluttered over time and slow down checks? Some cleaning option is maybe needed: checking for non-existing folders (except /media and /var/run/media)?
* My initial idea to have all thumbnail view down in ~/foto wasn't quite sharp... 
I'd need "list view" for folders with only subfolders and "thumbnail view" for the folders containing mostly or only pictures, but that's asking for the moon I guess ;)

Atm that's all, I'll edit when I find something else.

tsujan:
> Couldn't find delays or exceptional CPU usage also on the old laptop

That's very good news :)

> A folder which already has set some customized view will not follow the recursive setting

Correct, because the recursive setting is "false" by default. If you want it, you should toggle it by checking "Preserve Settings Recursively from Here".

> not really related but looking at ~/.config/pcmanfm-qt/lxqt/dir-settings.conf I noticed may "dead" entries,

Correct, because pcmanfm-qt doesn't clean it up on exiting. According to PCMan, it wasn't a problem (I remember he mentioned that MS Windows also had such files) but, since I didn't like it at all, I made a simple patch years ago (which I use without problem). For some reason, that I don''t remember, my patch was forgotten. I'll make a PR for it later.

> I'd need "list view" for folders with only subfolders and "thumbnail view" for the folders containing mostly or only pictures

That's impossible in the way folder settings are handled by libfm-qt. It would need a totally different approach, which is beyond this PR.

tsujan:
I opened https://github.com/lxqt/pcmanfm-qt/issues/1382 for dead entries.

tsujan:
I missed this in my previous comment:

> not sure about "recursively" which can be hard to translate

Translators will find a way in their native languages. A translation doesn't need to be made word by word; it should only convey the meaning.

>  "Preserve setting for this folder and subfolders"?

Too long for a menu-item, IMHO.

tsujan:
I also fixed an old problem (the view mode was never restored on removing customization) and removed the "draft" tag. Let's use it for a while...

tsujan:
@stefonarch, is the following idea acceptable to you?

Currently, the user has only one way to know if a folder is customized or not: he should take a look at "Customized View Settings".

So, I thought about adding tab icons. When a folder is customized — whether directly or by inheriting the customization from a parent folder — its tab can have the icon of its view mode, and If there's no customization, there will be no tab icon, like this screenshot (taken with Fusion):

tab_icons

stefonarch:
To me it looks like a nice feature!

tsujan:
Added a commit for it to this PR.

https://github.com/lxqt/pcmanfm-qt/issues/1382 will need a separate PR, after the current PR is thoroughly tested and merged. So, I think this PR won't change anymore (unless for fixing probable issues).

tsujan:
Tested enough. Merging for git users.

PR #1386, opened 2021-06-03
Update old entries of PCManFM to "-qt"

iamthesenate1:
While translating I spotted two lines that still had PCManFM without "-qt".
I think this behaviour was not intentional, but if it was, correct me.


tsujan:
Thanks! Yes, it should have been PCManFM-Qt.

But please remove the changes you made to `pcmanfm/translations/pcmanfm-qt.ts` — it'll be done by Weblate.

iamthesenate1:
> Thanks! Yes, it should have been PCManFM-Qt.
> 
> But please remove the changes you made to `pcmanfm/translations/pcmanfm-qt.ts` — it'll be done by Weblate.

Done!

PR #1389, opened 2021-06-08
Added an action to remove settings of nonexistent folders

tsujan:
A prompt dialog is shown that tells the user why those settings might be useful later.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1382

### DIFF
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -375,6 +375,31 @@ void Application::onAboutToQuit() {
     settings_.save();
 }
 
+void Application::cleanPerFolderConfig() {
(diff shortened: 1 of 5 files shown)


stefonarch:
GTM, tested and working fine.

tsujan:
Thanks!

> Good that always a backup is created.

Yes, it's created by `libfm-qt`, through Glib/GIO, whenever the settings are saved.

stefonarch:
I should've waited running` lxqt-transupdate` ... forgot about this.

tsujan:
> I should've waited running lxqt-transupdate ... forgot about this.

It isn't predictable. Running it once in a while may be the only way.

PR #1390, opened 2021-06-10
Toggle visibility of dir tree's hidden folders from view's context menu

tsujan:
Previously, the visibility of hidden folders of side-pane's directory tree was toggled only by the View menu or its shortcut (Ctrl+H).

Fixes https://github.com/lxqt/pcmanfm-qt/issues/1383


stefonarch:
Working fine.

PR #1407, opened 2021-06-25
Also allow creating of launchers with slash in their names

tsujan:
Not important but when it's possible, why not?


tsujan:
Straightforward. Merging...

PR #1408, opened 2021-06-26
Added option to make desktop items sticky by default

tsujan:
Closes https://github.com/lxqt/pcmanfm-qt/issues/1392


tsujan:
@stefonarch 
Would please test it when you find the time? Here, it has worked as expected for weeks.

tsujan:
OK, it has worked here without problem for a month. Making git users test it...

Issue #728, opened 2018-07-13
Not a bug: About moving desktop items


tsujan:
Before adding a workaround for a Qt bug -- with old pcmanfm-qt -- and now, after https://github.com/lxqt/pcmanfm-qt/pull/723 , desktop items can be put between each other by DND .

However, drop positions are really cells in the desktop grid. To put an item into a cell, it should be dropped when the cursor is inside the top left quarter of the rectangular cell. If it's dropped on to top right quarter, it'll stick to the right cell; if dropped on the left bottom quarter, it'll go to the bottom cell; with the right bottom quarter, it'll go to the cell diagonally positioned at the right bottom.

This behavior is a little troublesome or counterintuitive and can be changed. However, when it comes to dropping between 2 adjacent *folders*, I think it's good.

Ideas about this are welcome.

Please note that you should have https://github.com/lxqt/pcmanfm-qt/pull/723 to be able to put items between each other. However, what's described above can be seen with empty cell too.

tsujan:
OK, I dragged and dropped desktop items tens of times and feel/think that it's more natural to drop items into the cell behind the mouse cursor, regardless of the cursor position relative to the cell. Will add this to https://github.com/lxqt/pcmanfm-qt/pull/723

tsujan:
Done in https://github.com/lxqt/pcmanfm-qt/pull/723/commits/a7898c9dede712b2130e94820f8c0e053e99276d

IMO, DND feels natural on desktop now.

tsimonq2:
I see one more problem after just sitting and messing with it for a while.

When I drag an icon to the lower part of another icon, it should move the icon I'm hovering over up, not down.

tsujan:
> When I drag an icon to the lower part of another icon, it should move the icon I'm hovering over up, not down.

Thanks for your attention! I'm glad you mentioned that. Yes, I wanted to do so at first but saw that it was too complicated for the code -- almost an overkill -- because then, I should consider the right and the left too (dropping on the right half, for example, should move the item to the left). Moreover, moving up may not be possible but moving down (and to the right, if needed) is always possible because the grid starts from the top left corner. That makes the implementation much more complex.

I'd kept this idea to myself to see if I could reduce its algorithmic complexity ;) No luck so far.

PR #1440, opened 2021-09-07
close floating windows when pcmanfm-qt primary instance quits

antis81:
This PR depends on [libfm-qt #728](https://github.com/lxqt/libfm-qt/pull/728) and closes potential floating windows pointers before the pcmanfm-qt primary instance quits.


antis81:
I have investigated it further and PCMan's MainWindow is an actual QMainWindow always unparented!

kwin_pcmanfm_mainwindow_bug

PR #1445, opened 2021-09-13
Added and changed accelerators for view settings

tsujan:
An accelerator is changed because it could become ambiguous in tab's right-click menu. Missing accelerators are also added.

@stefonarch, this might remove some translated strings. Please test and merge it if you see no problem.

### DIFF
--- a/pcmanfm/main-win.ui
+++ b/pcmanfm/main-win.ui
@@ -151,7 +151,7 @@
     </widget>
     <widget class="QMenu" name="menuView_Settings">
      <property name="title">
-      <string>Customized View Settings</string>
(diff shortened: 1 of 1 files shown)


stefonarch (approved):
GTM. But merging only after actual rebase issue is solved... then a transupdate.

tsujan:
Accelerators can make translation difficult. The translator needs to see the GUI and avoid two instances of the same accelerator in a single menu. But I guess there's no way around it.

stefonarch:
I'm afraid most do like me: just choose some random character pertinent (C in "Change") and no control if it's already used, because you can cycle trough if there are more with the same character.

PR #1463, opened 2021-10-19
Never close file operation dialog on closing window

tsujan:
It was closed when pasting from Edit menu (or by `Ctrl+V`) because the destination window was its parent.

Fixes https://github.com/lxqt/pcmanfm-qt/issues/1462


tsujan:
Straightforward. Merging...

PR #1485, opened 2021-11-12
Added "Recent Files" to File menu

tsujan:
It covers non-folder, native files that are launched (non-native files could cause trouble). The number of recent files can be set from 0 (the default) to 50.

On exiting the current session, the list of recent files is saved to a separate config file inside pcmanfm-qt's user config directory. But, if the user clears the list, the config file will be emptied immediately, for the sake of privacy.

NOTE1: This needs https://github.com/lxqt/libfm-qt/commit/6b8d5acd76e7c7eb7707257b006995b17231e72f
NOTE2: It is independent of GVFS' `recent:///`, which is useless to us.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1483

### DIFF
--- a/pcmanfm/launcher.cpp
+++ b/pcmanfm/launcher.cpp
@@ -19,6 +19,7 @@
  */
 
 #include "launcher.h"
+#include <QFileInfo>
 #include "mainwindow.h"
 #include "application.h"
 #include <libfm-qt/core/filepath.h>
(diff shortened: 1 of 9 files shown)


tsujan:
OK, I must say that I'm disappointed with the integrated checks. This time, I waited for 3 hours after I merged https://github.com/lxqt/libfm-qt/pull/749 but still, I got a false positive.

tsujan:
Tested by @stefonarch and me. Merging for more tests by git users.

PR #1495, opened 2021-11-25
Fixed DND on RTL desktop

tsujan:
The code is minimal. The patch mostly consists of explanatory comments on the difference between right and left, which wasn't so obvious in this case ;)

Fixes https://github.com/lxqt/pcmanfm-qt/issues/1494

### DIFF
--- a/pcmanfm/desktopwindow.cpp
+++ b/pcmanfm/desktopwindow.cpp
@@ -1300,6 +1300,12 @@ void DesktopWindow::relayoutItems() {
 
     QRect workArea = screen->availableVirtualGeometry();
     workArea.adjust(WORK_AREA_MARGIN, WORK_AREA_MARGIN, -WORK_AREA_MARGIN, -WORK_AREA_MARGIN);
+
+    // switch between right and left with RTL to use the usual LTR calculations
+    if(layoutDirection() == Qt::RightToLeft) {
+        workArea.moveLeft(rect().right() - workArea.right());
+    }
+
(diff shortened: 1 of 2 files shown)


stefonarch:
I think you saw that wile compiling, to me it's arabic as italians say:

```
/home/stef/git/lxqt/pcmanfm-qt/pcmanfm/autorundialog.cpp: In destructor ‘virtual PCManFM::AutoRunDialog::~AutoRunDialog()’:
/home/stef/git/lxqt/pcmanfm-qt/pcmanfm/autorundialog.cpp:53:34: warning: cast between incompatible function types from ‘void (*)(gpointer)’ {aka ‘void (*)(void*)’} to ‘GFunc’ {aka ‘void (*)(void*, void*)’} [-Wcast-function-type]
   53 |     g_list_foreach(applications, (GFunc)g_object_unref, nullptr);

```

tsujan:
The compilation warning is a very old GLib warning and can be ignored. In happens with libfm-qt too, alongside a few others. At some point, I'll silence them.

Your second attachment seems to be triggered by lximage-qt but it doesn't seem to be *about* lximage-qt either.

As for "qt.qpa.xcb: QXcbConnection: XCB error: 3 (BadWindow)...", it's also a very old message that can be found everywhere Qt is used under X11. To silence worthless messages like it, you could create `~/.config/QtProject/qtlogging.ini` as:

```
[Rules]
*.warning=false
```

PR #1500, opened 2021-12-01
fix: The project fails to compile in the debian qt5.15 environment.

GongHeng2017:
In launcher.h, fuction "launchedFiles()" and "launchedPaths()" are marked "override",but the parent
don't have the two function.
I also find, the two function  don't be used.
error


tsujan:
The latest git libfm-qt is needed by the latest git pcmanfm-qt.

PR #1512, opened 2022-01-01
Spelling: Ctrl+, media type

comradekingu:
Consistent with https://translate.lxqt-project.org/translate/lxqt-desktop/lximage-qt/sv/?checksum=27f72b4b6a7dea5c (Ctrl+M)

### DIFF
--- a/pcmanfm/preferences.ui
+++ b/pcmanfm/preferences.ui
@@ -162,7 +162,7 @@
             <item row="2" column="0" colspan="2">
              <widget class="QCheckBox" name="ctrlRightClick">
               <property name="text">
-               <string>Show folder context menu with Ctrl + right click</string>
+               <string>Show folder context menu with Ctrl+right click</string>
               </property>
              </widget>
             </item>
@@ -899,7 +899,7 @@ A value of -1 means that there is no limit for the file size (the default).</str
             <item>
              <widget class="QCheckBox" name="templateTypeOnce">
               <property name="text">
-               <string>Show only one template for each MIME type</string>
+               <string>Show only one template for each media type</string>
               </property>
              </widget>
             </item>


pcmanfm/preferences.ui
```diff
@@ -899,7 +899,7 @@ A value of -1 means that there is no limit for the file size (the default).</str
             <item>
              <widget class="QCheckBox" name="templateTypeOnce">
               <property name="text">
-               <string>Show only one template for each MIME type</string>
+               <string>Show only one template for each media type</string>
```
tsujan, line 902: This isn't about media but mime.

pcmanfm/preferences.ui
```diff
@@ -162,7 +162,7 @@
             <item row="2" column="0" colspan="2">
              <widget class="QCheckBox" name="ctrlRightClick">
               <property name="text">
-               <string>Show folder context menu with Ctrl + right click</string>
+               <string>Show folder context menu with Ctrl+right click</string>
```
tsujan, line 165: Unnecessary changes to strings break translations. " + " is as valid as "+".

comradekingu: What it does is fuzzify translations, and alerting translators to how it could be more consistent is in this case warranted.

pcmanfm/preferences.ui
comradekingu: Unless https://en.wikipedia.org/wiki/MIME then that is all https://en.wikipedia.org/wiki/Media_type

tsujan: → https://www.freedesktop.org/wiki/Specifications/shared-mime-info-spec/

comradekingu: https://www.iana.org/assignments/media-types/media-types.xhtml am I missing something?

pcmanfm/preferences.ui
tsujan: "Right click" is a compound word with a space in the middle. We don't have a compound word like "Ctrl+right click".

pcmanfm/preferences.ui
tsujan: > am I missing something?

Yes, `freedesktop.org`, which we follow.

comradekingu: So it becomes valid here once freedesktop.org updates their wording?

pcmanfm/preferences.ui
comradekingu: Could you rephrase that? How I understood it doesn't appear valid to me.

pcmanfm/preferences.ui
tsujan: There's nothing to update. "image/jpeg is an example of a MIME type where image is the media type, and jpeg is the subtype identifier." (→ https://help.gnome.org/admin/system-admin-guide/stable/mime-types.html.en)

pcmanfm/preferences.ui
tsujan: > Could you rephrase that?

"2 red apples" vs. "2red apples"

pcmanfm/preferences.ui
comradekingu: Yes, but from my understanding the word to reference the concept changed from "MIME type" to "media type" a while ago, regardless of underlying components.
https://www.iana.org/assignments/media-types/media-types.xhtml was updated 8 days ago.

There is a "top-level media type" and "media subtype" in https://www.rfc-editor.org/rfc/rfc2046.html from 1996

pcmanfm/preferences.ui
comradekingu: So you are saying it should be "Ctrl+ right click"?

PR #1521, opened 2022-01-23
Changed shortcut for bulk rename

stefonarch:
Never worked. Ctrl+F2 is already used in lxqt-globalkeys for "Switch to Desktop 2", same in ` ~/.config/kglobalshortcutsrc`


tsujan:
I have no problem with it. Please just make sure that `Shift+F2` isn't used elsewhere, in pcmanfm-qt itself, and if not, merge it. Thanks!

PR #1524, opened 2022-02-03
Added "Open in Terminal" to top of folder menu

tsujan:
It had been suggested but I didn't do it to keep the folder menu clean.

However, adding a single item doesn't clutter the menu (provided that we don't do it again) and it's especially useful with Desktop.

### DIFF
--- a/pcmanfm/view.cpp
+++ b/pcmanfm/view.cpp
@@ -146,7 +146,17 @@ void View::prepareFileMenu(Fm::FileMenu* menu) {
     }
 }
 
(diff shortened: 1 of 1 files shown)


stefonarch:
Never used this menu for terminal - always F4 but yes, on the desktop F4  doesn't work.
 GTM

tsujan:
> Never used this menu for terminal - always F4

Me too.

While adding it, I found an ancient bug in libfm-qt, which almost no one would encounter. Will fix it soon.

PR #1525, opened 2022-02-05
Implementing org.freedesktop.FileManager1

### LINKED ISSUE #1194: [Feature request] Implement org.freedesktop.FileManager1 DBus interface
##### Expected Behavior

When I select "show in folder" for a download in a browser, the Downloads folder should be opened and the downloaded file should be selected.

##### Current Behavior

When I select "show in folder" for a download in a browser, then only the Downloads folder opens but I have to search for the downloaded file manually, which can be quite annoying if I don't sort by date.

##### Possible Solution

Looks like GNOME and KDE use https://www.freedesktop.org/wiki/Specifications/file-manager-interface/, `org.freedesktop.FileManager1` DBus name and the `/org/freedesktop/FileManager1` object path. 

Looking at https://github.com/linuxdeepin/dde-file-manager/search?q=dbusfilemanager1 the implementation seems to be relatively straightforward.

And e.g., Chromium implements it:
https://chromium.googlesource.com/chromium/src/+/master/chrome/browser/platform_util_linux.cc

Since pcmanfm-qt is all about XDG standards, I'd ask for consideration to implement this feature.

##### Context

My downloads folder is always full and it is annoying that pcmanfm-qt cannot do what other file managers do in this regard.

ahsand97:
This PR closes #1194 and implements the three methods of [org.freedesktop.FileManager1](https://www.freedesktop.org/wiki/Specifications/file-manager-interface/) specification.

It is based on [Filer](https://github.com/helloSystem/Filer/) (a fork of PCManFM) [implementation](https://github.com/helloSystem/Filer/commit/191826b3f46870288ce69f8ee3ad365845c526f8) 

Behavior: 

**ShowFolders**
`dbus-send --session --print-reply --dest=org.freedesktop.FileManager1 /org/freedesktop/FileManager1 org.freedesktop.FileManager1.ShowFolders array:string:"file:///home,file:///etc" string:""`

**Result:** It will add a new tab showing the content of "/home" and another one showing the content of "/etc", those tabs are going to be in a new window (if there's none or if the setting to use single window mode is disabled).

**ShowItems**
`dbus-send --session --print-reply --dest=org.freedesktop.FileManager1 /org/freedesktop/FileManager1 org.freedesktop.FileManager1.ShowItems array:string:"file:///home,file:///etc/hosts" string:""`

**Result:** It will add a new tab showing the content of "/" highlighting the folder "home" and another tab showing the content of "/etc" highlighting the file "hosts", those tabs are going to be in a new window (if there's none or if the setting to use single window mode is disabled).

**ShowItemProperties**
`dbus-send --session --print-reply --dest=org.freedesktop.FileManager1 /org/freedesktop/FileManager1 org.freedesktop.FileManager1.ShowItemProperties array:string:"file:///home,file:///etc/hosts" string:""`

**Result:** It will open two property dialogs, one for the folder "/home" and another one for the file "/etc/hosts".


tsujan:
Thanks for your contribution and for trying to make this app better!

Since this is a large patch and because my time is taken by coding, I'll review it part by part, whenever I find the time.

pcmanfm/application.cpp
```diff
@@ -48,6 +48,7 @@
 #include "applicationadaptor.h"
 #include "preferencesdialog.h"
 #include "desktoppreferencesdialog.h"
+#include <libfm-qt/filepropsdialog.h>
```
tsujan, line 51: Please put it beside other `libfm-qt` headers.

ahsand97: got it!

pcmanfm/application.cpp
```diff
@@ -690,6 +691,119 @@ void Application::setWallpaper(QString path, QString modeString) {
     }
 }
 
+/* This method receives a list of file:// URIs from DBus and for each URI opens
+ * a tab showing its content.
+ */
+void Application::ShowFolders(const QStringList uriList, const QString startupId __attribute__((unused))) {
```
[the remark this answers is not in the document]
ahsand97: > 

Got it!

pcmanfm/tabpage.cpp
```diff
@@ -465,7 +465,12 @@ void TabPage::onFolderFinishLoading() {
     // After finishing loading the folder, the model is updated, but Qt delays the UI update
     // for performance reasons. Therefore at this point the UI is not up to date.
     // For example, the scrollbar ranges are not updated yet. We solve this by installing an Qt timeout handler.
-    QTimer::singleShot(10, this, &TabPage::onUiUpdated);
+    QTimer::singleShot(10, this, [this] {
+        onUiUpdated();
+        // When the UI is fully loaded if there are files to select we select them
+        if(!filesToSelect_.isEmpty())
+            onSelectItems();
+    });
```
tsujan, line 473: There's no need to this change and to the new method `onSelectItems()`. Instead, you could add the contents of `onSelectItems()` to the end of `onUiUpdated()` (with an appropriate comment) and remove it. I'll review the contents later (it seems you've changed it).

pcmanfm/tabpage.cpp
```diff
@@ -1078,4 +1081,24 @@ bool TabPage::canOpenAdmin() {
...
+void TabPage::onSelectItems() {
+    Fm::FileInfoList filesToSelect;
+    for(QString item : filesToSelect_) {
+        Fm::FilePath path = Fm::FilePath::fromPathStr(item.toStdString().c_str());
+        Fm::FileInfoPtr fileInfoPtr = proxyModel_->fileInfoFromPath(path);
+        filesToSelect.push_back(std::move(fileInfoPtr));
+    }
+    folderView_->selectFiles(filesToSelect);
+    // workaround to scroll and select the first item when there's multile files to select
+    if(filesToSelect_.count() > 1) {
```
tsujan, line 1093: I think you wouldn't need this workaround if you added the contents to `onUiUpdated()` and also set `scrolled` (in the latter) to `true` if `filesToSelect_` wasn't empty. Would you test that?

tsujan (changes requested):
I wanted to continue reviewing but, I'm afraid, I found a fundamental problem in the PR:

Since `org.freedesktop.FileManager1` is used, neither a second window of pcmanfm-qt can be opened nor pcmanfm-qt could even run if another file manager like dolphin is running.

ahsand97:
> I wanted to continue reviewing but, I'm afraid, I found a fundamental problem in the PR:
> 
> Since `org.freedesktop.FileManager1` is used, neither a second window of pcmanfm-qt can be opened nor pcmanfm-qt could even run if another file manager like dolphin is running.

Yeah I noticed the same thing when actually installing pcmanfm on my system, I could see the desktop but I couldn't open new windows for the app, is it something I did bad?? or is something still misssing about the implementation??

pcmanfm/application.cpp
```diff
@@ -272,7 +272,7 @@ bool Application::parseCommandLineArgs() {
     }
     else {
         QDBusConnection dbus = QDBusConnection::sessionBus();
-        QDBusInterface iface(QLatin1String(serviceName), QStringLiteral("/Application"), QLatin1String(ifaceName), dbus, this);
+        QDBusInterface iface(QLatin1String(serviceName), QStringLiteral("/org/freedesktop/FileManager1"), QLatin1String(ifaceName), dbus, this);
```
tsujan, line 275: This doesn't fix the main problem. If another file manager also registers `/org/freedesktop/FileManager1` (as dolphin does), pcmanfm-qt can't run after it. Of course, we don't want that; pcmanfm-qt should be usable everywhere.

tsujan: You need to

 1. Restore "org.pcmanfm.PCManFM", "org.pcmanfm.Application", and "/Application" in `application.cpp`;
 2. Restore `<interface name="org.pcmanfm.Application">` in `pcmanfm/org.pcmanfm.Application.xml`; and
 3. Register the service "org.freedesktop.FileManager1" and the object "/org/freedesktop/FileManager1" in `application.cpp`.

Then, if pcmanfm-qt is started before Dolphin, it'll control "org.freedesktop.FileManager1". If Dolphin is already started, pcmanfm-qt will start without problem but "org.freedesktop.FileManager1" will be controlled by Dolphin but we don't expect anything else.

Of course, I mentioned Dolphin just as an example.

ahsand97:
@tsujan I think now is done, I had  to create and register a new xml file, hope I did it the right way, thank u so much for ur reviews.

tsujan:
> I think now is done

Yes, you did. Well done!

Let's continue the review. ~ 50% is already done.

pcmanfm/tabpage.cpp
```diff
@@ -1078,4 +1081,24 @@ bool TabPage::canOpenAdmin() {
     return false;
 }
 
+void TabPage::onSelectItems() {
```
tsujan, line 1084: No workaround would be be needed if the code was in the right place. Please remove this method completely and change `onUiUpdated()` exactly as follows:

```c++
void TabPage::onUiUpdated() {
    bool scrolled = false;
    // if there are files to select, select them
    if(!filesToSelect_.isEmpty()) {
        Fm::FileInfoList filesToSelect;
        for(const auto& file : filesToSelect_) {
            Fm::FilePath path = Fm::FilePath::fromPathStr(file.toStdString().c_str());
            if(auto info = proxyModel_->fileInfoFromPath(path)) {
                filesToSelect.push_back(proxyModel_->fileInfoFromPath(path));
            }
        }
        filesToSelect_.clear();
        if(folderView_->selectFiles(filesToSelect)) {
            scrolled = true; // scrolling is done by FolderView::selectFiles()
            QModelIndexList indexes = folderView_->selectionModel()->selectedIndexes();
            if(!indexes.isEmpty()) {
                folderView_->selectionModel()->setCurrentIndex(indexes.first(), QItemSelectionModel::NoUpdate);
            }
        }
    }
    // if the current folder is the parent folder of the last browsed folder,
    // select the folder item in current view.
    if(!scrolled && lastFolderPath_ && lastFolderPath_.parent() == path()) {
.....
.....
```

pcmanfm/application.cpp
```diff
@@ -690,6 +699,88 @@ void Application::setWallpaper(QString path, QString modeString) {
...
+        launchFiles(QDir::currentPath(), uriList, false, false);
+    }
+}
+
+/* This method receives a list of file:// URIs from DBus and opens windows
+ * or tabs for each folder, highlighting all listed items within each. The
+ * input list is not sorted or grouped so we need to marshal it into groups
+ * by folder, then call our "reveal" method to show each group
+ */
+void Application::ShowItems(const QStringList uriList, const QString startupId __attribute__((unused))) {
```
tsujan, line 716: This method has 4 problems: its starting comment has redundant lines, it uses `QFileInfo`, the order of `uriList` isn't kept by it, and its choice of window isn't compatible with that of `launchFiles()`. Please change it to:

```c++
/* This method receives a list of file:// URIs from DBus and opens windows
 * or tabs for each folder, highlighting all listed items within each.
 */
void Application::ShowItems(const QStringList uriList, const QString startupId __attribute__((unused))) {

    QMap<QString,QStringList> groups;
    QStringList keys;
    for(const auto& u : uriList) {
        QString folder = u.section(QLatin1Char('/'), 0, -2);
        if(!folder.isEmpty()) {
            if(!keys.contains(folder)) {
                groups[folder] = QStringList();
                keys << folder; // keep the original order (QMap is sorted by key)
            }
            groups[folder].append(u);
        }
    }

    if(groups.isEmpty()) {
        return;
    }

    PCManFM::MainWindow* window = nullptr;
    if(settings_.singleWindowMode()) {
        window = MainWindow::lastActive();
        if(window == nullptr) {
            QWidgetList windows = topLevelWidgets();
            for(int i = 0; i < windows.size(); ++i) {
                auto win = windows.at(windows.size() - 1 - i);
                if(win->inherits("PCManFM::MainWindow")) {
                    window = static_cast<MainWindow*>(win);
                    break;
                }
            }
        }
    }
    if(window == nullptr) {
        window = new MainWindow();
    }

    for(const auto& k : qAsConst(keys)) {
        window->openFolderAndSelectItems(k, groups[k]);
    }

    window->show();
    window->raise();
    window->activateWindow();
}
```

pcmanfm/tabpage.cpp
ahsand97: Got it

pcmanfm/tabpage.cpp
```diff
@@ -328,6 +328,24 @@ void TabPage::onFolderStartLoading() {
...
+            scrolled = true; // scrolling is done by FolderView::selectFiles()
+            QModelIndexList indexes = folderView_->selectionModel()->selectedIndexes();
+            if(!indexes.isEmpty()) {
+                folderView_->selectionModel()->setCurrentIndex(indexes.first(), QItemSelectionModel::NoUpdate);
+            }
+        }
+    }
     // if the current folder is the parent folder of the last browsed folder,
     // select the folder item in current view.
     if(lastFolderPath_ && lastFolderPath_.parent() == path()) {
```
tsujan, line 351: You forgot change this line to:

```c++
    if(!scrolled && lastFolderPath_ && lastFolderPath_.parent() == path()) {
```

pcmanfm/org.pcmanfm.Application.xml
```diff
@@ -27,4 +27,4 @@
     </method>
     <property name="desktopManagerEnabled" type="b" access="read"/>
   </interface>
-</node>
+</node>
```
tsujan, line 30: GitHub is complaining about the lack of an empty end line here. Please add it.

pcmanfm/tabpage.h
```diff
@@ -280,6 +280,10 @@ class TabPage : public QWidget {
 
     void ceateShortcut();
 
+    void setFilesToSelect(QStringList files) {
+        filesToSelect_ = files;
+    }
```
tsujan, line 285: Please change `QStringList` to `const QStringList&`  and also sort the list, like:

```c++
    void setFilesToSelect(const QStringList& files) {
        filesToSelect_ = files;
        filesToSelect_.removeDuplicates();
    }
```

tsujan: Sorry, I meant removing duplicate items but wrote "sorting". Corrected now.

pcmanfm/tabpage.cpp
ahsand97: ohh you right, my bad

pcmanfm/mainwindow.cpp
```diff
@@ -2355,4 +2355,11 @@ void MainWindow::on_actionCleanPerFolderConfig_triggered() {
     }
 }
 
+void MainWindow::openFolderAndSelectItems(QString folder, QStringList items) {
```
tsujan, line 2358: Please make this small change here and also in `mainwindow.h`:
```c++
void MainWindow::openFolderAndSelectItems(const QString& folder, const QStringList& items) {
```

pcmanfm/tabpage.h
ahsand97: Got it, already pushed, sorry if I make a lot commits/push too fast haha

tsujan:
You're the fastest contributor I've ever worked with :)

Now, only `ShowItemProperties` remains — plus a hidden, old problem in the code, that is revealed by your PR. Fortunately, fixing the latter is very easy; so let's deal with it first.

If pcmanfm-qt is started in the split mode, calling `MainWindow()` with an empty path in `ShowItems` will cause a bad situation. The problem can be fixed by changing https://github.com/lxqt/pcmanfm-qt/blob/274f03549f3ac38080888adf2eff37cb63549928/pcmanfm/mainwindow.cpp#L149 to:

```c++
    splitView_ = path && settings.splitView(); // splt view needs a path
```

Would you make that change here?

ahsand97:
> Would you make that change here?

Yes of course, done :)

tsujan:
> Yes of course, done :)

Thank you very much.

Almost done. I didn't expect this PR to be ready so soon. Only `ShowItemProperties` and a general check remain. I'll do them later.

tsujan:
`ShowItemProperties` needs a complete rewrite for several reasons. You'll see why below.

Please do this:

1. Add `#include <libfm-qt/core/fileinfojob.h>` below the last occurrence of "#include <libfm-qt/core/...".

2. Define `ShowItemProperties` as:

```c++
void Application::ShowItemProperties(const QStringList uriList, const QString startupId __attribute__((unused))) {
    // FIXME: Should we add "Fm::FilePropsDialog::showForPath()" to libfm-qt, instead of doing this?
    Fm::FilePathList paths;
    for(const auto& u : uriList) {
        Fm::FilePath path = Fm::FilePath::fromPathStr(u.toStdString().c_str());
        if(path) {
            paths.push_back(std::move(path));
        }
    }
    auto job = new Fm::FileInfoJob{std::move(paths)};
    job->setAutoDelete(true);
    connect(job, &Fm::FileInfoJob::finished, this, &Application::onPropJobFinished, Qt::BlockingQueuedConnection);
    job->runAsync();
}

void Application::onPropJobFinished() {
    auto job = static_cast<Fm::FileInfoJob*>(sender());
    for(auto file: job->files()) {
        Fm::FilePropsDialog::showForFile(std::move(file));
    }
}
```

3. Add `void onPropJobFinished();` to `private Q_SLOTS:` in "application.h" and define it below `ShowItemProperties` as:

``` c++
void Application::onPropJobFinished() {
    auto job = static_cast<Fm::FileInfoJob*>(sender());
    for(auto file: job->files()) {
        Fm::FilePropsDialog::showForFile(std::move(file));
    }
}
```

tsujan:
Also added an emptiness check. It should be OK now.

ahsand97:
> Also added an emptiness check. It should be OK now.

Got it, I knew `ShowItemProperties` needed to be rewritten, that was the hardest one to do hahaha, fixing it now!

ahsand97:
> > shouldn't we add something like?
> 
> Yes, please do it. Only it's `dialog->`, not `dialog.`

done :)

tsujan (approved):
Thanks. Let's have it! I'll make some minor changes later — they aren't limited to this PR.

tsujan:
@stefonarch,  if you  want to test it, you could use these commands:

```
qdbus org.freedesktop.FileManager1 /org/freedesktop/FileManager1 ShowFolders "(" "FOLDER1" "FOLDER2" ")" "0"

qdbus org.freedesktop.FileManager1 /org/freedesktop/FileManager1 ShowItems "(" "FILE1" "FILE2" ")" "0"

qdbus org.freedesktop.FileManager1 /org/freedesktop/FileManager1 ShowItemProperties "(" "FILE1" "FILE2" ")" "0"
```

"0" is an arbitrary number. "FOLDERn"/"FILEn" is the full path of a folder/file.

pcmanfm/application.cpp
probonopd: @tsujan, @ahsand97 since this is probably relevant for Filer too I would like to learn about how to do this "the right way". I never quite understood why we need XML files when we can expose services over D-Bus. Isn't that kind of defeating the purpose of publishing stuff on D-Bus?

Since you apparently understand how this all works, can you point me to good documentation to learn about why XML files are needed here?

> If another file manager also registers /org/freedesktop/FileManager1 (as dolphin does)

Wouldn't we want whatever file manager is running to react to those D-Bus messages? E.g., if the user is running a KDE session with Dolphin, then clicking on "Show in file manager" in the Downloads window of a web browser should show the downloaded file in Dolphin. But on Lxqt where the user is running PcmanFM-Qt as part of the session, clicking on "Show in file manager" in the Downloads window of a web browser should show the downloaded file in PcmanFM-Qt, no? To achieve this, I was assuming that the first file manager that tries to register `/org/freedesktop/FileManager1` "wins". But maybe this assumption was wong?

tsujan: > Wouldn't we want whatever file manager is running to react to those D-Bus messages?

We do but we don't want that file manager to prevent pcmanfm-qt from running. Therefore, we need to register `org.freedesktop.FileManager1` only for that purpose. Previously, in this PR, it was also registered for providing a single-instance of the app, while `org.pcmanfm.PCManFM` should have been used.

probonopd: Thanks, got it.

ahsand97:
> I tested it for real immediately, worked fine already then. All those commands work fine too, except that the 2 properties windows are at the same position. It's a very nice feature to have, thanks to @ahsand97 ! Are there other use cases except open downloaded files?

I've seen that when u click show in folder in IntelliJ IDE's it uses it, VSCode too

pcmanfm/application.cpp
ahsand97: Hey @tsujan I just noticed this
```C++
QString folder = u.section(QLatin1Char('/'), 0, -2);
```
desn't work properly with some URIS, I just put one line to debug cuz I was getting an error when using

`qdbus org.freedesktop.FileManager1 /org/freedesktop/FileManager1 ShowItems "(" "file:///dev" "file:///etc/hosts" ")" "0"`

and it prints this:

image

as you can see for the first uri the folder should be "/" (root) but is empty so it gives an error, the other one works properly

tsujan: I'll fix it. Thanks for reporting.

tsujan: The fix is here: https://github.com/lxqt/pcmanfm-qt/pull/1529. We should have worked with `libfm-qt`'s path, instead of manipulating `QString`.

Now, relative paths are also supported, like this from `~/`:
```
qdbus org.freedesktop.FileManager1 /org/freedesktop/FileManager1 ShowItems "(" "Desktop" "./Videos" ")" "1"
```

PR #1528, opened 2022-02-08
application: Allow FileManager1 registration queuing

palinek:
If there is other entity already serving, just queue our request and
register as soon as the previous one deregisters.


tsujan:
Very nice. Thanks!

PR #1529, opened 2022-02-10
A fix for `org.freedesktop.FileManager1`

tsujan:


tsujan:
Merging for more tests by git users.

PR #1531, opened 2022-02-12
Proposal for new default settings

stefonarch:
Related https://github.com/lxqt/lxqt/discussions/2163
As mentioned, the wallpaper setting  here will override the one from the theme, better choose the same.

### DIFF
--- a/config/pcmanfm-qt/lxqt/settings.conf.in
+++ b/config/pcmanfm-qt/lxqt/settings.conf.in
@@ -1,5 +1,6 @@
 [System]
 IconThemeName=elementary
+FallbackIconThemeName=breeze
 SuCommand=lxsudo dbus-run-session -- %s
 TerminalCommand=
 Archiver=lxqt-archiver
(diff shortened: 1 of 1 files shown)


tsujan (approved):
Wow! I'd mistaken this for changing defaults in the code; should have been tired. GTM, of course.

Please merge it if you don't want to add other keys.

stefonarch:
It's still about first run on fresh vanilla installation, not sure for the default wallpaper and theme yet (clearlooks or light I'd say) - it should fit.

tsujan:
> not sure for the default wallpaper and theme yet 

OK. There's no rush.

And sorry for my mistake. IMO, it's exactly the default conf file that should be changed, instead of the default values in the code.

stefonarch:
IMO "waves-logo.png" is the most suitable choice for the default wallpaper.

tsujan:
I agree with what you see fit.

BTW, why has the automatic check failed? Old branch?

stefonarch:
https://repo.archlinuxcn.org/ è offline, no CI

PR #1554, opened 2022-03-22
Corrected a condition for transient filter-bar

tsujan:
Changing filter by typing inside the view wasn't a necessary condition for the existence of a transient filter-bar.


tsujan:
This is just more logical. Most users won't see a difference.

PR #1565, opened 2022-04-12
Prerelease changes

tsujan:


tsujan:
As it's usual before releases, the checks should fail. It's a false alarm.

PR #1583, opened 2022-05-09
Added search history

tsujan:
This patch requires https://github.com/lxqt/libfm-qt/pull/806.

The history items can be selected by Up and Down arrow keys or by opening the combo menu.

For respecting the user's privacy, the maximum search history is zero (= disabled) by default. It can be from 0 to 50, with separate lists for the name and content entries.

Also, a button is added to Preferences dialog for clearing the search history.

The options are in preferences → Advanced → Search.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1580

### DIFF
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -523,6 +523,8 @@ void Application::onFindFileAccepted() {
     settings_.setSearchContentRegexp(dlg->contentRegexp());
     settings_.setSearchRecursive(dlg->recursive());
     settings_.setSearchhHidden(dlg->searchhHidden());
+    settings_.addNamePattern(dlg->namePattern());
+    settings_.addContentPattern(dlg->contentPattern());
 
     Fm::FilePathList paths;
     paths.emplace_back(dlg->searchUri());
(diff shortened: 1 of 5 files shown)


tsujan:
False positives are expected from the automatic check.

stefonarch:
Real nice feature really.
I noticed that the window title isn't translatable:

schermata-13 mag 20:40

tsujan:
> I noticed that the window taskbar title/tab title isn't translatable

You're right. The reason is that "Search Results" is inside the file `vfs-search.c`, which is written in C. It was borrowed from the old `libfm`.

Later, I'll see if it's possible to put the text inside a C++ method and translate it.

tsujan:
So, no objection to merging this? I merged its libfm-qt requirement last night, to see how much it takes for CI to see it.

stefonarch:
No objection sure. I re-run the check.

stefonarch:
Wrong button ;)

tsujan:
It seems 12 hours are enough for CI to see the changes.

PR #1586, opened 2022-05-14
Localize "Search Results"

tsujan:
I didn't find a way of doing it in `libfm-qt` because it's deep inside the VFS C code.


stefonarch (approved):
schermata-14 mag 14:15

That was fast :)

tsujan:
Your review and translation was fast. Thanks!

PR #1596, opened 2022-06-02
Consider device pixel ratio when drawing wallpaper

tsujan:
Previously, wallpapers weren't sharp with scale factors > 1.

Also, an old problem is fixed in centering per-screen wallpapers with multi-screen setups. Previously, if the centered wallpaper was smaller than the screen size, the dektop background color would be removed from the screen part to its right and bottom.

Fixes https://github.com/lxqt/pcmanfm-qt/issues/1595


tsujan:
Tested by @hirayaku and me (also with multi-screen setups). Merging for more tests by git users....

PR #1601, opened 2022-06-08
Prevent an empty desktop path

tsujan:
Fall back to `$HOME/Desktop` if `~/.config/user-dirs.dirs` doesn't exist or doesn't contain a desktop path. But if `~/.config/user-dirs.dirs` contains an empty desktop path, accept it.

See https://github.com/lxqt/lxqt-session/issues/439 for the story.


tsujan:
Any objection to merging? Here, I just have used `QRegularExpression` instead of `QRegExp` and prevented returning of `QString()` under reasonable circumstances.

PR #1606, opened 2022-06-18
Allow customizing workspace margins on desktop

tsujan:
It's especially useful with panels/docks that don't reserve space but auto-hide on overlapping windows.

### DIFF
--- a/pcmanfm/desktop-preferences.ui
+++ b/pcmanfm/desktop-preferences.ui
@@ -136,10 +136,50 @@
           <string>Spacing</string>
          </property>
          <layout class="QGridLayout" name="gridLayout_3">
+          <property name="verticalSpacing">
+           <number>10</number>
+          </property>
+          <item row="0" column="4">
+           <widget class="QCheckBox" name="lockMargins">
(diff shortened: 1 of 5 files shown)


stefonarch (approved):
It's working fine. And yes: it's a kind of doubling the "reserve space on display" option in panels configuration and it's only reason  I see would be for autohiding panels (which doesn't really make sense IMO). No idea why but on the right side it's not possible in any way to have icons going under the panel even with 0px.

For me it's on the limit to have those options but I've no strong obiections.

tsujan:
> on the right side it's not possible in any way to have icons going under the panel even with 0px.

That depends on the screen size, panel size, desktop item size and desktop item spacing. There may not be enough space for another item inside that region.

tsujan:
It works fine with RTL too. Right and left are swapped both on Desktop and in the GUI of Desktop Preferences dialog.

tsujan:
As a side note, this feature also solved my old problem with Conky: icons don't go behind it anymore (I put special items on the right side of Desktop, where Conky is).

PR #1609, opened 2022-06-24
Add option for leading zeros in "Bulk Rename"

fredins:
Useful when there is more than ten files and you want to keep the right order in eg. the terminal.

image

### DIFF
--- a/pcmanfm/bulk-rename.ui
+++ b/pcmanfm/bulk-rename.ui
@@ -2,6 +2,14 @@
 <ui version="4.0">
  <class>BulkRenameDialog</class>
  <widget class="QDialog" name="BulkRenameDialog">
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>388</width>
+    <height>155</height>
+   </rect>
+  </property>
   <property name="windowTitle">
    <string>Bulk Rename</string>
   </property>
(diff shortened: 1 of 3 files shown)


tsujan:
Thanks for the PR and, especially, for the idea!

The situation is a little complex:

On the one hand, neither the original implementation nor this PR is aware of the localization (zero isn't `0` in all languages). On the other hand, the user may or may not want his locale to be respected with numbers (because, for example, `1001` becomes `1,001` in German).

To cover all possibilities and also implement your idea, I made another PR.

Superseded by https://github.com/lxqt/pcmanfm-qt/pull/1610

stefonarch:
Nice thing to have! Testing works fine here, didn't find issues, apart when localisation here is checked and numbers >999 at the second run the basename will contain the numbers over 1000  because of the dot in italian:

schermata-24 giu 21:02

tsujan:
@stefonarch 
You should mean https://github.com/lxqt/pcmanfm-qt/pull/1610; otherwise, a miracle has happened ;)

stefonarch:
Sure but I though to write here because the idea generated here :)

tsujan:
Thanks for testing!

I also tested with German and Egyptian Arabic (`LANG=ar_EG`; it seems standard Arabic uses Latin digits).

stefonarch:
I learned that it's the other way round: it's us that use arabic digits -Latin is MMCCXVI stuff :)

tsujan:
> Latin is MMCCXVI stuff :)

My bad English! I meant ۳ versus 3: ۰۱۲۳۴۵۶۷۸۹

tsujan:
> ... ending up with dots and ' in filenames

You probably mean `۰` and `۱`, which are `0` and `1`, respectively.

Putting `1` inside a Persian string is as weird as putting `۱` inside an English one. However, a Persian user may want to rename files with English names; hence the new option.

PR #1610, opened 2022-06-24
Added options for locale-awareness and zero-padding to bulk renaming

tsujan:
* An option is added for using localized numbers while renaming.
 * Another option is added for padding numbers with zero if possible (thanks to @fredins for the idea).

Both options are unchecked by default.


tsujan:
> It's not a feature that is used very much IMO from average users

I agree.

I added the option for localized numbers for the sake of consistency, but padding with zeros (@fredins's idea) can be useful.

PR #1619, opened 2022-07-25
Not really a pull request

ulliperk:


tsujan:
What's this?!

stefonarch:
The author deleted the description. I've it my email. The essential question was:
```
On Tanuki 0.3 filer closes on logout, shutdown or reboot, like it should.
Unfortunately on linux - Plasma 5, in x11 filer would not do so.
An error message would occur: "Logout/shutdown/restart cancelled by filer-qt."

Is there a way to end filer-qt automatically at logout, or shutdown?
```
the answer could be insert `killall filer-qt && ...` in  leave desktop files, but the question could have been posed in a better way at filer's GH page.

PR #1626, opened 2022-08-06
Ignore auto-selection delay when single-click isn't chosen

tsujan:
And even when the DE's setting is single-click activation.

Fixes https://github.com/lxqt/pcmanfm-qt/issues/1625


tsujan:
Straightforward. Merging...

PR #1630, opened 2022-08-10
Added an entry for Categories to launcher creator

tsujan:
Only because @stefonarch wanted it and it's costless.

I also found a small bug (in trusting the launcher) and fixed it.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1605

### DIFF
--- a/pcmanfm/desktopentrydialog.ui
+++ b/pcmanfm/desktopentrydialog.ui
@@ -178,6 +178,20 @@
        </item>
       </widget>
      </item>
+     <item row="3" column="0">
+      <widget class="QLabel" name="catLabel">
+       <property name="text">
+        <string>Categories:</string>
+       </property>
+      </widget>
+     </item>
+     <item row="3" column="1">
+      <widget class="QLineEdit" name="catEdit">
+       <property name="clearButtonEnabled">
+        <bool>true</bool>
+       </property>
+      </widget>
+     </item>
     </layout>
    </item>
    <item>
--- a/pcmanfm/desktopentrydialog.cpp
+++ b/pcmanfm/desktopentrydialog.cpp
@@ -31,6 +31,8 @@ void DesktopEntryDialog::onChangingType(int type) {
     else if(type == 1) {
         ui.commandLabel->setText(tr("URL:"));
     }
+    ui.catLabel->setVisible(type == 0);
+    ui.catEdit->setVisible(type == 0);
 }
 
 void DesktopEntryDialog::onClickingIconButton() {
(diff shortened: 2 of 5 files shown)


pcmanfm/desktopentrydialog.ui
```diff
@@ -178,6 +178,20 @@
...
      </item>
+     <item row="3" column="0">
+      <widget class="QLabel" name="catLabel">
+       <property name="text">
+        <string>Categories:</string>
+       </property>
+      </widget>
+     </item>
+     <item row="3" column="1">
+      <widget class="QLineEdit" name="catEdit">
```
stefonarch, line 189: <property name="toolTip">
        <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Main valid Categories are: AudioVideo;Audio;Video;Development;&lt;br/&gt;Education;Game;Graphics;Network;Office;Settings;&lt;br/&gt;System;Utility;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
       </property>
       <property name="toolTipDuration">
        <number>-1</number>

tsujan: I think such a tooltip would have no real use. Moreover, it would pretend that there's something special about categories, while exactly this section has no use for the app itself.

Also, here we don't want to explain the rules of `freedesktop.org`; otherwise, we should tell much more — while we even don't mention the difference between the types.

IMO, the proper place for explaining these things is https://github.com/lxqt/pcmanfm-qt/wiki. I'll add a section later.

stefonarch: IMO think it can have an use - specially for users who manage only very basic english and want only to add an app visible in the main menu. It doesn't hurt.

And there is something special about categories: for example  it's not "Internet" (as in the italian menu) but "Networking".

tsujan: I agree that the "Categories" section is useful, but it isn't related to file management. Attracting attention to it with an extra tooltip feels like a GUI inconsistency to me.

I think there's a way to make you happy without GUI inconsistency: adding a "What's this?" button, which will show info on every section. "What's-this" is better because
 * It doesn't attract the user's attention or distract him by appearing automatically;
 * It doesn't  disappear after a short time; so, it can contain more info compared to tooltips.

If you agree with that, I'll add it later, in a separate PR — it'll need some work.

stefonarch: Nice, I saw that in QtCreator but didn't tamper with it.

tsujan: "What's-this" texts are in the ui file, but "What's-this" itself is shown by the code.

I'm merging this....

PR #1643, opened 2022-08-19
Apply changes in shared settings to Desktop too

tsujan:
The settings shared by Desktop and windows are auto-delay selection, right-click behavior, showing of thumbnails and treating backup files as hidden. Now, they're applied on the fly.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1642

### DIFF
--- a/pcmanfm/desktopwindow.cpp
+++ b/pcmanfm/desktopwindow.cpp
@@ -127,6 +127,7 @@ DesktopWindow::DesktopWindow(int screenNum):
         proxyModel_ = new Fm::ProxyFolderModel();
         proxyModel_->setSourceModel(model_);
         proxyModel_->setShowThumbnails(settings.showThumbnails());
+        proxyModel_->setBackupAsHidden(settings.backupAsHidden());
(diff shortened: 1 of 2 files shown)


tsujan:
@stefonarch, could you please test it too? Particularly, enabling/disabling of thumbnails from the view menu should affect all tabs, all windows and Desktop.

stefonarch:
Checked thumbnails, autoselection and hidden files, all working over tabs, windows and desktop. 
What would "right-click behaviour" be?  I cannot find any  difference if "show context menu of the folder with ctrl+right-click" is checked or not, it will always be shown with or without Ctrl.

Btw  why have backup files a trash symbol, they are not meant for trashing ;)

tsujan:
Thanks!

> I cannot find any difference if "show context menu of the folder with ctrl+right-click" is checked

With that option checked, if you press Ctrl and right click a selected file, it'll be deselected and the folder menu will be shown. It's useful where there isn't an empty space to click (especially in the detailed list mode).

> Btw why have backup files a trash symbol....

That's what your icon set does. It makes sense: a trash shape implies uselessness, and backup files are useless as long as they're backup files ;)

PR #1649, opened 2022-08-21
Set QTerminal in default config file for LXQt

tsujan:
Also, removed redundant keys from old times

Closes https://github.com/lxqt/pcmanfm-qt/issues/1647


stefonarch:
GTM

tsujan:
It closes https://github.com/lxqt/pcmanfm-qt/issues/1647, but I'll open another issue for what I said in the last comment on that page.

PR #1693, opened 2022-12-09
Control flow statement spaces

ianfarrell47:
```
sed -i 's/if(/if (/g' pcmanfm/*.cpp pcmanfm/*.h
sed -i 's/for(/for (/g' pcmanfm/*.cpp pcmanfm/*.h
sed -i 's/while(/while (/g' pcmanfm/*.cpp pcmanfm/*.h sed -i 's/switch(/switch (/g' pcmanfm/*.cpp pcmanfm/*.h
```


tsujan:
The code style of each source should be respected. This is the code style of the original author in `pcmanfm-qt`, `libfm-qt`,..., and I've kept it everywhere intentionally.

PR #1694, opened 2022-12-12
Corrected and improved `README.md`

tsujan:
I hadn't read it before ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,14 +2,15 @@
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


stefonarch (commented):
Sometimes we stumble upon some real old stuff it seems :)
GTM, what about inserting 

## Features
 * tabs
 * split view
 * file search
 * last opened files
  and so on?

PR #1699, opened 2022-12-19
Enabled tab DND under Wayland

tsujan:
Based on the experience gathered by wrestling with Wayland + `libfm-qt`.

Also, tab DND is improved in general.

### DIFF
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -91,9 +91,6 @@ Application::Application(int& argc, char** argv):
     userDirsWatcher_(nullptr),
     lxqtRunning_(false) {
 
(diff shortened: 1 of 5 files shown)


tsujan:
@stefonarch, your test would be appreciated — there's no rush though.

You should be able to freely drag and drop tabs between windows or detach them by throwing them outside all windows under Wayland, as long as the split view isn't enabled.

stefonarch:
Oh sorry  I thought I had already written here: it's working fine!

tsujan:
Thanks! A second pair of eyes is always needed in the case of nontrivial PRs.

I think point releases of libfm-qt and pcmanfm-qt can be made now, for a smooth Wayland experience. Will wait a few days to see if something else comes up.

PR #1701, opened 2022-12-22
Corrected some text in Hidden Shortcuts diialog

tsujan:
Replaced `Drag+...` with `Drop+...`, to not encourage users to press a modifier before dragging.

### DIFF
--- a/pcmanfm/shortcuts.ui
+++ b/pcmanfm/shortcuts.ui
@@ -132,7 +132,7 @@
      </item>
      <item>
       <property name="text">
-       <string>Drag+Shift</string>
+       <string>Drop+Shift</string>
       </property>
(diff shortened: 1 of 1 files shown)


stefonarch:
I missed that somehow, it showed up in the translations now. 
IMO it is hard to translate and doesn't make sense in the context as it's stated already that it actions on drop - and  to drop you have to drag first:

screen_area_sab_08:07:09_

I'd revert this.

tsujan:
> and to drop you have to drag first

This about pressing modifiers before dropping, not before dragging. You've probably forgotten https://github.com/lxqt/pcmanfm-qt/issues/220#issuecomment-1362545853 → second paragraph.

PR #1707, opened 2023-01-01
Fixed regression in Desktop DND after changes related to Wayland

tsujan:
This depends on https://github.com/lxqt/libfm-qt/pull/870.

Fixes https://github.com/lxqt/pcmanfm-qt/issues/1706


stefonarch:
Testing stuff I'm wondering if the popup-dialog after DND could be omitted at all if the drag/release is in the same folder.

tsujan:
The case of desktop aside, there's no difference between DND from other places or from inside the same folder: there are at most three possibilities plus cancelling (= 4 possibilities).

On Desktop, internal DND just means moving items visually.

PR #1716, opened 2023-01-20
Don't try to activate windows under Wayland

tsujan:
Because it would be futile and give a warning message.

Later, if we decide on the Wayland compositor(s) we want to support, we might replace the changed blocks with Wayland codes.


tsujan:
Straightforward. Merging...

PR #1722, opened 2023-01-29
Changed label of option for disabling smooth scrolling

tsujan:
This needs https://github.com/lxqt/libfm-qt/pull/875 for affecting all view modes.


tsujan:
I modified the wiki accordingly.

PR #1726, opened 2023-02-05
Qt6 port

tsujan:


stefonarch:
I found a bug in PCmanFm-qt 2.0 : missing author :)
screen_area_dom_15:20:56_

At the first glance everything works.

tsujan:
Quoting from https://github.com/lxqt/libfm-qt/pull/778 (that was the wrong place):

Please check "Always show the tab bar" and confirm these problems:

 1. If you open pcmanfm-qt in your home twice, the second window will have no tab-bar. But if you open the second window in another folder, the tab-bar will be visible.
 2. pcmanfm-qt can't be stopped from LXQt Session Settings.

The first problem seems like a magic to me — maybe I'm tired and don't see something obvious!

stefonarch:
1) no,  both windows have tab bars
2) yes, no way

and 
3) hit ctrl+m, open the hamburger menu and drag the window: menu doesn't close until mouse release.

screen_area_dom_15:30:26_

tsujan:
> no, both windows have tab bars

But I checked under both X11 and Wayfire!

> yes, no way

This problem should be in lxqt-session.

> hit ctrl+m, open the hamburger menu and drag the window: menu doesn't close until mouse release.

I bet you that's the magic of the Wayland compositor you use. Sway?

tsujan:
BTW, the third magic also happens under Wayfire with all menus and all apps (whether Qt or GTK) *if Wayfire's move plugin is used* (here, it's Alt+drag). It isn't our concern.

stefonarch:
> I bet you that's the magic of the Wayland compositor you use. Sway?

No, back on wayfire. And I noticed only now another thing, still on wayfire: double decoration, and I use client side deco.

screen_area_dom_15:45:28_

tsujan:
> still on wayfire: double decoration,

Let's not mix problems of Wayland compositors with this.

For now, my only confusion is about the lack of tab-bar in the second window. You added to my confusion by saying that it didn't happen to you ;)

stefonarch:
On sway it's similar but the open menu remains "attached" to window when moving.
In session-settings it will close all other pcmanfm-qt windows.

stefonarch:
> > still on wayfire: double decoration,
> 
> Let's not mix problems of Wayland compositors with this.

Sure, but no other application had this until now, so it's something in the code related to qt6.

tsujan:
> .... so it's something in the code related to qt6.

The problem you see (and I don't) can be about anything but our code. The code doesn't set the decoration; WM does.

stefonarch:
Logged back into wayfire and the double deco is gone.

tsujan:
> In other apps I can't reproduce the menu magic

I can reproduce it with all Qt and GTK apps (even Firefox) by using `Alt+Drag` under Wayfire.

Now I'm enjoying the safety of X11, and the tab-bar problem is still ridiculing me. It's absolutely self-consistent: The first window is opened in a directory; then the second window is opened in the same directory; then tab-bar of the second window is invisible, until I reload that directory.

stefonarch:
Maybe we can track it: I removed `~/.config/pcmanfm-qt`, started it again  and I can reproduce it now, it has to be a setting.
[lxqt.tar.gz](https://github.com/lxqt/pcmanfm-qt/files/10611778/lxqt.tar.gz)

tsujan:
I'm 101% sure that it's a regression in Qt6's tab-bar. Until now, I've found two workarounds for it, but I like to find its cause in Qt6 source → "qtabbar.cpp".

That file has been heavily changed since Qt 5.15. It had a crashing bug in Qt 6.3, which is fixed in 6.4. However, it has at least two other bugs. One of them is related to styling — I know its cause and have included a workaround for it in Kvantum — the other one creates the current problem.

stefonarch:
> the other one creates the current problem.

But not with all settings apparently.

tsujan:
> But not with all settings apparently.

I can reproduce it with 3 settings, yours included.

stefonarch:
I've no idea, I reloaded my old settings and it's still here now. 
So  yes, probably the settings aren't involved, but remains the mistery why I hadn't seen it through different sessions.

tsujan:
It's fun. I found 6 small workarounds. All of them call some private Qt methods indirectly and make the tab visible — by adding it without text and immediately setting its text, or by hiding and immediately showing it, or by setting its elide mode again, or...

For now, I think it's better to leave the code as it is. If I find the time, I'll make a small code and attach it to a Qt bug report, provided that it isn't reported yet. If the bug persists when we really want to jump into Qt6, we could use one of the available workarounds.

tsujan:
Now that I stopped wrestling with the tab-bar, in case you didn't believe me when I said I could create the menu magic with every app, this is Firefox under Wayfire:

firefox

stefonarch:
I tried many times both with titlebar+mouse and right mouse drag and $MOD, no way for me, the menu closes always on the click somewhere in the window.

stefonarch:
I see logout is slower, my script does wait 2 sec after `lxqt-leave --logout` and now this is not enough, because of the desktop module. I added  `pcmanfm-qt --desktop-off` and it's fast again.

tsujan:
It's https://github.com/lxqt/pcmanfm-qt/pull/1726#issuecomment-1417968765 → 2. Didn't find time to look into it tonight.

tsujan:
OK, I remembered that I'd dealt with Qt6's changed behavior on app quitting in FeatherPad. Based on that, I fixed quitting with the desktop module, especially on logout (it had nothing to do with `lxqt-session`).

Hopefully no other problem is created, but it needs to be tested. Please test it thoroughly!

stefonarch:
Session logout is good now as before and module can be handled again by `lxqt-config-session`, nice!

On the other side: under wayfire I see double deco on a  second and third ecc pcmanfm-qt window and the tab bug is gone again  ;)

tsujan:
> ... and the tab bug is gone again

The tab bug exists. I've added a workaround for myself :P Might add it here too.

tsujan:
I added the workaround. Its removal will be very easy if the bug is fixed, and it'll be harmless if not removed.

I couldn't find a a minimal code that shows the bug (for reporting it). There should be something about the window layout that triggers it. Might investigate it later.

tsujan:
Any objection to merging? It has been used, tested and completed for more than a year. And a patch for completing the Wayland support for Desktop is ready to be applied as a PR after merging it.

PR #1727, opened 2023-02-08
Leave the pixmap cache limit to Qt

tsujan:
The code caused extra CPU usage in special cases by reducing the pixmap cache limit to 2048 KB (Qt's default is 10240 KB).

Closes https://github.com/lxqt/libfm-qt/issues/874


tsujan:
Straightforward; merging...

PR #1729, opened 2023-02-09
Added a title to desktop

tsujan:
The title is `pcmanfm-desktop<N>`, where `<N>` is the screen number (0 for usual setups). It can be used for setting WM rules, especially under Wayland compositors.

For now, the same title isn't prevented in pcmanfm-qt windows, but that will be easy to implement if really needed.

### DIFF
--- a/pcmanfm/desktopwindow.cpp
+++ b/pcmanfm/desktopwindow.cpp
@@ -88,6 +88,9 @@ DesktopWindow::DesktopWindow(int screenNum):
     setAttribute(Qt::WA_X11NetWmWindowTypeDesktop);
     setAttribute(Qt::WA_DeleteOnClose);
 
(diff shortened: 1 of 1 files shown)


tsujan:
Will rebase Qt6 branches soon.

stefonarch:
Works :)

isf63:
Does this fix KWin not distinguishing windows from the Desktop? As described here: https://github.com/lxqt/lxqt/wiki/ConfigWindowManagers#issues

tsujan:
No, it doesn't. That's a problem in KWin.

This (perhaps temporary) change is mostly for Wayland compositors — until Wayland Desktop handling can be done directly in pcmanfm-qt by using layer-shell-qt6.

PR #1737, opened 2023-03-03
Localize tab texts for special folders

tsujan:
The special folders are "search", Applications, Computer, Network, Trash and Desktop. More folders could be added, but these seem enough for now.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1736

### DIFF
--- a/pcmanfm/tabpage.cpp
+++ b/pcmanfm/tabpage.cpp
@@ -34,6 +34,7 @@
 #include <QLabel>
 #include <QToolTip>
 #include <QDir>
+#include <QStandardPaths>
 #include "settings.h"
 #include "application.h"
 #include "desktopentrydialog.h"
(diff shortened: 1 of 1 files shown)


tsujan:
@stefonarch 
The new strings should be translated before you see a difference. Please check all the above-mentioned folders.

I'll rebase the Qt6 branch after this is merged.

stefonarch (approved):
GTM

screen_area_ven_13:07:23_

tsujan:
Rebased the Qt6 branch.

tsujan:
There is a minor problem, that I saw only by taking a screencast and seeing it frame by frame: When the directory is changed to a special folder, the tab text is in English for a moment, and then it changes to the localization language quickly.

I'll fix it in the next PR.

PR #1738, opened 2023-03-03
Updated *.ts files

stefonarch:


stefonarch:
Looking at the changes mostly the translations are already there but marked as "unfinished".
I'll run this snippet over it again:
`  sed --in-place -re '/<translation .*type="unfinished"[^>]*>[^<]/ s/ ?type="unfinished"//' *_*.ts`

stefonarch:
Looks like in most languages only "Applications" was not present.

tsujan:
Yes, they're already translated elsewhere in the app -- except for Application and Trash, I guess.

PR #1745, opened 2023-03-13
Remove desktop's drop indicator on leaving desktop

tsujan:
Leaving the desktop during DND over a folder can happen under rare circumstances.


tsujan:
Straightforward; merging...

PR #1752, opened 2023-03-26
Keep dragged and dropped items together on Desktop

tsujan:
This patch preserves the order of dragged and dropped Desktop items as far as possible.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1750


tsujan:
I know that this may be a little complex for the first-time code reader, but its logic is simple, and it has passed my tests for both LTR and RTL. Any objection to merging?

PR #1758, opened 2023-04-08
Add window icon to Desktop Preferences

isf63:
Desktop Preferences window: no icon → 'preferences-desktop' icon


tsujan:
"preferences-desktop"  doesn't exist in GTK icon sets. Moreover, the main Preferences dialog doesn't have an icon either.

Both problems will be solved if we add an icon named "pcmanfm-qt" to the app. Then, those dialogs will use it automatically, because it has the app name. The app itself could use it too, instead of the folder icon.

My proposal is the simple SVG icon I have in my icon set:

pcmanfm-qt

isf63:
How about 'system-file-manager'?

tsujan:
I think having an icon for the app is a cleaner solution.

tsujan:
Oh, we have a problem in lximage-qt too: its icon is just a 48-pixel PNG image — hadn't noticed it with my icon set.

IMHO, we shouldn't be shy about adding simple SVG icons to our apps.

isf63:
For the folder windows, how about the icon be set to the user directories' icons and likewise for user-defined folder icons, and then default to 'system-file-manager'. This may be the current behavior, sorry I'm AFK.) This helps differentiate different folders in the taskbar. For preferences dialogs I think 'system-file-manager' is good; the reason I am opposed to a hard-coded icon is that users may want to use icons from any icon set they install.

tsujan:
> the reason I am opposed to a hard-coded icon is...

Who wants a hard-coded icon?! It's an app icon that's installed alongside other things — like with most other apps. Any icon set could override it by adding its own icon — again, like it's the case with most apps.

tsujan:
To be more clear:

We won't need to touch the code; we'll just add a simple SVG icon with the name "pcmanfm-qt" (= the class name).

isf63:
That sounds fine; evidently user directories and custom folder icons don't propagate to window icon (which is most notable in the taskbar) -- that's okay, and removes my concern that they would clash with an app icon.

Because pcmanfm-qt is semi-independent from LXQt, an app icon is reasonable. If anyone wants to improve upon the icon that can be done, and a no-code solution is ideal.

tsujan:
> That sounds fine; evidently user directories and custom folder icons don't propagate to window icon (which is most notable in the taskbar)

lxqt-panel's task-bar has an option to use either the window icon or the class icon. If we add the icon "pcmanfm-qt", it could show it. By default, it shows the window icon (= folder icon).

> Because pcmanfm-qt is semi-independent from LXQt...

Yes, both`libfm-qt` and pcmanfm-qt are independent from LXQt. They can be used everywhere: KDE, GNOME, Wayfire,...

isf63:
Side question:
>we shouldn't be shy about adding simple SVG icons to our apps.

Would you consider `lxqt-config-*` and `lxqt-admin-*` apps, because they only do configuration and are small, to not necessitate an app icon? I made two PRs for matching some of their window icons to the .desktop icon.

tsujan:
> Would you consider the lxqt-config-* and lxqt-admin-* apps

I remember that I saw a kind of problem in their icons and solved it by adding "class" icons to my icon set years ago. But I don't remember what the problem was or even if it was a problem at all. Should take a look at them later.

Anyway, the next version of LXQt will be released soon. We don't have enough time to make changes before that.

tsujan:
Please check https://github.com/lxqt/pcmanfm-qt/pull/1779, for this and other cases.

PR #1765, opened 2023-04-18
Check if wallpaper cache is up-to-date on reading Desktop settings

tsujan:
The modification time of the wallpaper file is checked to know if the cache is up-to-date, but only when Desktop settings are read (i.e., at startup or on applying Desktop Preferences).

This is needed when the wallpaper is changed but its path and name are not. It fixes https://github.com/lxqt/pcmanfm-qt/issues/1764

### DIFF
--- a/pcmanfm/desktopwindow.cpp
+++ b/pcmanfm/desktopwindow.cpp
@@ -569,7 +569,8 @@ QImage DesktopWindow::getWallpaperImage() const {
     return image;
 }
 
-QImage DesktopWindow::loadWallpaperFile(QSize requiredSize) {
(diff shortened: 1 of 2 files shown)


papoteur-mga:
Hello @tsujan ,
I tested your patch. I have rebuilt pcmanfm-qt with your patch, then installed it. I have installed the package which changes the background, then disconnected the session and relaunched it.
The background has changed \o/
Thanks for your work.

tsujan:
@papoteur-mga, thanks a lot for testing the patch!

It also passed my several tests. I think it can be merged, but let's wait a little more to see if something comes up.

BTW, you could also press the OK or Apply button of Desktop Preferences dialog to update the cache.

PR #1779, opened 2023-06-07
Added an SVG app icon

tsujan:
It will be used automatically by app-modal dialogs and, in general, by any window/dialog that takes its icon from the app. (On Wayland, it depends on how the WM handles icons.)

Of course, icon themes can override it.


tsujan:
Any objection? The icon is only used by the dialogs that were iconless before. It can be used in the desktop file too, but I didn't do so in this PR.

tsujan:
> Under wayland I see no change

It's mentioned in the PR comment.

> so it changed my icon in the taskbar

It depends whether you've checked Task Manager Settings → Use icons by.... (the last option).

In short, the app lacked an icon of its own. This PR provides one, and icon themes can override it.

isf63:
Seems good to me. The icon isn't as detailed as LXImage or LXQt Archiver but that can change. People can use 'Use icons by WindowClass' unchecked if they want the folder icon.

isf63:
Testing it for me adds the new icon to windows without a previous icon (Preferences, Desktop Preferences, Desktop → Create New, Bookmarks → Edit Bookmarks, ...). However, window icons don't change to the new icon where there was a specific window icon set (Tools → Find Files, Go → Connect to server, Help → About, ...)

Taskbar icons with "Use icons by WindowClass" checked is the new icon. The taskbar icon doesn't match window icons where the window icon was set.

tsujan:
All is expected and correct. Anyway, this is just an icon, not a code. Merging...

isf63:
For me instances of https://github.com/lxqt/pcmanfm-qt/blob/da42f545ca2f4aaf4cb403c43d2521124ae99e0b/pcmanfm/main-win.ui#L16 and similar override the new icon. So I'm only getting the changes on windows without that property / no .setWindowIcon(). Thoughts? — could be a glitch in my system.

tsujan:
> could be a glitch in my system.

No glitch. As the PR comment says, it's for windows that take their icon from the app == don't have icons of their own. That was the goal, in the first place.

PR #1785, opened 2023-06-18
Tab-order fixes to Create Launcher and Desktop Preferences dialogs

isf63:
~Used Qt Designer to populate correct \<tabstops\>~ Rearranged .UI/XML blocks to correct tabbing for:
- Desktop -> Create Launcher dialog (pcmanfm/desktopentrydialog.ui)
- Desktop -> Desktop Preferences dialog (pcmanfm/desktop-preferences.ui)

I believe this is all for the tab order fixes on .ui files in PCManFM-Qt

### DIFF
--- a/pcmanfm/desktop-preferences.ui
+++ b/pcmanfm/desktop-preferences.ui
@@ -162,6 +162,13 @@
             </property>
            </widget>
           </item>
+          <item row="0" column="2">
+           <widget class="QLabel" name="label_8">
+            <property name="text">
+             <string>x</string>
+            </property>
+           </widget>
+          </item>
           <item row="0" column="3">
            <widget class="QSpinBox" name="vMargin">
             <property name="toolTip">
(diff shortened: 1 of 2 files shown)


tsujan:
In both cases, the tab order is already correct (except for one, which I'll change later).

Please note that the tab order depends on the order of widgets. There's no need to add `tabstop` in all cases. At some point, I might remove them from https://github.com/lxqt/pcmanfm-qt/commit/13a68fe39230742b019bd9f6b113e0a49a3af6c3 too. The less cluttered a file is, the better.

isf63:
Create Launcher had one widget in the wrong order, otherwise correct. For Desktop Preferences, the General and Slide Show tabs had incorrect orders.

I'm guessing by
> the tab order depends on the order of widgets.

you mean the order as specified in the UI file? Fixing that would clean up the UI file and remove the need for \<tabstops\>, which sounds good to me.

tsujan:
> you mean the order as specified in the UI file?

Yes. I'll change them where needed. It should be done manually and carefully.

The problem is that, in most cases, Designer doesn't write XML blocks in the same order they appear in its GUI. I guess it might have been too complex to respect the order in Designer's code.

tsujan:
@isf63
Since you've already done a similar thing for `desktopentrydialog.ui`, please also move the block of `label_8` to the top of the block of `vMargin` in `desktop-preferences.ui`.

isf63:
Should be all good now.

tsujan (approved):

PR #1787, opened 2023-06-28
Reordered XML elements: connect.ui, preferences.ui

isf63:
Affected files:
- pcmanfm/connect.ui
- pcmanfm/preferences.ui

After this there is no usage of \<tabstops\> in PCManFM-Qt's 10 .ui files. Also they have all been checked to see that tab-order is correct.


tsujan (approved):
GTM

PR #1793, opened 2023-07-11
Desktop Prefs: Change no wallpaper option to groupbox checkbox

isf63:
Change the _Wallpaper mode_ combobox option "Fill with background color only" to a wallpaper groupbox checkbox.

### DIFF
--- a/pcmanfm/desktop-preferences.ui
+++ b/pcmanfm/desktop-preferences.ui
@@ -368,9 +368,12 @@ are left clicked, even when it is not the default file manager.</string>
         </widget>
        </item>
        <item>
-        <widget class="QGroupBox" name="groupBox_4">
+        <widget class="QGroupBox" name="enableWallpaperGB">
          <property name="title">
           <string>Wallpaper</string>
+         </property>
+         <property name="checkable">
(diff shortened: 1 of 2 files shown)


tsujan:
While appreciating your work, I see no reason for this change. The GUI is clear enough; adding a checbox doesn't make it clearer.

isf63:
On further inspection, `DesktopPreferencesDialog::onWallpaperModeChanged` could be simplified with this change (as a groupbox checkbox's default behavior is to disable the contents when unchecked.)

I think it does simplify the UI, but that's only my opinion. If the UI isn't clear enough a tooltip to the groupbox label could be added. If and only if this is a desirable, further changes should be made due to potential errors.

tsujan:
> I think it does simplify the UI

I'm afraid an extra box to check is the opposite of simplification.

isf63:
Okay, closing the PR. Also I accidentally overwrote a previous PR with this one, which I'll reopen.

tsujan:
> Also I accidentally overwrote a previous PR with this one, which I'll reopen

Yes, please reopen it. I saw it and wanted to review it when I found the time.

PR #1806, opened 2023-09-03
Fix tabs being reopened in incorrect view frame

### LINKED ISSUE #1329: PCManFM should restore tabs for each split.
##### Expected Behavior
After a restart PCManFM should restore tabs for each split.

##### Current Behavior
After a restart PCManFM moves the tabs to the left panel.

##### Possible Solution
If the current behavior is intended maybe state it near the "Reopen last window tabs in a new window" setting. But it's odd that I see PCManFM in a state then it gets mixed after a restart. Maybe new tabs could be added directly on the left panel and the right should have only one.

##### Steps to Reproduce (for bugs)
* Open a clean PCManFM (no configuration).
* Browse the settings: Edit -> Preferences -> User Interface -> Check "Reopen last window tabs in a new window". "OK".
* Press F6 to open split panel. In the split panel on the right mousewheel click a directory. Now the right panel has two tabs while the left has one.
* Close PCManFM.
* Open PCManFM.
* Now PCManFM should restore both the left panel tabs and the right panel tabs.

Instead:
* Now the left panel has all the tabs while the right panel only tab path is reseted to the home directory.

##### Context
I'm trying to continue from what I left. I may have for example few download directories on a panel and few directory to organize content on the right and they should restore like the previous session.

##### System Information
* Distribution & Version: Arch Linux
* Kernel: 5.10.20-1-lts
* Qt Version: 5.15.2-5
* libfm-qt Version: 0.16.0-2
* libqtxdg Version: 3.6.0-2
* lxqt-build-tools Version: not installed
* Package version: 0.16.0-1

riverbl:
Fix all tabs being reopened in the first view frame when both split view and reopen last window tabs are enabled.

Fixes #1329.

### DIFF
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -566,13 +566,27 @@ void Application::launchFiles(const QString& cwd, const QStringList& paths, bool
     Fm::FilePathList pathList;
     Fm::FilePath cwd_path;
     auto _paths = paths;
+    size_t splitIndex = 0;
 
     reopenLastTabs = reopenLastTabs && settings_.reopenLastTabs() && !settings_.tabPaths().isEmpty();
     if(reopenLastTabs) {
         _paths = settings_.tabPaths();
-        _paths.removeDuplicates();
+        splitIndex = settings_.splitIndex();
+        if(settings_.splitView() && splitIndex > 0 && splitIndex < (size_t)_paths.size()) {
+            // Remove duplicates within each frame leaving duplicates between frames.
+            QStringList rightFramePaths = _paths.mid(splitIndex);
+            _paths.erase(_paths.begin() + splitIndex, _paths.end());
+            _paths.removeDuplicates();
+            splitIndex = _paths.size();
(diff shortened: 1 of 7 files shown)


tsujan:
Thanks a lot for your contribution! The core idea is very good, i.e., remembering the number of tabs of the first frame, and adding the other tabs to the second frame.

The PR itself needs some changes. I'll be happy to review it line by line, but that might take some time, and I'm not sure we could finish it before the next release (currently I'm busy with various kinds of codes). Hopefully, we'll do it cleanly.

tsujan:
Your patch basically does the job. But the launching class doesn't need to know about tab splitting. Instead, the splitting can be done by `MainWindow::addTab`, such that the required patch becomes much cleaner and more debuggable.

Due to the amount of changes and lack of time, I made https://github.com/lxqt/pcmanfm-qt/pull/1807 and gave you the credit in it.

Superceded by https://github.com/lxqt/pcmanfm-qt/pull/1807.

PR #1807, opened 2023-09-07
Consider split view on restoring tabs

tsujan:
The credit goes to @riverbl for the main idea.

 Also, restoring of tabs with equal paths are allowed now because the option is about tabs, not just their paths (this should have been done before).

Closes https://github.com/lxqt/pcmanfm-qt/issues/1329

### DIFF
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -88,7 +88,8 @@ Application::Application(int& argc, char** argv):
     editBookmarksialog_(),
     volumeMonitor_(nullptr),
     userDirsWatcher_(nullptr),
-    lxqtRunning_(false) {
+    lxqtRunning_(false),
+    openingLastTabs_(false) {
 
     argc_ = argc;
     argv_ = argv;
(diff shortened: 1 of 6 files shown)


riverbl:
Note that with this PR, if you open two instances, both with split view, then close one instance, enable single window mode and launch pcmanfm-qt, the tabs from the closed instance will all be reopened in the currently focused frame of the running instance.

With my PR, the tabs from the closed instance are reopened in the same frame as they were in previously (left frame -> left frame, right frame -> right frame).

tsujan:
> ... the tabs from the closed instance will all be reopened in the currently focused frame

That's both intentional (I could have added codes for it) and logical, as the code comment says:
```c++
            // other folders have been opened explicitly in this window;
            // restoring of tab split number does not make sense
```
We want to restore the previous *session*, and that's not possible in this case. So, we just restore the tabs.

riverbl:
> With the previous patch, the path of the first tab of the second view would change.

I just tested this and it worked correctly for me - I did not get the bug you describe.

> That's both intentional (I could have added codes for it) and logical

No worries. I commented in case you hadn't considered that edge case.

tsujan:
> I commented in case you hadn't considered that edge case.

Yes, I guessed that — and thanks!

Actually, at first I wanted to disable restoring of tabs in that special case — even without the split view — because it feels a little weird seeing them inside a *used* window on clicking the app icon. I might do so as soon as a user complains about it.

stefonarch (approved):
Tested all cases that came to my mind, works as described.
I find not very clear the wording in the settings: "Reopen last window tabs in a new window".  _t doesn't do so if the window with the tabs isn't closed first.
"Reopen_ last window tabs when reopening PCManFm-Qt"?

tsujan:
> Tested all cases that came to my mind, works as described.

Thanks!

> It doesn't do so if the window with the tabs isn't closed first.

There should be a "last window" *after* enabling the option. If there is, its tabs will be restored in the next window (if it doesn't open folders explicitly) and then forgotten, until a window is closed again and serves as the "last window", etc.

PR #1810, opened 2023-09-17
Allow users to add terminal commands

tsujan:
Now, the terminal command entered by the user will be added to the list of terminals if the Preferences dialog is accepted (by pressing the OK button or pressing Enter inside the combo).

The format is as shown by the GUI, except for the redundant `%s`, which I didn't remove for the sake of translations — it's harmless. In most cases, just the terminal command is enough (without option).

The user can also right click inside the combo and remove a terminal if added previously or open the user-defined list if existing.

This PR is made based on https://github.com/lxqt/libfm-qt/pull/924 and needs it to work.

### DIFF
--- a/pcmanfm/preferencesdialog.cpp
+++ b/pcmanfm/preferencesdialog.cpp
@@ -21,10 +21,12 @@
 #include "preferencesdialog.h"
 #include "application.h"
 #include "settings.h"
+#include <QMenu>
 #include <QDir>
 #include <QHash>
 #include <QStringBuilder>
 #include <QSettings>
+#include <QStandardPaths>
 
 #include <libfm-qt/folderview.h>
 #include <libfm-qt/core/terminal.h>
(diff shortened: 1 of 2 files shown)


stefonarch (approved):
It's working fine, but as it still looks like a dropdown it's not clear that it can be edited. Maybe a tooltip?
It adds  working entries like ( fdd being a symlink to qterminal)

```
[fdd]
open_arg=

[footclient]
open_arg=
```
and I'm wondering what the lines additional lines like below actually are used for:
```
desktop_id=konsole.desktop
noclose_arg=--noclose -e
open_arg=-e
```
Afaik the only argument is the directory, and for that none of this is needed.

tsujan:
> it's not clear that it can be edited.

It's an *editable* combo-box (unlike that of archivers). That property was redundant before; now it can be used :)

> I'm wondering what the lines additional lines like below actually are used for:

`open_arg` is for execution . You can add it by adding the command as `COMMAND -e`, for example (it depends on the terminal); then that `-e` will be remembered but not shown by the GUI later.

`noclose_arg` is for keeping the terminal open after finishing the execution. QTerminal doesn't have it, for example. From `desktop_id`, additional info might be extracted.

I didn't add GUI elements for `noclose_arg` and `desktop_id` because they weren't crucial. But, of course, an advanced user can edit `~/.local/share/libfm-qt/terminals.list` manually. I'll add the info to the Wiki.

Editing isn't possible with this minimalistic GUI either. To change a command, the user needs to remove it (by right clicking) and add the corrected one. Since terminal commands + `open_arg` are short, I didn't see any reason to clutter the GUI for editing.

tsujan:
To make the best of the current GUI, I added "Open user-defined list" to the context menu (the usual message will be shown if it doesn't exist).

stefonarch:
Nice, works fine also that.

PR #1822, opened 2023-10-28
Fixed `--wallpaper-mode` command-line option

tsujan:
Fixes https://github.com/lxqt/pcmanfm-qt/issues/1821

### DIFF
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -676,8 +676,6 @@ void Application::preferences(const QString& page) {
 }
 
 void Application::setWallpaper(const QString& path, const QString& modeString) {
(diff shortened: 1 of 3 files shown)


stefonarch (commented):
Works fine. Wondering if we should improve the manpage, `-set-wallpaper `works only fine without "=" while `-wallpaper-mode `works with and without "=" .

stefonarch (approved):

tsujan:
Thanks!

> Wondering if we should improve the manpage, -set-wallpaper works only fine without "=" 

If `=` is needed, `--help` should tell about it. `pcmanfm-qt --help` doesn't include `=`. It's just by chance that it may work in some places.

PR #1837, opened 2023-12-04
 Add 5 menu icons to the main window (main-win.ui)

isf63:
5 menu icons added to the main window's toolbar menus: namely 'edit-select-all', 'utilities-terminal', 'network-server'

The icons exist in well-known icon sets; the Adwaita icon set has them but in "-symbolic" form.

### DIFF
--- a/pcmanfm/desktopwindow.h
+++ b/pcmanfm/desktopwindow.h
@@ -17,7 +17,6 @@
     51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
 
-
 #ifndef PCMANFM_DESKTOPWINDOW_H
 #define PCMANFM_DESKTOPWINDOW_H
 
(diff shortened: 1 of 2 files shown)


isf63:
@tsujan: I think the commits are good to go. The one caveat is that Adwaita is troublesome with "-symbolic" icons.

tsujan (approved):
> I think the commits are good to go.

I assure you that I hadn't forgotten it ;) We're very busy with the Qt6 port of LXQt — especially, with testing it and fixing its issues — and some Wayland stuff.

> The one caveat is that Adwaita is troublesome with "-symbolic" icons.

Not our problem. All the icons you've added here are compatible with https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html.

Anyway, GTM.

isf63:
Thanks! I failed to mention but wanted to say that there are identical patches for LXImage-Qt and libfm-qt. The Adwaita problem is moot anyway, because the change would be an unspecified icon to an undefined one.

PR #1858, opened 2024-02-08
Fixed a regression is setting wallpaper with command-line

tsujan:
The regression resulted in an empty background when the wallpaper was set by command-line without specifying its mode, i.e., by

```
pcmanfm-qt --set-wallpaper <FILE>
```

Fixes https://github.com/lxqt/lxqt/issues/2511

### DIFF
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -685,9 +685,15 @@ void Application::setWallpaper(const QString& path, const QString& modeString) {
         }
     }
 
-    DesktopWindow::WallpaperMode mode = DesktopWindow::WallpaperMode(Settings::wallpaperModeFromString(modeString));
-    if(mode != settings_.wallpaperMode()) {
-        changed = true;
+    DesktopWindow::WallpaperMode mode;
+    if(modeString.isEmpty()) {
+        mode = settings_.wallpaperMode();
+    }
+    else {
+        mode = DesktopWindow::WallpaperMode(Settings::wallpaperModeFromString(modeString));
+        if(mode != settings_.wallpaperMode()) {
+            changed = true;
+        }
     }
 
     // FIXME: support different wallpapers on different screen.


tsujan:
@stefonarch 
Please test with `--set-wallpaper`, `--wallpaper-mode` and then both.

If your tests show no problem, I think a point release will be needed.

stefonarch (commented):
See https://github.com/lxqt/lxqt/issues/2511#issuecomment-1935451069

But it looks that this was broken before this commit too, but is working in debian with1.4.0 so probably another fix is needed.

tsujan:
> See lxqt/lxqt#2511 (comment)

What's the relation to `--desktop-preferences`?! This is about `--set-wallpaper` without `--wallpaper-mode`.

stefonarch:
The img path `pcmanfm-qt --desktop-pref bg` is never updated when changing theme with "use wallpaper from theme". 
Before this commit it got black.
 So there is something broken, as in 1.4 it works.

tsujan:
> The img path pcmanfm-qt --desktop-pref bg is never updated

I tested with Qt5, and it's being updated here.

You should first close the Preferences dialog, set wallpaper, and then reopen Preferences dialog. It has never changed *on the fly* when Preferences dialog is kept open.

tsujan:
What I did:

 1. Compiled (Qt5) pcmanfm-qt with the patch.
 2. Stopped pcmanfm-qt's process completely.
 3. Started `./pcmanfm-qt --desktop` from a terminal opened inside the compilation directory.
 4. Used `--set-wallpaper` in another terminal and saw the wallpaper changed correctly.
 5. Used `--desktop-pref bg` and saw the path was set correctly.

If you see a different outcome with the above procedure, please tell me at which stage.

stefonarch:
Did exactly follow. It is **not** changed at 4)  here, also using `--wallpaper-mode=` options.

screen_area_ven_17:36:44_

As isn't  updated from `lxqt-config-appearance` changing themes with wallpaper as I said.

The weird thing is that in custom-actions `Exec=pcmanfm-qt -w %f` does work. Testing master qt5 now.

tsujan:
> Did exactly follow. It is not changed at 4)

What is not changed? The wallpaper? What exactly happens at 4 for you?

Let's be as clear as possible: You're saying that, after applying the patch to Qt5 pcmanfm-qt, `--set-wallpaper <FILE>` doesn't work? Sorry to mention the obvious, but are you sure you've used the *compiled* instance of `pcmanfm-qt` in your tests and not the installed instance, which of course has the bug? In other words, have you checked the process of pcmanfm-qt?

tsujan:
preferences

tsujan:
Oh, in your command (inside your screenshot), you haven't set any path. Of course nothing should be changed -- and it wasn't before.

stefonarch:
I was inside the directory, but yes, giving the full path it works. But in `lxqt-config` no way.

tsujan:
> But in lxqt-config no way.

I bet you've done something to it after so may tests related to Qt6 WIP PRs ;) Anyway, it isn't related to this PR (although I tested it).

I'm merging this, taking the full responsibility. Then I'll make a point release. If something's still wrong, the users will tell us, but it was a bad regression that needed an immediate fix.

Bluey26:
Hello.

Sorry for the intrusion, but i did compiled myself the commit and tried the described here.

I have downloaded the  commit (`b4a3243283b5b5853262089dd38a220098a7b6cd`), then i ran:

```
mkdir build && cd build
cmake ..
make
```

After that, i opened 2 terminals into the /build/pcmanfm directory, then i stopped the 'desktop' LXQt module(under session settings), because qps showed that killing the process will restart it afterwards.

Then, i ran in 1 of the terminals:

`./pcmanfm-qt --desktop`

In the other one, i wanted to run:

`./pcmanfm-qt -w ABSOLUTE_PATH_WALLPAPER`

 but it did not worked in the 1st time, because the 'default' background mode was 'color'. (I verified this 2 times, i think it is because i have not installed this, so there are no 'saved settings').

After running 1 time:

`./pcmanfm-qt --wallpaper-mode=stretch -w ABSOLUTE_PATH_WALLPAPER`

Then the background mode changed and things started to work as expected:

Whenever i ran :

`./pcmanfm-qt -w ABSOLUTE_PATH_WALLPAPER_2`

the wallpaper changed successfully to the wallpaper2, and so on.

After that, i checked that the lxqt-config-appearance  worked properly too, which was:

Marking the 'Use the wallpaper provided by theme' into 'LXQT Theme' tab, then changing between the different LXQt themes changed the background properly(which was the issue reported).

tsujan:
@Bluey26

Many thanks for the confirmation!

> ... because the 'default' background mode was 'color'.

Yes, because it was changed by the regression. That might be the case for @stefonarch too.

tsujan:
BTW, in the Qt6 version of lxqt-config, we might want to use `--set-wallpaper FILE --wallpaper-mode stretch`, because the mode may be color-only.

I'll definitely forget this until then ;)

stefonarch:
I found the reason: https://github.com/lxqt/lxqt-config/issues/978#issuecomment-1936962782

PR #1876, opened 2024-03-07
Use `layer-shell-qt` for Desktop under Wayland

tsujan:
Previously, workarounds provided by some Wayland compositors (like LabWC) were needed for having a usable Desktop under Wayland.

With this simple patch, and thanks to the developers of `layer-shell-qt`, PCManFM-Qt's Desktop is completely ready for Wayland. Don't forget to remove probable workarounds before using it.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1871


stefonarch:
GTM

"Let's give it to git users" (cit.)

stefonarch:
Maybe the AUR packages should be updated for all the merged Qt6 branches?
` yay -U pcmanfm-qt-git-1.4.1.4.gf219969e-1-x86_64.pkg.tar` while it's 2.0.0.
@yan12125

tsujan:
> Let's give it to git users

Agreed.

PR #1880, opened 2024-03-15
Fixed rare misplacement of desktop icons with Qt6

tsujan:
Apparently, Qt 6.6 requires that the desktop layout to be reset after the desktop folder is loaded. Sometimes it's needed after showing the desktop too.

Also, the scope of the layer shell is set to "desktop" under Wayland (it was "window" by default).


tsujan:
As nothing is really changed here except for adding a safeguard against item misplacement (by using `queueRelayout();` after showing the desktop), and because that safeguard worked here with several cold boots, I'm merging this for more tests by git users....

PR #1895, opened 2024-04-23
Add option to toggle ShellQt in CMake

mr360:
So today I went about building `pcmanfm-qt` from source.  

I am targeting X11 rather than Wayland. However, the build step fails because it requires `LayerShellQt`. To use it requires building it from source. Unfortunately, this requires that I have Qt configured with QWayland module. 

If I enable `QWayland` then Qt complains that `Error: ‘QX11Application’ is not a member of ‘QNativeInterface` So that is why I've added the following option so I can exclude LayerShellQt if I am not targeting Wayland in any way. 

@tsujan @stefonarch


tsujan:
We don't do this. → https://github.com/lxqt/lxqt/issues/2579#issuecomment-2137050865

PR #1913, opened 2024-06-01
Use own icon also in menu for consistency

stefonarch:


tsujan (approved):
GTM

tsujan:
BTW, do we want to keep the folder icon for the main window (on X11 title-bars, of course)? Since pcmanfm-qt opens folders, I think it's better to leave it as it is, namely, `folder`.

stefonarch:
>I think it's better to leave it as it is, namely, folder.
 
Me too.

PR #1915, opened 2024-06-06
Trim leading and trailing spaces to prevent issues when copying text …

Vescrity:
Trim leading and trailing spaces to prevent issues when copying text  with extra spaces.  

Sometimes when I use it to connect to SFTP, I just copy the host into the input box, which may contain extra spaces, causing the connection to fail. Then I have to input everything again. So, I used `.trimmed()` to ignore these spaces.   

I hope this PR doesn't affect other functionalities and can be considered for adoption.

### DIFF
--- a/pcmanfm/connectserverdialog.cpp
+++ b/pcmanfm/connectserverdialog.cpp
@@ -75,10 +75,10 @@ void ConnectServerDialog::onCurrentIndexChanged(int /*index*/) {
 
 void ConnectServerDialog::checkInput() {
   bool valid = true;
-  if(ui.host->text().isEmpty()) {
+  if(ui.host->text().trimmed().isEmpty()) {
     valid = false;
   }
-  else if(ui.loginAsUser->isChecked() && ui.userName->text().isEmpty()) {
+  else if(ui.loginAsUser->isChecked() && ui.userName->text().trimmed().isEmpty()) {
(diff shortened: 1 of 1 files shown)


pcmanfm/connectserverdialog.cpp
```diff
@@ -41,11 +41,11 @@ QString ConnectServerDialog::uriText() {
   uri = QString::fromLatin1(serverType.scheme);
   uri += QStringLiteral("://");
   if(ui.loginAsUser->isChecked()) {
-    uri += ui.userName->text();
+    uri += ui.userName->text().trimmed();
     uri += QLatin1Char('@');
   }
 
-  uri += ui.host->text();
+  uri += ui.host->text().trimmed();
```
tsujan, line 48: Since you trim the texts of `ui.userName` and `ui.host`, please also consider that inside the method `ConnectServerDialog::checkInput()`, by replacing `ui.userName->text().isEmpty()` and `ui.host->text().isEmpty()` with `ui.userName->text().trimmed().isEmpty()` and `ui.host->text().trimmed().isEmpty()` respectively.

Vescrity: Sorry for that. I've updated it.

tsujan (approved):
Thanks! GTM.

PR #1921, opened 2024-06-20
Separate multi-screen desktops under Wayland

tsujan:
For avoiding the problems of multi-screen desktops, each screen is given its separate desktop with a separate items layout under Wayland.

X11 isn't touched.

NOTE. IMO, this is the best solution for having multi-screen desktops under Wayland. The reason is a Wayland issue that limits an extended desktop to one screen as soon as any window is shown on a layer. A workaround for that issue was possible, but flickers couldn't be avoided.

### DIFF
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -459,7 +459,8 @@ void Application::desktopManager(bool enabled) {
             // NOTE: there are two modes
             // When virtual desktop is used (all screens are combined to form a large virtual desktop),
             // we only create one DesktopWindow. Otherwise, we create one for each screen.
-            if(primaryScreen() && primaryScreen()->virtualSiblings().size() > 1) {
+            // Under Wayland, separate desktops are created for avoiding problems.
+            if(!underWayland_ && primaryScreen() && primaryScreen()->virtualSiblings().size() > 1) {
                 DesktopWindow* window = createDesktopWindow(-1);
                 desktopWindows_.push_back(window);
             }
@@ -934,7 +935,8 @@ void Application::onScreenAdded(QScreen* newScreen) {
         connect(newScreen, &QScreen::availableGeometryChanged, this, &Application::onAvailableGeometryChanged);
         connect(newScreen, &QObject::destroyed, this, &Application::onScreenDestroyed);
         const auto siblings = primaryScreen()->virtualSiblings();
-        if(siblings.contains(newScreen)) { // the primary screen is changed
+        if(!underWayland_ // Under Wayland, separate desktops are created for avoiding problems.
+           && siblings.contains(newScreen)) { // the primary screen is changed
             if(desktopWindows_.size() == 1) {
                 desktopWindows_.at(0)->setGeometry(newScreen->virtualGeometry());
                 if(siblings.size() > 1) { // a virtual desktop is created
--- a/pcmanfm/desktoppreferencesdialog.cpp
+++ b/pcmanfm/desktoppreferencesdialog.cpp
@@ -83,7 +83,8 @@ DesktopPreferencesDialog::DesktopPreferencesDialog(QWidget* parent, Qt::WindowFl
 
   if (mode == DesktopWindow::WallpaperStretch || mode == DesktopWindow::WallpaperCenter
       || mode == DesktopWindow::WallpaperFit || mode == DesktopWindow::WallpaperZoom) {
-    ui.perScreenWallpaper->setEnabled(true);
+    // Under Wayland, separate desktops are created for avoiding problems.
+    ui.perScreenWallpaper->setEnabled(!static_cast<Application*>(qApp)->underWayland());
   }
   else
     ui.perScreenWallpaper->setEnabled(false);
@@ -246,7 +247,8 @@ void DesktopPreferencesDialog::onWallpaperModeChanged(int index) {
 
   if (mode == DesktopWindow::WallpaperStretch || mode == DesktopWindow::WallpaperCenter
       || mode == DesktopWindow::WallpaperFit || mode == DesktopWindow::WallpaperZoom) {
-    ui.perScreenWallpaper->setEnabled(true);
+    // Under Wayland, separate desktops are created for avoiding problems.
+    ui.perScreenWallpaper->setEnabled(!static_cast<Application*>(qApp)->underWayland());
   }
   else
     ui.perScreenWallpaper->setEnabled(false);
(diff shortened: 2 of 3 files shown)


stefonarch (commented):
Not sure how this happened but without the external monitor icons are misaligned even after reboot:


And when I change icon disposal on the second monitor and disconnect it after reconnecting they are aligned at left again, it looks like I need to stop the module first before disconnect it.

tsujan:
That isn't related to the patch. The patch is very simple: it gives a separate desktop to each Wayland screen. The code for multiple desktops already exists but is practically never used with X11.

The positions of desktop items are saved in `desktop-items-X.conf`, where `X` is the screen number. During your tests,  your external monitor should have become the screen No. 0 at some point. If you disconnect a monitor while the desktop module is running, its desktop will be removed and so, forgotten.

stefonarch:
It looks like tooltips on the desktop are broken here, noticed it only now.

tsujan:
I see no problem with tooltips, and you didn't explain what you meant by "broken". Even if there was a problem, (1) it would be a problem in Qt+Wayland, and (2) how could it be related to multiple screens when you say you have only one monitor?

stefonarch:
Broken=do not appear, but only on the desktop, inside folders they work.
Under x11 all is ok with this branch.

stefonarch:
Hm, items got partly rearranged by switching WMs, now I moved the trash icon and that made tooltips reappear.

tsujan:
Tooltips of desktop items appear under Wayland only if you focus the desktop (e.g., by clicking it). That isn't related to this PR: Wayland's background layer isn't focused automatically. I think we've already reported it to LabWC.

tsujan:
In an ideal world, lxqt-panel shouldn't be focused at all under Wayland, such that, if all windows are closed/minimized, the desktop should be focused, as it is under X11. We're still far from that ideal world ;)

marcusbritanicus:
These are a few drawbacks of the layer-shell protocol. It has three levels of "keyboard focus" - NoFocus, OnDemand, and Exclusive. Even though this protocol was developed for DE components, it is sadly, insufficient.

So, when a surface is OnDemand, most compositors give it focus when it first starts. In LXQt, the background starts first, gains focus, then the panel starts, and it takes the focus from background. And eventually, the other apps open and these layer-surfaces lose the focus.

Here is the tricky part: When all the normal windows are minimized or closed, there is no one to tell the compositor to give focus to the background. The protocol does not include a call where-in we can explicitly ask for a surface to be given focus.

The only solution I can think of for this is a sort of hack. We start the panel, the background and other layer-surfaces in `NoFocus` mode. Once all the components have started, and if there is no window that has focus, set the focus mode of the background to "OnDemand". When some window gains focus (and equivalently, background loses focus - something Qt can track), we set the mode back to `NoFocus`. If we receive a click, set the focus mode to `OnDemand`.

This will involve a decent amount of code. Whether this is to be implemented or not needs to be discussed. If at all we're implementing this, it should be after other major wayland issues are dealt with. But, personally, I feel that it's worth doing this. The situation is unlikely to change for the next couple of years at the very least - I have seen the draft of this protocol proposed to be added to wayland protocols, and it does not look very promising in this regard.

tsujan:
@marcusbritanicus
Thanks for the detailed explanation. It was in parallel with what I thought.

> This will involve a decent amount of code. Whether this is to be implemented or not needs to be discussed..., personally, I feel that it's worth doing this

On this, I disagree completely and firmly. IMO, such problems should be solved by Wayland compositors. Now that there are good and usable compositors like labwc and wayfire, the practical problems of Wayland are becoming more and more evident — previously, only a few devs might think about them, and almost no one heard their voice — and the solutions need to be implemented in the Wayland land, not as complex workaround inside apps.

marcusbritanicus:
> IMO, such problems should be solved by Wayland compositors.

Wayfire will not do this. And the reason is simple: it's not defined in the protocol. If it's not defined in the protocol, it will break compatibility with apps that follow the protocol.

> the solutions need to be implemented in the Wayland land

Yes, Absolutely. But as it turns out, for the Wayland devs, DEs are not a priority. It has taken 3-4 years for ext-toplevel-lists protocol to be accepted into Wayland. And we need two more, toplevel-states and toplevel-controls, which are in the pipeline, but are yet to be accepted. This is the situation when we already have a well functioning plasma's window management protocol and wlr-foreign-toplevel protocol. This is the sad reality of Wayland at the moment.

And we have additional issues. Even after the ext-toplevel-* protocols are accepted, they do not provide a means to identify their workspaces even though all compositors (barring cage and friends) have a well-defined concept of workspaces. This is the reason why I said it will take atleast a 2-3 years for things to improve, and that we should invest in code to make it easier for the users.

tsujan:
> And we have additional issues....

I agree with your analysis. My point is just that we shouldn't mess with our codes because of problems in Wayland. Sooner or later, Wayland devs need to remember that there's such a thing called a DE ;)

Let's use https://github.com/lxqt/lxqt/discussions/2531 for discussions on Wayland. This page is only about a simple workaround.

stefonarch:
Got a new adapter and tested again, it works fine. GTM

stefonarch:
Found one issue: if under x11 the  option "individual wallpapers for each monitor" is unchecked
immagine

under wayland wallpapers are as if this option is checked, no split can be seen. 

As this option can be changed only under X11, and if it's checked the wallpaper on the bigger monitor is split at the size of the smaller monitor:

immagine

tsujan:
Very good catch! Is it fixed now? (Sorry, I won't have access to my external monitor for a while.)

stefonarch:
I added a comment this morning but it disappeared. Now it's always splitted, with or without that setting checked (done in x11):

20240720_07h16m04s_grim

If I start with one monitor and add the second it's fine, until i restart the desktop module.

tsujan:
> If I start with one monitor and add the second it's fine, until i restart the desktop module.

That shouldn't be possible. Are you sure you have the latest commit?

tsujan:
Oh, it should be the case; I forgot to exclude another variable. Your tests ridiculed my logic ;)

tsujan:
I think it's fixed now, but I'm ready for your test to disprove it — it's Wayland, after all ;)

stefonarch:
Yes, it is fixed if my tests are correct, nice!

stefonarch:
Another small issue: icons are doubled on the screen and it's possible to hide them on one, but this setting is not preserved. At session restart both screen hide icons.

tsujan:
I'm aware of it, and I didn't add any code for changing it intentionally, because first, I want this PR to be only about excluding Wayland from the case of extended desktop (for the reasons I mentioned in the PR comment), and second, it doesn't seem an important UX issue to me.

Anyway, that small; issue isn't specific to Wayland; it's just that, under usual circumstances, you don't encounter a non-extended desktop under X11.

PR #1936, opened 2024-08-11
Enabled shortcut for inverting selection on Desktop

tsujan:


tsujan:
Straightforward. Merging....

PR #1950, opened 2024-10-08
Enable translucency for desktop under Wayland

tsujan:
Closes https://github.com/lxqt/pcmanfm-qt/issues/1949


kRHYME7 (approved):
Werks

stefonarch (approved):
Works fine. 
Why didn't you use  the  `setAttribute(Qt::WA_TranslucentBackground);`?
And why only  Wayland?

tsujan:
> Why didn't you use the setAttribute(Qt::WA_TranslucentBackground);?

I did ;) Only under Wayland.

> And why only Wayland?

Because window translucency makes sense only with compositing, and X11 can be without compositing.

stefonarch:
> I did ;) Only under Wayland.

I need new glasses or should were them ;)

PR #1956, opened 2024-10-19
Populate Bookmarks menu icons to match side pane

isf63:
One-line change

Uses the same icons in the side pane for the Bookmarks menu.

### DIFF
--- a/pcmanfm/mainwindow.cpp
+++ b/pcmanfm/mainwindow.cpp
@@ -1796,6 +1796,7 @@ void MainWindow::loadBookmarksMenu() {
     QAction* before = ui.actionAddToBookmarks;
     for(auto& item: bookmarks_->items()) {
         BookmarkAction* action = new BookmarkAction(item, ui.menu_Bookmarks);
+        action->setIcon(item->icon()->qicon());
         connect(action, &QAction::triggered, this, &MainWindow::onBookmarkActionTriggered);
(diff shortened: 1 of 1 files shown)


tsujan:
Sorry but, IMO, this ruins that sleek menu with an excess of duplicated icons. I've kept the menu intentionally as it is.

isf63:
> Sorry but, IMO, this ruins that sleek menu with an excess of duplicated icons. I've kept the menu intentionally as it is.

It was inspired by Thunar, which is in quite a few ways similar to PCManFM-Qt.

It depends on the icon theme whether the icons are distracting or not. But I'll heed your judgement.

----

I also noticed that for "Places" in the side pane and the "Go" menu, only 2 of the 6 items share the same position. Also 'Applications' and 'Network' have common icons in the side pane, where with `mate-icon-theme`, those items under "Go" have no icons. Thoughts?

tsujan:
> I also noticed that for "Places" in the side pane and the "Go" menu,...

Having menu icons shared by other widgets isn't a problem. For example, the Go menu has 5 *different* Places icons that make it easy to find the relevant item. In contrast, multiple folder icons in the Bookmarks menu would have no real use — most users don't change the icons of their bookmarked folders.

IMHO, giving icons to every menu-item may defeat the purpose of having icons, which is finding items more quickly. The Plus icon in the Bookmarks menu is enough; no Edit icon is needed either.

PR #1957, opened 2024-10-19
Updated *.ts files including Luganda

stefonarch:


stefonarch:
Have to redo that, after the release.

PR #1959, opened 2024-10-20
Silenced a warning about `GList`

tsujan:
Used `g_list_free_full`, which is exactly `g_list_foreach` followd by `g_list_free`.


stefonarch (approved):
No warnings at all now.

PR #1960, opened 2024-10-20
Add remaining icons to "Go" menu

isf63:
Give icons to 'Network' and 'Applications'

Icons (and fallback icons) were taken from the side pane (libfm-qt): https://github.com/lxqt/libfm-qt/blob/3375204b4d30a5239b9f56bb58a3a8c077fc9ba5/src/placesmodel.cpp#L63

It is not obvious how to set fallback icons in .ui XML, so I resorted to `setIcon`

### DIFF
--- a/pcmanfm/main-win.ui
+++ b/pcmanfm/main-win.ui
@@ -396,10 +396,6 @@
    <property name="text">
     <string>&amp;Network</string>
    </property>
-   <property name="icon">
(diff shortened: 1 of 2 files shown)


tsujan:
Please change `QLatin1String` to `QStringLiteral`.

isf63:
Done.

tsujan (approved):
Thanks.

PR #1970, opened 2024-11-05
Better drag-and-drop on Desktop

tsujan:
Previously and under certain circumstances, if a selection of desktop items was dragged and dropped where no other item existed, some neighboring items might be displaced.

Fixes https://github.com/lxqt/pcmanfm-qt/issues/1969


tsujan:
Tested by @isf63 and me. Its logic is simple too. Merging...

PR #1975, opened 2024-11-16
Workaround for a case of wrong wallpaper size on Wayland

tsujan:
Due to a bug in Qt, `QScreen::devicePixelRatio()` may give a rounded (ceiling?) integer on Wayland, while `QWindow::devicePixelRatio()` always gives the correct ratio.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1974

NOTE: This also works with an on-the-fly change of the scaling with kanshi.

### DIFF
--- a/pcmanfm/desktopwindow.cpp
+++ b/pcmanfm/desktopwindow.cpp
@@ -697,12 +697,16 @@ void DesktopWindow::updateWallpaper(bool checkMTime) {
         Settings& settings = static_cast<Application* >(qApp)->settings();
         const auto screens = screen->virtualSiblings();
 
+        // the pixmap's device pixel ratio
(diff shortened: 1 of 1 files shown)


stefonarch (approved):
For the wallpapers all fine now.
There is an issue with the icons, not sure if it's new: 
1. Order icons
2. Change scale to 1.2
3. Change back, icons are somehow ordered as the screen would be smaller: in a rectangle up left.

tsujan:
Ah, that may be because of the same Qt problem. I'll check if this workaround could be applied there too. Thanks!

tsujan:
I was wrong; it isn't a problem.

When you apply change the scaling, reserved positions may become invalid, and so, desktop items may change their positions. It usually doesn't happen on up-scaling, depends on the arrangement, and isn't predictable in general.

PR #1981, opened 2024-12-08
Support custom options for terminal

tsujan:
Entries for "Execution option" and "Custom option(s)" are added. Also, explanatory tooltips are added, instead of labels.

Also, closes https://github.com/lxqt/pcmanfm-qt/issues/1980

NOTE: This PR needs https://github.com/lxqt/libfm-qt/pull/1026.


tsujan:
BTW, later we should order our terminal and archiver lists alphabetically, either in codes that use them, or in our lists files.

stefonarch (approved):
Works fine now, didn't find issues.

PR #1983, opened 2024-12-09
Reorder "Go" menu

isf63:
To match the default Places in the side pane.

The two menus are effectively the same, so matching them makes sense. The side pane's order takes precedent.


tsujan (approved):
Yes, this is better.

PR #1985, opened 2024-12-09
Set fallback icon to support more iconsets

isf63:
There are two standard names for this icon, depending on if an iconset has GTK or KDE roots. Fixes no icon being displayed in certain cases.


tsujan:
Good, but please also remove the icon from the ui file. That'll result in 1 millisecond less computation ;)

isf63:
Done.

tsujan (approved):
Thanks.

PR #1988, opened 2024-12-20
Update filesystem info appropriately, with a delay

tsujan:
This patch updates the filesystem info ("Free space" on the statusbar) when all changes to the folder's contents are finished and the folder settles down, such that there is no extra I/O or CPU usage during changes.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1987

NOTE:
I'm a little concerned about this. As far as I've tested, and also based on what I know, I think it's compatible with the app's lightness. However, Dolphin doesn't do it, and Thunar does it only partially (only when files are added or removed, not when the size of a file grows or decreases). So, I'm not sure whether the devs of those apps — especially Thunar — have missed something, or I have ;)

So, I might not merge it (soon). Thorough tests would be very appreciated, particularly with remote filesystems that support this kind of info.

### DIFF
--- a/pcmanfm/tabpage.cpp
+++ b/pcmanfm/tabpage.cpp
@@ -110,6 +110,7 @@ TabPage::TabPage(QWidget* parent):
     verticalLayout{nullptr},
     overrideCursor_(false),
     selectionTimer_(nullptr),
+    filesystemInfoTimer_(nullptr),
     filterBar_(nullptr),
     changingDir_(false) {
 
@@ -406,7 +407,7 @@ void TabPage::onFileSizeChanged(const QModelIndex& index) {
 void TabPage::onFilesAdded(Fm::FileInfoList files) {
     if(static_cast<Application*>(qApp)->settings().selectNewFiles()) {
         if(!selectionTimer_) {
-            selectionTimer_ = new QTimer (this);
+            selectionTimer_ = new QTimer(this);
             selectionTimer_->setSingleShot(true);
             if(folderView_->selectFiles(files, false)) {
                 selectionTimer_->start(200);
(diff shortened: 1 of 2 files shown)


tsujan:
@isf63, could you please also test it?

isf63:
I moved a large ISO from `/home/...` to `/run/media/foo/...` with the two tabs open, the destination free space was always up-to-date when opening Folder Properties, but the source didn't update until I reopened the tab.

tsujan:
So fast. Thanks!

Please forget about the properties dialog for now, because this is about the "Free space…" label on the right side of the statusbar. Isn't it updated in both the source and the destination?

isf63:
For the status label it's again not updated on source for me. For destination it's updated in realtime.

tsujan:
Are you sure that both the source and destination folders are *open*? I can't reproduce it here.

isf63:
I tried both windows and tabs with the same outcome. My system tends to experience bugs though.

tsujan:
Never mind! And sorry for bothering you with this.

The fact is that, if someone pastes something into a subfolder of an open folder, we'll have no way of knowing about the change unless we check every subfolder, with any depth — and that's madness ;) Then, why should we update the info only when something changes immediately inside an open folder?

This was what I missed. The current behavior of pcmanfm-qt is quite acceptable: reloading is needed for updating FS info.

Thunar cheats: it updates and shows the FS info only when the selection is changed. If folder X is open in Thunar, and a heavy file is pasted into X *inside another file manager*, Thunar will show no change. It assumes that the user should select X on pasting, and once the selection is removed, Thunar updates and shows the FS info.

We have no reason to cheat.

PR #1991, opened 2024-12-27
Better tab sizes

tsujan:
Previously, a tab with a large enough text made the small texts of other tabs get elided.

The patch does two things. First, it limits the tab size to 2/3 of the width of the tabbar and, if needed, elides the text. Second, it prevents small tab texts from being elided, by making the best use of tab scroll buttons.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1990


stefonarch (approved):
Looks fine, wondering if it works for LTR too (`ElideRight`).

tsujan:
You mean RTL? Yes, it does. Qt puts the eliding at the left of RTL texts with `Qt::ElideRight` .

PR #1999, opened 2025-01-10
Prevent going up with auto-repeated Backspace in transient filter-bar

tsujan:
If the transient filter-bar is shown and Backspace is kept pressed, no going-up will happen. But if the transient filter-bar is emptied and Backspace is released, its next press will still cause going up (although without auto-repeating).

This is the only possible compromise between keeping Backspace for going-up and preventing an unwanted going-up with the transient filter-bar.

NOTE: Pressing Backspace isn't a good way of clearing the transient filter-bar because the selected file may change. Pressing Escape is always the best and easiest way. I might add that to pcmanfm-qt's Wiki page.

### DIFF
--- a/pcmanfm/mainwindow.cpp
+++ b/pcmanfm/mainwindow.cpp
@@ -339,16 +339,6 @@ MainWindow::MainWindow(Fm::FilePath path):
         connect(shortcut, &QShortcut::activated, this, &MainWindow::onShortcutJumpToTab);
     }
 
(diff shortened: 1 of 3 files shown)


stefonarch (approved):
Looks working fine, thanks!

tsujan:
Thanks for testing! Let's use it for a day or two before merging, although its code is simple.

morfikov:
It also works just fine for me now, thanks.

PR #2006, opened 2025-01-22
Handle Wayland screens by using their names

tsujan:
The main reason behind this change is that screen setting by using numbers may not work with all Wayland compositors.

With this patch, and on Wayland, the screen names are used for saving the arrangements of desktop items.

No change is made to X11 desktops.

NOTE: This is a backward incompatible change, in the sense that the arrangement of Wayland desktop items will be reset at first. But Wayland users could copy the contents of `desktop-items-0.conf` and paste it into `desktop-items-SCREEN_NAME.conf` after stopping the desktop module. That couldn't be done by the code, because it isn't known to which screen the old config file belonged (the number proves nothing).

### DIFF
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -470,7 +470,7 @@ void Application::desktopManager(bool enabled) {
                 int n = qMax(allScreens.size(), 1);
                 desktopWindows_.reserve(n);
                 for(int i = 0; i < n; ++i) {
-                    DesktopWindow* window = createDesktopWindow(i);
+                    DesktopWindow* window = createDesktopWindow(i, !allScreens.isEmpty() && underWayland_ ? allScreens.at(i)->name() : QString());
                     desktopWindows_.push_back(window);
                 }
             }
@@ -810,8 +810,8 @@ void Application::onPropJobFinished() {
     }
 }
 
-DesktopWindow* Application::createDesktopWindow(int screenNum) {
-    DesktopWindow* window = new DesktopWindow(screenNum);
+DesktopWindow* Application::createDesktopWindow(int screenNum, const QString& screenName) {
+    DesktopWindow* window = new DesktopWindow(screenNum, screenName);
 
     if(screenNum == -1) { // one large virtual desktop only
         QRect rect = primaryScreen()->virtualGeometry();
@@ -955,14 +955,14 @@ void Application::onScreenAdded(QScreen* newScreen) {
             }
         }
         else { // a separate screen is added
-            DesktopWindow* window = createDesktopWindow(desktopWindows_.size());
+            DesktopWindow* window = createDesktopWindow(desktopWindows_.size(), underWayland_ ? newScreen->name() : QString());
             desktopWindows_.push_back(window);
         }
     }
 }
 
 void Application::onScreenRemoved(QScreen* oldScreen) {
-    if(enableDesktopManager_){
+    if(enableDesktopManager_) {
         disconnect(oldScreen, &QScreen::virtualGeometryChanged, this, &Application::onVirtualGeometryChanged);
         disconnect(oldScreen, &QScreen::availableGeometryChanged, this, &Application::onAvailableGeometryChanged);
         if(!underWayland_) {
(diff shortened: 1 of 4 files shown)


tsujan:
The strange thing here is that — according to our tests — we can't keep a desktop without its screen (otherwise, a crash will be imminent), but we could do so in the case of lxqt-panel (https://github.com/lxqt/lxqt-panel/pull/2226).

I wonder if we were just lucky with lxqt-panel, or Wayland makes an exception for the background layer and doesn't keep it without its screen. @marcusbritanicus?

marcusbritanicus:
@tsujan I'm sorry, but I'm a little confused.. What do you mean by this: `we can't keep a desktop without its screen`?

So, here is what I did. I started pcmanfm-qt in desktop mode: `pcmanfm-qt --desktop` on wayfire with two screens. I closed one of the screens. pcmanfm-qt was still alive on the other screen. But I am fairly certain that this is not what you're asking me.

tsujan:
> What do you mean by this: we can't keep a desktop without its screen?

@stefonarch tested and saw that it crashed on Niri an some other compositors. That didn't happen with lxqt-panel.

Anyway, there was no reason to a keep a desktop on a nonexistent screen: the code removes it and will add it again if the screen is connected.

tsujan:
> But I am fairly certain that this is not what you're asking me.

Actually, it was ;)

marcusbritanicus:
Hmmm... I do not know how Niri works. But surely, that's not a "wayland" thing. Sway is one of the best references of how wayland should work. So, what I would do is test the same code on sway. If pcmanfm-qt survived the screen-closure on sway, then you can be sure it's a Niri thing, which I suspect it is.

Alternatively, another simple test would be to set the layer of lxqt-panel to background and check if the crash happens on Niri. That would also prove that it's a Niri problem.

Generally, (and this is not mandated by wayland) what happens when a screen dies is that the compositor, before releasing the output's resources, will move all the clients to a headless output or another working display, and then destroy the output. That way, the clients are not affected.

tsujan:
Actually, I'm not worried about pcmanfm-qt's desktop (we remove it when its screen is removed) but about lxqt-panel, which is on the top layer. We've assumed that lxqt-panel will be made hidden if its screen is removed. All compositors do that, as far as @stefonarch tested. I just wanted to know if this behavior is standard or happens only by chance because, in the second case, we should change the code and delete the panel as soon as its screen is removed.

marcusbritanicus:
Here, check out the attached code. You can compile it as `meson setup build && meson compile build` Start your compositor and run this app as `WAYLAND_DISPLAY=wayland-X ./build/layershell-test`, and then go ahead and close the "screen" window. You'll see that this app survives the screen death. Only when you kill the compositor, the app is killed.

Tested on wayfire, sway and labwc. 

I do believe, that should be the case on any other compositor as well: When the screen dies, the layer_surface object is closed, and as a result the widget associated with the layer_surface is also closed. And all this happens before the `wl_output *` and the associates are destroyed. Therefore, there is nothing that can kill the app.

[LayerShellTest.zip](https://github.com/user-attachments/files/18538919/LayerShellTest.zip)

marcusbritanicus:
As to your question, I think lxqt-panel code is perfectly fine. You can add a debug statement to the closeEvent(...),, and you're most likely going to see it being printed when the screen closes. After all, that's what happens in the case for the above test code as well.

tsujan:
Thanks! I'm not worried anymore :)

stefonarch:
> @stefonarch Could you please test the above code on Niri and see what happens?

```meson.build:29:6: ERROR: Dependency "wayqt-qt6" not found, tried pkgconfig and cmake
```
The behavior on Niri is not consistent, if I use
```
spawn-at-startup "sh" "-c" "swayidle -w timeout 300 'niri msg action power-off-monitors'"
``` 
when the monitors go off the desktop crashes, when I trigger the same command by shortcut no crash happens. If I disable one monitor and then the other one (by  `niri msg output <ID> off` then the desktop crashed too. Didn't test each and every compositor but it looks like a niri/smithay issue.

tsujan:
> when the monitors go off the desktop crashes

That's really a Niri issue, especially because the desktop is removed on removing its screen.

Here, the question was whether lxqt-panel was closed (= made hidden) when its screen was removed. Your tests, with so many compositors, showed that the answer was "yes".

marcusbritanicus:
> > @stefonarch Could you please test the above code on Niri and see what happens?
> 
> ```
> meson.build:29:6: ERROR: Dependency "wayqt-qt6" not found, tried pkgconfig and cmake
> ```

@stefonarch I'm sorry, I did not check the meson.build file. If you're still interested, I've updated the zip file. 

But from your comments, it seems that niri has a few problems, and not our apps.

[LayerShellTest.zip](https://github.com/user-attachments/files/18546862/LayerShellTest.zip)

stefonarch:
@marcusbritanicus The app survives fine, but `meson compile build` didn't work, used `ninja -C build`. The issue is somewhat deeper afaik on Niri as it depends also on the standby mode of my monitor afaik. Short suspends work fine. Thanks!

PR #2012, opened 2025-02-08
Change desktop Launcher cmd search path Home -> /usr/bin/

isf63:
"Select an executable file" -> vast majority are in /usr/bin/


tsujan:
Paths should never be hard-coded like `/usr/bin/`.

Apart from that, it's up to user to go to any directory from Home, which is the safest directory by default.

tsujan:
Not all binaries are in `/usr/bin/` — there can be user binaries in Home — and the user can easily go there if he wants to; we don't need tell him.

tsujan:
In old times, `~/bin` was a binary dir in any distro I tried. Now users should add it to their paths if they want it.

stefonarch:
If I remember well I once tried that battle too -- and lost too :)

tsujan:
Add to it the fact that most UNIX-based distros use `/usr/local/bin/`.

PR #2013, opened 2025-02-08
Replace `qMin`,`qMax`,`qBound`,`qRound` with standard funcs

isf63:


isf63:
There were no instances of `qAbs` here.

tsujan (approved):
GTM

PR #2014, opened 2025-02-09
Add missing <cmath> headers

isf63:
To check:
```
grep -R -P "(std::abs|std::round)"
grep -R "#include <cmath>"
```


tsujan (approved):
GTM

PR #2025, opened 2025-03-12
increase default maximum size of thumbnailed images

### LINKED ISSUE #2024: [Feature request] Increase default thumbnailer limit as photos nowadays are commonly larger than 4MB
**Is your feature request related to a problem? Please describe.**

I'm always frustrated when thumbs are missing for part of my photos. See https://askubuntu.com/questions/1543150/how-can-i-fix-qgis-logo-appearing-instead-of-thumbnail-of-a-photo

**Describe the solution you'd like**

Default settings more fitting more powerful computers and larger images would be nice

nowadays photos larger than 4MB are normal

**Describe eventual alternatives you've considered**

Replacing pcmanfm-qt or Lubuntu

(BTW, what is the point of "Describe eventual alternatives you've considered"? I never put anything really useful here, and maybe I should lie that I have not considered alternatives as above sounds quite passive-aggressive)

##### Context

I will try to send a PR but not promising that I will be able to get it over finish line (depends on how hard is to make minimum viable PR)

matkoniecz:
fixes #2024

note: if it is not autogenerated then docs also need to be manually changed, see https://github.com/search?q=repo%3Alxqt%2Fpcmanfm-qt+4+image+size+limit&type=code&p=3

### DIFF
--- a/config/pcmanfm-qt/lxqt/settings.conf.in
+++ b/config/pcmanfm-qt/lxqt/settings.conf.in
@@ -52,5 +52,5 @@ SidePaneMode=0
 
 [Thumbnail]
 ShowThumbnails=true
-MaxThumbnailFileSize=4096
+MaxThumbnailFileSize=40960
 ThumbnailLocalFilesOnly=true


matkoniecz:
I considered bumping it to 400 MB but not sure about potential performance problems

If thumbnailer explodes at processing 20MB photos then it is a thumbnailer bug. Though if it explodes at 350MB files it would be more reasonable (as of 2025).

Likely this limit should be bumped again in a few years. Maybe should be higher already.

matkoniecz:
note: this PR is untested

I can try testing it if it will not be flat-out rejected

tsujan:
→ https://github.com/lxqt/pcmanfm-qt/issues/2024#issuecomment-2716471326

matkoniecz:
→ https://github.com/lxqt/pcmanfm-qt/issues/2024#issuecomment-2716480328

Chiitoo:
I think the default should stay relatively low, because for one, the space consumed could very much explode without the user realising it.

I'm one of those who does his best with disabling the creation of these files on all the applications I use.  : ]

Also as mentioned elsewhere, distributions can change these defaults during packaging, though I think that might also 
be unlikely to happen (saying that as one such packager).

PR #2027, opened 2025-03-12
Allow string replacement in bulk renaming

tsujan:
With regular expressions, capturing groups are automatically supported (by Qt); for example, with `([^\s]+)\s+([^\s]+)` as the text and `\2 \1` as the replacement, `this that` becomes `that this`.

Also, locale-aware case change is supported.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1888

### DIFF
--- a/pcmanfm/bulkrename.h
+++ b/pcmanfm/bulkrename.h
@@ -32,6 +32,7 @@ Q_OBJECT
 public:
     explicit BulkRenameDialog(QWidget* parent = nullptr, Qt::WindowFlags flags = Qt::WindowFlags());
 
+    // renaming
     QString getBaseName() const {
         return ui.lineEdit->text();
     }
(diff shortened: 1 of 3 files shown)


pcmanfm/bulkrename.h
```diff
@@ -45,6 +46,31 @@ Q_OBJECT
         return ui.localeBox->isChecked();
     }
 
+    // replacement
+    bool getReplace() const {
+        return ui.relaceGroupBox->isChecked();
```
stefonarch, line 51: Tipo? `rePlaceGroupBox` ?

stefonarch (commented):
Didn't find issues testing it.

stefonarch (approved):

tsujan:
It wasn't easy to test; thanks.

> Didn't find issues

Me neither, but let's see what git users might find …

isf63:
Radio buttons might be better for three group boxes [Rename selected files; Replacement; Change Case]. Would be straightforward if Qt had radio button groupboxes, I don't think it does though.

stefonarch:
I thought some similar thing - ideally the options should be treated as equal, but didn't think of radio buttons. This would indeed much easier to switch back to rename only.

isf63:
-> https://github.com/lxqt/pcmanfm-qt/pull/2029

PR #2028, opened 2025-03-13
Updated *.ts files

stefonarch:


tsujan:
There are two new strings: `Nothing to find.`, when the user leaves the find entry empty, and `Invalid regular expression.`, when he chooses the regex search but enters an invalid regex.

I think the next step is showing the dialog again, with its previous state and texts, when these kinds of error happens. Maybe later …

stefonarch:
Just noticed them, showing again the dialog would be good.

PR #2029, opened 2025-03-13
Use QRadioButton's in Bulk Rename UI

isf63:


stefonarch:
Nice, didn't test using it. I wonder if it's possible to add some margin here:

immagine

isf63:
The margins are moderately better in Fusion:
capture

I think for some themes the margins are too narrow because radio buttons are typically inside a group box or otherwise not top-level.

tsujan:
Sorry, but those radio buttons are redundant. The original code already makes the group-boxes exclusive, and I saw no reason to add yet another group-box for the main functionality, exactly because it's the main functionality.

tsujan:
A compromise is made here: https://github.com/lxqt/pcmanfm-qt/pull/2030

PR #2030, opened 2025-03-13
Show bulk rename dialog again on failure

tsujan:
A failure means one of these: (1) nothing is renamed, (2) an empty string is searched, or (3) an invalid regex is searched. Then the dialog will be shown again with its previous state, after the error message is closed.

Also, as a compromise between different opinions, a group-box is added and all group-boxes are made mutually exclusive. This also simplifies the code a little.

### DIFF
--- a/pcmanfm/bulk-rename.ui
+++ b/pcmanfm/bulk-rename.ui
@@ -5,91 +5,81 @@
   <property name="windowTitle">
    <string>Bulk Rename</string>
   </property>
-  <layout class="QGridLayout" name="gridLayout">
-   <item row="0" column="0" colspan="3">
-    <widget class="QLabel" name="mainLabel">
-     <property name="text">
-      <string>Rename selected files to:</string>
+  <layout class="QVBoxLayout" name="verticalLayout">
+   <item>
+    <widget class="QGroupBox" name="serialGroupBox">
+     <property name="title">
+      <string>Serial Renaming</string>
      </property>
+     <property name="checkable">
+      <bool>true</bool>
+     </property>
+     <layout class="QVBoxLayout" name="verticalLayout_2">
+      <item>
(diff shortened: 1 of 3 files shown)


isf63:
Regarding the GUI, same as mine but with checkboxes, which is fine. QGroupBox has checkboxes built-in.

tsujan:
> QGroupBox has checkboxes built-in.

Yes. My point was that there's no need to more widgets. Also, we don't need to add extra indentations; the style draws them as they are.

stefonarch (commented):
Works fine, except that I can put anything as regex and I never get the "invalid regex" error, it just says "Nothing was renamed".
Tried also `^lxqt-.*|.*conf$` and it didn't filter the matches.

Nice the compromise, it's much better like that.

tsujan:
> I can put anything as regex and I never get the "invalid regex" error

A simple invalid regex: `[`

stefonarch:
>A simple invalid regex: [ — don't forget to check the regex check-box.

Maybe I've no idea about invalid regexes...

stefonarch (approved):

stefonarch:
> BTW, I had to change a string: `Serial Renaming`. The translations will be affected.

Oh, yes, nice catch.

tsujan:
The code has a part that prevents vertical resizing of the dialog — it was good when the dialog was simple, but I saw no harm in keeping it. kwin_wayland respects it, but strangely, labwc ignore it. I wonder why all Wayland compositors except for KWin are unkind to Qt! After all, Qt is the only platform-independent toolkit.

stefonarch:
Niri is Qt friendly and respect it :)


https://github.com/user-attachments/assets/b2d41a3f-53fc-4753-811a-596f17edc15c

tsujan:
> Niri is Qt friendly and respect it :)

Nice to see that.

PR #2050, opened 2025-04-16
Updated *.ts files

stefonarch:
@tsujan The base file update looks weird, but the translations look ok, please double check.

### DIFF
--- a/pcmanfm/translations/pcmanfm-qt.ts
+++ b/pcmanfm/translations/pcmanfm-qt.ts
@@ -918,359 +918,379 @@ are left clicked, even when it is not the default file manager.</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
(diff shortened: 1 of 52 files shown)


tsujan:
I issued `lxqt2-transupdate` inside it and nothing changed. I also compiled it without problem. So, it should be OK. The following warnings were also expected, weren't they?
```
File /home/pedram/Desktop/pcmanfm-qt-transupdate1111/pcmanfm/translations/pcmanfm-qt_kab.ts won't be updated: it contains translation but the target language is not recognized
...
File /home/pedram/Desktop/pcmanfm-qt-transupdate1111/pcmanfm/translations/pcmanfm-qt_lg.ts won't be updated: it contains translation but the target language is not recognized
```

> The base file update looks weird

Why weird? I found nothing suspicious in it.

PR #2060, opened 2025-05-20
Fixed shortcut removal with multiple desktops

tsujan:
With multiple desktops, if shortcuts were removed, error dialogs would be shown. The reason was that the shortcuts were removed in the first instance of desktop with a delay, and the code tried to remove them again in the other instances.

The problem didn't surface until each Wayland screen was given a separate desktop — on X11, there's only one extended desktop.

Fixes https://github.com/lxqt/pcmanfm-qt/issues/2059

### DIFF
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -853,11 +853,15 @@ void Application::updateFromSettings() {
     }
 }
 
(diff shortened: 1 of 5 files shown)


stefonarch (approved):
Works fine no errors displayed anymore. Testing some more days?

tsujan:
> Testing some more days?

Why not?

PR #2061, opened 2025-05-25
Prevent an imminent crash after removing a Wayland screen

tsujan:
Qt → `QWindowSystemInterface::handleScreenRemoved` removes the screen from the list `qApp->screens()` before emitting `qApp->screenRemoved`. Therefore, a nullity check should be added to the code. It's also safe under X11.


tsujan:
Nothing to review; only an important nullity check. Merging immediately…

PR #2074, opened 2025-06-28
cmake: fix build with Qt 6.10

Chiitoo:
The 'Qt6FooPrivate' targets have been split into separate CMake files in Qt 6.9, and require a 'find_package(Qt6FooPrivate)' call starting with Qt 6.10.

See also: https://bugreports.qt.io/browse/QTBUG-87776

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,6 +28,10 @@ find_package(fm-qt6 ${LIBFMQT_MINIMUM_VERSION} REQUIRED)
 find_package(lxqt2-build-tools ${LXQTBT_MINIMUM_VERSION} REQUIRED)
 find_package(LayerShellQt ${SHELLQT_MINIMUM_VERSION} REQUIRED)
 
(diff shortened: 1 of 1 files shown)


Chiitoo:
I've been indecisive with regards to style in that, where to put the `if` block, since on one hand it might look prettier below them all, but on the other hand, I like them to be close to the other Qt package calls.

For some, like this and `screengrab`, I noticed that the `if` call doesn't quite work always if before some of the other calls (sorted alphabetically), I guess because the Qt Gui target (or Core even) has not been found yet, so putting it below them all like here works without additional shenanigans... so maybe that makes most sense?

tsujan:
Strange! pcmanfm-qt doesn't depend on any private header and so, this shouldn't be needed.

tsujan:
Does this error happen *after* you apply https://github.com/lxqt/libfm-qt/pull/1060 and compile and install libfm-qt with it?

PR #2078, opened 2025-07-15
add --select compatibility flag

ItsZariep:
Some scripts rely on the `--select` flag to highlight files in file managers. While `caja`, `dolphin`, and `nautilus` support this flag, `pcmanfm-qt` currently does not.

Currently, `pcmanfm-qt` triggers an argument parser error when using --select, and if not, it open the file with the user selected mime app.

---

This simple PR adds support for the `--select` flag in `pcmanfm-qt` using the already existing `Application::ShowItems method`.

With this change, `--select` will be accepted by `pcmanfm-qt` without errors, improving compatibility with scripts expecting this flag.

### DIFF
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -210,6 +210,9 @@ bool Application::parseCommandLineArgs() {
     QCommandLineOption findFilesOption(QStringList() << QStringLiteral("f") << QStringLiteral("find-files"), tr("Open Find Files utility"));
     parser.addOption(findFilesOption);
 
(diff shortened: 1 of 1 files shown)


tsujan:
I'm afraid this is a duplication of a standard D-Bus method, which is independent of FM:
```
qdbus org.freedesktop.FileManager1 /org/freedesktop/FileManager1 ShowItems 'FILE_PATH' "0"
```

> While caja, dolphin, and nautilus support this flag…

While they support it, it isn't a standard, and we avoid duplication.

BTW, Caja and some other GTK file managers were forked from Nautilus.

PR #2081, opened 2025-07-26
Fixed showing of ampersand in recent files menu

tsujan:
Previously, ampersands in file names inside the recent menu were left to Qt, resulting in underlines or nothing, depending on the widget style.


stefonarch (approved):
Works nicely (although I've no file nowhere with an ampersand...)

PR #2091, opened 2025-09-10
Fixed a crash on mount failure after tab is closed

tsujan:
Fixes https://github.com/lxqt/pcmanfm-qt/issues/2090


tsujan:
> The message isn't shown anymore when tab is closed

Correct, because closing the tab means cancelling the operation now, which is what users may expect.

Previously, the operation wasn't cancelled and the code wanted to show an error message for a nonexistent tab; hence the crash. I encountered it only when I was checking https://github.com/lxqt/pcmanfm-qt/issues/1940#issuecomment-3262125642 and closed the window before the result came in.

PR #2095, opened 2025-09-27
Set icon for preferences window

City-busz:


City-busz:
These dialogs currently have no window icons, because these are not modal to the application window.

isf63:
I run kwin_x11, it shows the app icons on modal PCManFM-Qt windows. Please test another window manager if you can. If anything the app icon should be put in the `windowIcon` for WM edge cases.

tsujan:
In Qt, the dialog icon is set to the app icon. If Openbox (dead) or another X11 WM doesn't follow this rule, that should be fixed in it. I'm afraid we don't change these things because of Openbox.

tsujan:
> I opened https://github.com/lxqt/pcmanfm-qt/pull/2096 to set an icon for the application, so all windows without an icon inherit it.

Thanks! That's a better approach. Will check it later.

PR #2096, opened 2025-09-27
Set window icon

City-busz:
This ensures that the icon is set on all windows on X11.


tsujan:
I don't think there can be a problem with this but should use it under Wayland first.

@isf63 
Could you please check it under X11 too?

isf63:
No change on kwin_x11 as expected, but `pcmanfm-qt --desktop --profile=lxqt` with Openbox I think is not setting modal icons, unless I'm doing something wrong.

PR #2098, opened 2025-09-28
Small code cleanups (Qt5 -> Qt6)

isf63:
- Replace `Q_FOREACH`
- Rely on Qt6 disambiguated signals/slots


isf63:
Just an unknowing question: is there are reason `static_cast` is used so much throughout? It has readability advantages and is in some places needed, so I dared not remove them.

tsujan:
> is there are reason static_cast is used so much throughout?

It was the older way, before `QOverload` was introduced.

isf63:
I mean the multitudes of `static_cast<Application*>(qApp)`, and other instances.

tsujan (approved):

PR #2099, opened 2025-09-30
Assign rename file dialog to the parent window

City-busz:
This ensures that the dialog is attached to the main window.


tsujan (approved):
GTM.

PR #2100, opened 2025-10-01
Assign file selection dialogs to the parent window

City-busz:
This ensures that the dialogs are attached to the main window.


isf63:
Looks good and straightforward, but I'll wait for more input.

I believe I found an interesting bug in libfm-qt by testing this - navigating into `/home/user/Pictures` in Select Wallpaper Folder lets you hit Open to select the folder, but Open is greyed when immediately clicking Bookmarks → Pictures.

PR #2101, opened 2025-10-02
Make dialogs as modal only for the window

City-busz:
Without this, the dialogs block all opened windows of the application, not just the parent window.

### DIFF
--- a/pcmanfm/bulkrename.cpp
+++ b/pcmanfm/bulkrename.cpp
@@ -120,6 +120,7 @@ BulkRenamer::BulkRenamer(const Fm::FileInfoList& files, QWidget* parent) {
     bool showDlg = true;
     while(showDlg) {
         BulkRenameDialog dlg(parent);
+        dlg.setWindowModality(Qt::WindowModal);
         dlg.setState(baseName,
                      findStr, replaceStr,
                      replacement, caseChange,
--- a/pcmanfm/desktoppreferencesdialog.cpp
+++ b/pcmanfm/desktoppreferencesdialog.cpp
@@ -271,6 +271,7 @@ void DesktopPreferencesDialog::onBrowseClicked() {
 
   // use LXQt file dialog directly to set its view to thumbnail mode
   Fm::FileDialog dlg(this);
+  dlg.setWindowModality(Qt::WindowModal);
   dlg.resize(settings.wallpaperDialogSize());
   dlg.setSplitterPos(settings.wallpaperDialogSplitterPos());
   dlg.setAcceptMode(QFileDialog::AcceptOpen);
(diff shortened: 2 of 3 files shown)


tsujan:
That's exactly what we want: "Deal with this first!"

isf63:
I know little about *-modality, I approved the ones that were very straightforward (e.g. one of two buttons with a QDialog had a missing parent.) The icon PRs were gainful for Openbox, so I appreciate those. Again, like @tsujan said, please do not create duplicate PRs - if you can explain in very clear terms why these modalities are beyond-a-doubt gainful, then they may be reopened.

City-busz:
@tsujan @isf63 

A window in Qt can be `NonModal`, `WindowModal` and `ApplicationModal`:
| Value  | Description|
| ------------- | ------------- |
| `Qt::NonModal	`  | The window is not modal and does not block input to other windows. |
| `Qt::WindowModal	`  | The window is modal to a single window hierarchy and blocks input to its parent window, all grandparent windows, and all siblings of its parent and grandparent windows.
| `Qt::ApplicationModal` | The window is modal to the application and blocks input to all windows.

Source: https://doc.qt.io/qt-6/qt.html#WindowModality-enum

Dialogs in Qt are `ApplicationModal` by default:
> When an application modal dialog is opened, the user must finish interacting with the dialog and close it before they can access any other window in the application. Window modal dialogs only block access to the window associated with the dialog, allowing the user to continue to use other windows in an application.
Source: https://doc.qt.io/qt-6/qdialog.html#modal-dialogs

The `ApplicationModal` setting is suitable for applications that can have only one main window per process. But in PCManFM-Qt, you can open multiple main windows from the same application process. And the same process handles the desktop icons and the desktop settings to. So when an application modal dialog is opened in PCManFM-Qt, it blocks all these windows and also the desktop.

Test case:
1. Run the following commands, and leave the windows opened:
```
pcmanfm-qt
pcmanfm-qt --desktop
pcmanfm-qt --desktop-pref=general
pcmanfm-qt --show-pref behavior
pcmanfm-qt
```
2. Now open "Help > About" or "Help > Hidden Shortcuts" in one of the main window. Or click on the "Browse" button on the "Wallpaper" page of the desktop settings and leave it open.
3. Try to interact with the other windows or with the desktop icons.

Expected result:
Other windows and the desktop are usable, and only that window is blocked that the dialog is belong to.

Actual result:
Clicking on the other windows or desktop icons doesn't do anything, the access is blocked for all as long as the dialog is open.

Set the dialogs as `WindowModal` solves the problem. For the same reason, the dialogs within the Panel settings are set to `WindowModal` to avoid blocking the Panel while they are opened:
https://github.com/lxqt/lxqt-panel/commit/2aac43da7e04cca04c862ade091503f852044492

PR #2102, opened 2025-10-02
Make dialogs as modal only for the window

City-busz:
Without this, the dialogs block all opened windows of the application, not just the parent window.


tsujan:
Please don't make a duplicate PR/issue when the first one is decided.

City-busz:
Then don't close it without discussion.

PR #2110, opened 2025-11-02
Show desktop preferences only in LXQt

City-busz:
The desktop manager is [autostarted only in LXQt](https://github.com/lxqt/pcmanfm-qt/blob/f73e319d1b3cee9f18c0866937213fd5b70c45d6/autostart/lxqt-desktop.desktop.in#L5). Most desktop environments have their own desktop manager, therefore showing "Desktop" entry in their application launchers is confusing.

### DIFF
--- a/pcmanfm/pcmanfm-qt-desktop-pref.desktop.in
+++ b/pcmanfm/pcmanfm-qt-desktop-pref.desktop.in
@@ -4,3 +4,4 @@ Exec=pcmanfm-qt --desktop-pref=general
 Icon=user-desktop
 Categories=Settings;Qt;DesktopSettings;
 StartupNotify=true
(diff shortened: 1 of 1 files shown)


tsujan:
We've received comments by those who use pcmanfm-qt outside LXQt and are positively surprised to see that their desktops can really be used by this DE-agnostic app. So, my final answer is "no".

> I think it's more confusing to see the launcher for the pcmanfm-qt's desktop preferences in GNOME Shell or KDE Plasma.

Those DEs intrude on others with desktop files whose corresponding apps require lots of their specific apps/libraries. They don't deserve our pity ;)

stefonarch:
It can even be good :)

image

Issue #2157, opened 2026-05-05
No class/appid set on wayland on some systems


Kommynct:
##### Expected Behavior
when checking hyprland clients pcmanfm shows a title but no class it should be

class: pcmanfm-qt

##### Current Behavior

it outputs
class: 

##### Possible Solution
set the appid/class

##### Steps to Reproduce (for bugs)
1. open pcmanfm-qt
2. check appid/class

##### Context
this makes it impossible to set windowrules for, and also makes it impossible to add an icon for using hyprland-autoname-workspaces

##### System Information
* PcManFm-qt version: 2.4.0
* Distribution & Version: nixos unstable
* Qt Version: 6.10.2
* liblxqt Version:2.3.0

this has been an issue for years, I forgot about it until I decided to try pcmanfm again.

it does set the class if you QT_QPA_PLATFORM=xcb, but then you run it in x11 mode which is terrible

Kommynct:
```
	(final: prev: {
		lxqt = prev.lxqt // {
			pcmanfm-qt = prev.lxqt.pcmanfm-qt.overrideAttrs (old: {
				postPatch = (old.postPatch or "") + ''
					substituteInPlace pcmanfm/application.cpp \
						--replace-fail 'setApplicationVersion(QStringLiteral(PCMANFM_QT_VERSION));' \
							'setApplicationVersion(QStringLiteral(PCMANFM_QT_VERSION));
					    setDesktopFileName(QStringLiteral("pcmanfm-qt"));' \
						--replace-fail 'preferencesDialog_ = new PreferencesDialog(page);' \
							'preferencesDialog_ = new PreferencesDialog(page, MainWindow::lastActive());'
				'';
			});
		};
	})
```

this overlay fixes it on nixos, someone else do the PR please

this patch fixes it and makes the preferences dialog into an actual dialog and not a mainwindow, making it a floating window on tiling window managers


Set QGuiApplication::desktopFileName so that the Qt Wayland plugin emits
xdg_toplevel.set_app_id, allowing compositors to identify the window.

Pass the last active MainWindow as parent to PreferencesDialog so that the
Qt Wayland plugin emits xdg_toplevel.set_parent, allowing compositors to
treat it as a transient/dialog window.

```
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -97,6 +97,7 @@
     argv_ = argv;
[...]

PR #2158, opened 2026-05-05
wayland appid/class fix

Kommynct:
fixes the appid and class across wayland compositors when QT_QPA_PLATFORM=wayland is set

the current behaviour only works when it's QT_QPA_PLATFORM=xcb

### DIFF
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -97,6 +97,7 @@ Application::Application(int& argc, char** argv):
     argv_ = argv;
 
     setApplicationVersion(QStringLiteral(PCMANFM_QT_VERSION));
+    setDesktopFileName(QStringLiteral("pcmanfm-qt"));
     setWindowIcon(QIcon::fromTheme(QStringLiteral("pcmanfm-qt")));
 
     underWayland_ = QGuiApplication::platformName() == QStringLiteral("wayland");


Kommynct:
https://code.qt.io/cgit/qt/qtbase.git/commit/?h=dev&id=61ad604

here's documentation from qt showing this is the correct thing to do to set the app_id

fixes #2157 

"Add QGuiApplication::setDesktopFileName()
This property might be set by applications whose desktop entry file name
cannot be determined by heuristics already in place.
It is particularly useful for QtWayland as it can be used to determine
the app_id simply by stripping the ".desktop" suffix from this property.
Without a correct app_id, Wayland compositors won't be able to e.g.
show the application icon on task managers.
This property is also very interesting for X11 as there are various
desktop environments trying to map windows to launchers.
It will be possible to export desktopFileName as a xproperty, making
such mapping less error prone."

Kommynct:
additional context, kde does it nearly treewide

https://github.com/search?q=org%3AKDE%20QGuiApplication%3A%3AsetDesktopFileName&type=code

https://github.com/search?q=org%3AKDE%20setDesktopFileName(&type=code

also this confirms this is correct: https://doc.qt.io/qt-6/qguiapplication.html#desktopFileName-prop

the reason you haven't gotten reports is that this mostly only effects non-standard fhs systems, this should be added for compat and qt recommends it.

you are currently relying on undefined behavior to list your program.

stefonarch:
I'm fine with it as it doesn't add unwanted issues as far as I can tell.

Kommynct:
can I actually set the preferences window to have the class pcmanfm-qt-preferences that way it can be used with a separate windowrule? as it currently stands there's no way to select the preferences window and NOT a folder named Preferences

alternatively an xdg-tag could be used?

edit: nvm i'll do this in a separate pr if you say yes and which approach is better, I think pcmanfm-qt-preferences is probably simpler, way less likely to break something

stefonarch:
> currently stands there's no way to select the preferences window and NOT a folder named Preferences

No reason for.

```
Window ID 21: (focused)
  Title: "Preferenze della scrivania"

Kommynct:
> > currently stands there's no way to select the preferences window and NOT a folder named Preferences
> 
> No reason for.
> 
> ```
> Window ID 21: (focused)
>   Title: "Preferenze della scrivania"
> ```

I don't understand, the title is simply the name of the folder, if you wanted the preferences window to be a floating window you could match class:pcmanfm-qt and title:Preferences

but that would also impact you if you entered a folder named Preferences (or in your case Preferenze della scrivania), there is no way currently to distinguish between these

If you want the class to stay the same you could use an xdg-tag which is exactly what that is for.

stefonarch:
Hardly there will be another window with that exact title.

Kommynct:
> Hardly there will be another window with that exact title.

That's true for your language, but a preferences folder is quite common, I have some.

Kommynct:
Alt option, add "— pcmanfm" to the end of the preferences title

Kommynct:
Please merge that other question should be tackled in another mr, also apparently qt doesn't support xdgtags yet so nvm on that option

basil:
@stefonarch I tested this by copying the binary before and after this PR. Before this PR, the `initialClass` was (incorrectly) the name of the copied binary. After this PR, PCManFM-Qt explicitly sets its desktop file name/app ID to the correct `pcmanfm-qt`, making the Wayland app ID stable and no longer depends on what the executable file happens to be named.

LGTM

tsujan:
Although I have a bad feeling about compensating for others' shortcomings or meeting their specific needs in our apps, this one seems harmless and can be merged. *But please don't go further in this direction!*

PR #2162, opened 2026-05-23
set preference class to pcmanfm-qt-preferences

Kommynct:
this allows setting up window rules for the preferences dialog with no real downsides?

any reason not to do this would be appreciated, it really helps with tiling WM's especially and doesn't hurt anything that I can distinguish, very small change

### DIFF
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -677,7 +677,9 @@ void Application::preferences(const QString& page) {
     else {
         preferencesDialog_.data()->selectPage(page);
     }
-    preferencesDialog_.data()->show();
+    QGuiApplication::setDesktopFileName(QStringLiteral("pcmanfm-qt-preferences"));
+	preferencesDialog_.data()->show();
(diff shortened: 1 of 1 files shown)


Kommynct:
@tsujan any reason to oppose this change? it should be harmless.

Kommynct:
Any reason to oppose this change? @stefonarch @basil 

any variant that would be accepted to fix this issue?

tsujan:
→ https://github.com/lxqt/pcmanfm-qt/pull/2158#issuecomment-4516984976 (and many other places). Simply put, patchworks should be avoided, to say nothing of backward incompatible changes.

Kommynct:
> → [#2158 (comment)](https://github.com/lxqt/pcmanfm-qt/pull/2158#issuecomment-4516984976) (and many other places). Simply put, patchworks should be avoided, to say nothing of backward incompatible changes.

But this isn't a backwards incompatible change and I see no reason to avoid it?

PR #2164, opened 2026-05-27
Remove translations CMake artifact

luis-pereira:
From the Razor-qt days.


tsujan (approved):
GTM.

> From the Razor-qt days.

Interesting! I didn't know that.

PR #2173, opened 2026-06-30
Clean up wording in Desktop Preferences menu

lq259:
**Change "Slide Show" to "Slideshow".**
This matches more common word usage.

Note: Translations not updated.

### DIFF
--- a/pcmanfm/desktop-preferences.ui
+++ b/pcmanfm/desktop-preferences.ui
@@ -470,13 +470,13 @@ are left clicked, even when it is not the default file manager.</string>
      </widget>
      <widget class="QWidget" name="slidePage">
       <attribute name="title">
-       <string>Slide Show</string>
+       <string>Slideshow</string>
       </attribute>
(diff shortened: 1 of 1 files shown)


isf63:
"Slideshow" and "Slide show" are the two most common variants in English. "Slide Show" (both capitalized) is the one that is uncommon, but may appear in title-case.

I'm unsure of the change mostly because it would reset translations

lq259:
Fair

tsujan:
Changing a translatable string has a cost, which is justified only if that change is really needed. While appreciating your attention to such details, I agree with @isf63.

BTW, I'd searched Internet before choosing "Slide Show". In English, compound words are usually split when written — unlike in German, where something like *Freundschaftsbeziehungen* is quite common ;)  Of course, this doesn't mean that we should change "slideshow" to "slide show" everywhere in LXQt; both are correct.

stefonarch:
>Freundschaftsbeziehungen

You meant the recent  _Freundschaftsbeziehungskrisenzunahme_ I guess ;)

PR #2174, opened 2026-07-04
Workaround for a bad behavior of wlroots

tsujan:
WARNING: This is *not* intended to be merged — at least not soon. I put it here just for the reference.

The problem started with wlroots 0.19. See https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5127 for a reverting patch as well as discussion.


stefonarch:
I'd include this in the point release too.

tsujan:
Are you sure it has no side effect?

stefonarch:
I didn't notice any and I was running it often, as every time I switched to the testuser the desktop was gone. So not 100% sure, but afaik there's little hope for wlroots to fix that soon. If there are issues I didn't notice git users could spot them at least.

tsujan:
OK. I can't predict a side effect on the coding level either. I'm merging it, but very reluctantly.

PR #2180, opened 2026-07-31
pcmanfm: persist detailed-list column order

### LINKED ISSUE #1248: persist column position
Sorry how to persist column position settings? for example I need to have "Modified" as 2nd column after "Name"

romildo:
Store the visual order of detailed-list columns in the active profile and restore it whenever the detailed view is created.

Validate the stored order against the current model before applying it, so a change in the available columns falls back to the model's current order.

This complements the existing persistence of column widths and visibility.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1248


tsujan:
It should be done in libfm-qt, with due care for several other factors, and that's why I'm not sure it's worth the extra needed codes there.

tsujan:
→ https://github.com/lxqt/libfm-qt/pull/1113

tsujan:
Closed in https://github.com/lxqt/libfm-qt/commit/182b0cb96c6aafc72654ec4f3756bee196b612bd

PR #2190, opened 2026-08-29
Enhance search experience

ahsand97:
## 🔍 Enhance the Search Experience (pcmanfm-qt)

**Find Files** (F3) used to replace the tab you were browsing with a blank view and zero feedback — no spinner, no progress, no way to stop it short of closing the tab, and a busy cursor that made the whole app look unresponsive from an user experience I'd say. This PR enhances that with an actual progress bar, a cancel button and the Search appears in a new tab so you don't lose where you were.

- 🗂️ **Search opens in its own new tab** instead of replacing the tab you were browsing, and **activates immediately** — added an opt-in `forceSwitch` flag threaded through `Launcher::openInNewTab(bool)` → `MainWindow::addTab()` → `addTabWithPage()`, defaulting to `false` everywhere so every other caller (session restore, CLI paths) keeps respecting the existing "switch to new tab" setting unchanged.
- 🌀 **Status bar controls**, grouped together as permanent widgets: a fixed-width indeterminate `QProgressBar`, a static "Searching…" label, and a ⏹️ **Stop button** (`process-stop` icon) that cancels the search for the active tab — all shown only while that tab is actively searching.
- 🖱️ **Cursor fix**: the busy cursor for a search is now scoped to just the folder view (`folderView_->setCursor(...)`) instead of `QApplication::setOverrideCursor()`, so it no longer makes the Stop button (or the rest of the app) look "busy" for the whole duration of a long search. Regular folder loads keep the original global cursor, since those are near-instant.
- 🔢 **Live "found" count** — the normal status bar message updates in real time as results stream in, reusing the exact same text/pipeline that already showed the final count (no wording jump between "searching" and "done"). Phrased as **"N item(s) found"** (not "file(s)") since matches can be folders or symlinks too.
- 👆 **Selection no longer clobbers the count** — for search tabs specifically, selecting files now shows both side by side ("64 item(s) found  —  3 item(s) selected") instead of the selection text replacing the found-count entirely. Non-search folders keep the original (replace) behavior.
- 🚫 **Empty results** — a centered inline "No files found." message in the content area if the search finishes with zero matches.
- 🧼 **Interrupted-search cleanup** — navigating away from a still-running search (e.g. clicking a bookmark) now correctly clears the cursor, the inline label, and the status-bar controls, instead of leaving them stuck since `onFolderFinishLoading()` never fires for a job you've already disconnected from.
- 🔍🚫 Disabled the Search button/menu item while on a search-results tab, since its default location would otherwise be the bogus search://... path instead of a real folder.

https://github.com/user-attachments/assets/702fb37a-23f0-4bab-b8df-09f9af578e01

Companion PR in `libfm-qt` https://github.com/lxqt/libfm-qt/pull/1114


tsujan:
Thanks. At a glance, for now:

> Search opens in its own new tab

I'm afraid it shouldn't. Switching between the folder and the search result by going backward and forward is a feature.

> Selection no longer clobbers the count

Not needed.

> a centered inline "No files found." message in the content area if the search finishes with zero matches.

Not needed either. Please keep it as minimal as possible.

ahsand97:
> I'm afraid it shouldn't. Switching between the folder and the search result by going backward and forward is a feature.

Why not? I think it's better to open a new tab when you want to search so you don't lose where you were (I know you can click in the go back button and it goes back to where yo were) but I think having a separate tab for the search is visually better (just my opinion, I can revert that if you really think it would bloat the application).

> Not needed.

I disagree also in this one, let's say the search found 100 items, you see the label in the status bar "n items found" in real time, and then you select 3 items then the select label overrides the items found label (instead appears "n items selected") that's why I combined them.

> Not needed either. Please keep it as minimal as possible.

I do agree with this one hehe.

tsujan:
> Why not?

I told above. Moreover, finding in which folder the search was done by checking the path-bar may not be easy for everyone, jumping to a new tab to search **IN** a tab isn't good, etc.

> I disagree

Please let it be like everywhere else. There's no real gain.

> navigating away from a still-running search (e.g. clicking a bookmark) now correctly clears the cursor

That's also a feature: the user knows something is being done somewhere.

> since onFolderFinishLoading() never fires for a job you've already disconnected from.

Then it should be fixed.

tsujan:
In general, doing several things in a single PR isn't a good idea, although they may seem related to each other. The reason is that, if one of them breaks something, it'll be hard to find the commit that did it.

ahsand97:
> In general, doing several things in a single PR isn't a good idea, although they may seem related to each other. The reason is that, if one of them breaks something, it'll be hard to find the commit that did it.

My bad, I'm sorry haha, I'll be adding a new commit with the things discussed, in future PRs I'll try to just do minimal stuff then so it's easier to track any bug or commit that may introduce bugs.

tsujan:
Please use `QApplication::setOverrideCursor(QCursor(Qt::WaitCursor));` for searching too. As I said, the user needs to know something is happening elsewhere, like before.

> since onFolderFinishLoading() never fires for a job you've already disconnected from.

Why do you think so?

tsujan:
Another thing: the new progress-bar increases the status-bar's height a little, even if it's never shown. It may depend on the font size and other new widgets.

ahsand97:
> Please use `QApplication::setOverrideCursor(QCursor(Qt::WaitCursor));` for searching too. As I said, the user needs to know something is happening elsewhere, like before.

Oh I changed it because I tought it would be better to show the busy cursor in the main view (where the actual search is happening) so it doesn't look odd if you go to the status bar and click on the stop search button

tsujan:
> Oh I changed it because … so it doesn't look odd …

I know. But that's a small price to pay for a more important feature. Suppose that there's a heavy/long search and the user feels slowness when he's inside another tab/window. That global busy cursor tells him that something's happening. It isn't limited to searching.

tsujan:
In the middle of this, I found that screengrab doesn't show notification on saving/copying anymore — I hate it when several unrelated issues show up at the same time ;) I might be away for a while …

EDIT: I also found an edge-case and extremely rare crash in pcmanfm-qt. It isn't related to your PR. Now my mind is on the verge of exploding ;)

ahsand97:
> In the middle of this, I found that screengrab doesn't show notification on saving/copying anymore — I hate it when several unrelated issues show up at the same time ;) I might be away for a while …

Is it related to this?

> Another thing: the new progress-bar increases the status-bar's height a little, even if it's never shown. It may depend on the font size and other new widgets.

Fixed now.

Also reverted the logic that changed the cursor.

tsujan:
> Is it related to this?

Of course not :) I just don't know why unrelated issues are revealed to me when I'm dealing with a problem ;) That has happened to me many times.

Forget about `statusBarRowHeight` and especially `setFixedHeight` or `setIconSize`. They can easily result in side effects with some styles. The increase in the height was only 6 px here; I wouldn't call it a real issue.

# BUT:

Frankly I'm not convinced why we should cancel the search *and see an incomplete result*. The idea seemed good to me at first, but now I ask myself: What's the use of an incomplete search? It can be very misleading. If the user changes his decision, he could easily cancel the search by going backward.

If you have a good reason, please tell me. The PR's comment didn't include any.

And don't get me wrong: I'd ask the same question even if I myself wrote the patches for libfm-qt and pcmanfm-qt.

EDIT: Years ago, I made a PR twice, thinking that I was fixing an issue, while I was actually adding another issue. I closed it twice ;)

ahsand97:
> If you have a good reason, please tell me. The PR's comment didn't include any.

Let's say you're doing a search that is taking forever, you get some results and you don't want to lose them, so the stop search is the right thing to do, the search stops and then you get whatever results you have at that moment without losing them completely (what would happen if you click on go back).

ahsand97:
> Forget about `statusBarRowHeight` and especially `setFixedHeight` or `setIconSize`. They can easily result in side effects with some styles. The increase in the height was only 6 px here; I wouldn't call it a real issue.

So should I revert that logic also?

tsujan:
> you get some results and you don't want to lose them

Why should I want to keep them?! I haven't chosen how the search ends.

> So should I revert that logic also?

For now we're discussing if this and https://github.com/lxqt/libfm-qt/pull/1114 should be kept at all.

stefonarch:
My 2 cents to this change of the holy grail: the progress bar and increment showing of results is a nice improvement.

tsujan:
It isn't about liking or voting but reasoning. It's as simple as this: I can't know that what I want is among the shown results until the search is complete. The rest is random.

This reminds me of how Thunar (and perhaps some other FMs) tricks the user into believing it's faster than it is. GLib can be very slow on loading a huge folder on HDD because it may need to check the contents of files to know their mimetypes. Thunar pretends that it loads the folder instantly, while it adds files incrementally, changes the contents of the view, sets the scroll-bar, etc. until the folder is really loaded. Such a loading isn't usable until it's finished.

pcmanfm-qt (libfm-qt) doesn't cheat and doesn't waste resources on changing the view constantly while loading is being done. Incremental loading has no use; on the contrary, it can be misleading, to say nothing of wasting the CPU time on updating the view.

ahsand97:
I know about keeping everything as minimal as possible and also not bloating the app but I also think in the user experience and also keeping the DE modern, of course, without losing the philosophy of being very lightweight and modular (being an lxqt user myself for years and not comparing to other files managers) but I thought it'd be better if I can see the results appearing in real time, maybe the search is taking forever I still want to see the app is doing something and showing actual results (as an user experience visually I'd say) and I don't think the resource usage will grow that much (I can be very wrong in this), that's the whole point of this enhancement hehe.

Also to answer your previous question:


> Why should I want to keep them?! I haven't chosen how the search ends.

Let's say you're searching for thousands of files and it's taking a long time, you want to delete them or something, you can stop the search, delete the already found files and continue.

stefonarch:
>It isn't about liking or voting but reasoning.

This is your point of view as insider/coder. 
Users may have other views, and liking/unliking this or that feature or application is maybe the first one.

tsujan:
> This is your point of view as insider/coder.

Yes, of course. That's why liking/disliking isn't relevant: it's a matter of practicality and resource usage.

> I also think in the user experience and also keeping the DE modern

Oh, please! The word "modern" is usually used where there's no practical reason. KDE uses it a lot to rationalize its monstrosity. Incremental loading is neither "modern" nor proves anything. It can be used to cheat the user though: "Our app is so fast!"

> I still want to see the app is doing something

The busy cursor shows that.

> delete the already found files and continue

There's no guarantee that the files I want to delete are inside an incomplete result. It's random *and misleading*.

Please don't repeat the same argument with other words — I don't like to repeat the same replies. Give me a real reason. See it as a professional discussion.

ahsand97:
> What's the use of an incomplete search? It can be very misleading.

The angle I'd argue isn't "the user wants partial results" (I already made that case, and I get why it didn't land) — it's that an incomplete result isn't a new risk this PR introduces. It's already reachable today: click a bookmark or go back while a search is running, and you land on whatever the job had found up to that point. That's exactly why the "clear cursor/label/status-bar controls on navigate-away" cleanup was needed in the first place — onFolderFinishLoading() never fires for a job you've disconnected from. The PR doesn't create the "incomplete = misleading" scenario, it just gives it a deliberate entry point instead of it being a side effect of clicking away.

On the Thunar comparison: the concern there was the view re-rendering itself on every increment (re-sort, scrollbar, re-layout). That's not what's happening here — items are already appended to the model as the search job finds them, that's existing libfm-qt search behavior, not something this PR adds. The only new per-batch cost is one label update reusing the exact pipeline that already produces the final count string. So the CPU argument applies to search's existing incremental delivery, not to what's new in this PR.

If the real concern is that a stopped search could be mistaken for a complete one, I'd rather fix that directly — e.g. tag the status text "(stopped — incomplete)" on cancel — than drop the ability to stop at all.

Genuinely fine if this still doesn't clear the bar, just wanted to give you the actual mechanism instead of repeating the UX argument.

stefonarch:
>>This is your point of view as insider/coder.
>>
> Yes, of course. That's why liking/disliking isn't relevant: it's a matter of practicality and resource usage. 

Ehm - that's why  liking is relevant too - as your apps has users ;)

tsujan:
> If the real concern is that a stopped search could be mistaken for a complete one

No. An incomplete search is not only useless but also misleading. You didn't give a real use case. It's misleading because of its randomness.

Honestly I didn't understand the rest of your comment (clicking bookmarks, going back, "deliberate entry point"?!). If you're worried about the busy cursor, I told why it's needed and why it isn't limited to searching.

> just wanted to give you the actual mechanism

I have no problem to see the mechanism, being familiar with every code line of libfm-qt/pcmanfm-qt :) The discussion is about usability and practicality.

As for UX, I was fooled by it at first — that's why I wrote "It's a good idea" at https://github.com/lxqt/libfm-qt/pull/1114#issuecomment-5464877598.

> as your apps has users

They need to trust my judgements sometimes. There have been so many cases that I made changes proposed by users who distinguished between what *they* liked/disliked and what's practical.

ahsand97:
> You didn't give a real use case.

One other case, I'm not always searching for an exhaustive list. Often I'm looking for one specific file I already know exists — I stop as soon as I see it. E.g. searching a large home folder (or a slow/network mount) for a project file I vaguely remember the name of. The moment it appears, the search has done its job; I don't need it to keep grinding through the rest of the tree before I'm "allowed" to act on the one result I already confirmed with my own eyes. That's not trusting an incomplete set as a complete answer — it's ending a lookup exactly when it answered the question I asked.

I totally understand why you think is not practical the "delete the already found files" case I gave, you're right that relying on the partial set being exhaustive is misleading.

tsujan:
> The moment it appears, the search has done its job

How do you know that? There may be other files with the same name. If you know that it exists *uniquely*, most probably you also know where it is, and then a simple filtering will be enough. This use case is valid but I'm afraid it's too personal, rare, and based on luck (→ randomness). Am I missing something? I'm really asking.

Making the argument harder for myself, I don't think you started to implement the incremental search just because of this use case, but I may be wrong.

> I already confirmed with my own eyes

More personal, rarer, and more random ;)

ahsand97:
> I don't think you started to implement the incremental search just because of this use case, but I may be wrong.

Honestly, no, it wasn't. You're right that the unique-file case is weak — luck-based, like you said — I won't defend it further.

The real origin is more ordinary..., I don't even use pcmanfm-qt's search that often myself, I mostly search inside IDEs or with QTerminal. But the few times I did use it, it was usually a vague search over a large scope — root, or a big home folder — where I didn't know where the file was or how long the scan would take. Right now that gives you nothing to go on until it's fully done: not "is it close," not "is it finding anything at all," just a frozen-looking tab. Seeing matches accumulate isn't about trusting the partial list as an answer — it's information to decide whether to keep waiting, narrow the search, or give up, instead of deciding that blind.

I get that's closer to a comfort improvement than a hard correct argument, and I don't want to keep justifying just because I thought this would be a good idea. If this kind of improvement isn't something you think it'll be good, fine, I won't keep trying to find a harder argument for it.

tsujan:
> it's information to decide whether to keep waiting, narrow the search, or give up, instead of deciding that blind.

IMO, the search dialog is equipped with enough options for having a good strategy. I use it a lot but have never found a reason to do it blindly.

To sum it up, I'm suspicious about the usefulness of randomness (randomness is what I'd call "blind").

Let's keep the 2 PRs open for a while. IMO, it's too soon to decide. In the meantime, I might experiment with more ideas if I find the time.

Needless to say, I always appreciate your PRs, whether I agree with them or not.

ahsand97:
> Needless to say, I always appreciate your PRs, whether I agree with them or not.

Thank you, I also appreciate your knowledge in all lxqt ecosystem, also discussions are fine to see different ideas and how other peers think hehe.

tsujan (approved):

tsujan:
Contradicting myself: merging …

PR #2197, opened 2026-09-18
Small whitespace syntax fixes

isf63:
- `rg "\)\{"`
- `rg "(\}else|else\{)"`
- `rg " & "` (`Type&` seemed the most common)

Next to do is libfm-qt which will be today or tomorrow.

I'm back using LXQt somewhat after cleanly installing KDE Plasma alongside


tsujan:
> I'm back using LXQt somewhat after cleanly installing KDE Plasma alongside

It's good that at least one of us has a complete Plasma installation to test some things under it if needed (very rarely). I had it for a long time but got rid of it last year, leaving only the minimal stuff needed by KWin.

isf63:
>Plasma [...] LXQt

I've also managed to have no `xserver-xorg` installed; and using `kwin_wayland` for both. I plan on solely (if possible) using and testing kwin_wayland. Ubuntu 26.10 releases soon which should further enable Wayland.

tsujan:
> using kwin_wayland for both

Then you'll get the best support in lxqt-panel.

> Ubuntu 26.10 releases soon which should further enable Wayland.

Using LXQt+Wayland does not depend on the distro.

